PR #99, opened 2020-04-02
Introduce sphinx

KengoTODA:
This PR introduces Sphinx as site generator. You can download the generated file from the GitHub Actions run.

## Overview

* Maintain the source in the `docs` directory in the `master` or other release branches
  * We can change code and doc in one PR
* Put English content in the root dir, and others in a directory named its language code
  * https://jspecify.dev/ -> Top page for English user
  * https://jspecify.dev/ja/ -> Top page for Japanese user
  * If we need to maintain doc for several versions, need to add version number into the URL like https://jspecify.dev/1.0/ja/
    * For better automation, it would be better to use [ReadTheDocs](https://docs.readthedocs.io/en/stable/versions.html) or other service.
* Use [reStructuredText](https://docutils.sourceforge.io/docs/user/rst/quickref.html) as markup lang
  * [CommonMark](https://www.sphinx-doc.org/en/master/usage/markdown.html) is also able to use, however, it lacks some feature such as table and footnote

## TODO

- [x] Translation for Japanese
- [x] Automate deploy onto the GitHub Pages


cpovirk:
Sorry for taking so long to look at this. Thanks for doing it, and thanks for the instructions on how to view the generated file. It looks good to me.

I'm not sure if there are any other steps we need to take to make this work (like setting up the secret token?), but this sounds like a case in which I can just merge and see what happens :)

cpovirk:
(It sounds like we might [need](https://help.github.com/en/github/working-with-github-pages/about-github-pages#static-site-generators) a `.nojekyll` file in the `gh-pages` branch? But again, I'll merge and then see what happens. If we're really lucky, maybe Jekyll will just copy over all the generated `.html` files, etc.)

cpovirk:
And I see that Sphinx made its own `.nojekyll` file. I hadn't thought to look for hidden files in the generated zip :)

It's live: https://jspecify.dev/

Thanks again.

PR #103, opened 2020-04-21
Introduce Gradle, Errorprone, Spotless and GitHub Actions

### LINKED ISSUE #54: Which buildtool and CI we prefer?
Current `master` branch has no `pom.xml`, `build.gradle` nor `build.xml` then I want to make some of them to ease coming works. My suggestion is [Gradle with Spotless and Errorprone](https://github.com/google/codeanalysis-annotations/commit/97e7b7c9a06c6e72ae5f98007f2c932bc1fff5cb), let's have a talk to make our daily hack great.

I also want to introduce CI such as Travis CI and Circle CI, but I think that we have some limitation while this repository is private... is it better to wait until we publish it?

KengoTODA:
This change introduces build and CI toolchain. The package name is
still experimental, will be decided in #1.

close #54


build.gradle
```diff
@@ -0,0 +1,29 @@
...
+    id 'maven-publish'
+    id 'com.diffplug.gradle.spotless' version '3.28.1'
+    id 'net.ltgt.errorprone' version '1.1.1'
+}
+
+apply from: "$rootDir/gradle/format.gradle"
+apply from: "$rootDir/gradle/publish.gradle"
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
```
JakeWharton, line 12:
```suggestion
sourceCompatibility = JavaVersion.VERSION_1_8
targetCompatibility = JavaVersion.VERSION_1_8
```

gradle/license-header.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2018-present The Code Analysis Annotations Authors.
```
cpovirk, line 2: - Change "The Code Analysis Annotations Authors" to "The jspecify Authors."

- I don't know much about copyright statements, but our internal guidelines recommend a specific year rather than "present." I see a similar recommendation to avoid "present" in [this random Stack Exchange answer](https://opensource.stackexchange.com/a/4890/17871).

gradle/publish.gradle
```diff
@@ -0,0 +1,26 @@
+// run `./gradlew publishToMavenLocal` to install artifact to your Local Maven Repository
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            pom {
+                groupId = 'org.jspecify.experimentaldonotuse'
+                name = 'Code Analysis Annotations'
```
cpovirk, line 8: jspecify here, as well.

gradle/publish.gradle
```diff
@@ -0,0 +1,26 @@
...
+                description = 'An artifact of well-named and well-specified annotations to power static analysis checks'
+                url = 'https://jspecify.org/'
```
cpovirk, line 10: We have made jspecify.org redirect to jspecify.dev. That means that both are fine, but maybe it's nice to link directly to the .dev. domain?

src/main/java/org/jspecify/experimentaldonotuse/DefaultNullable.java
```diff
@@ -0,0 +1,30 @@
...
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Documented
+@Target({TYPE, PACKAGE})
+@Retention(CLASS)
+public @interface DefaultNullable {}
```
cpovirk, line 30: For at least our first release, we'll have only `@DefaultNonNull` and `@Nullable`. You can remove the others now, or we can do it in a followup.

cpovirk:
The CLA Bot appears to be upset about https://github.com/jspecify/jspecify/commit/1174f01a9a8c2c41a1fe656beb2ff45251161adb, which has:

> Co-Authored-By: Jake Wharton <github@jakewharton.com>

Jake is a Google employee.

JakeWharton:
Whoops! I'll switch my email for this org. Thanks.

cpovirk:
For my reference:

I wanted to squash this down to one commit, rebase that on top of master, and commit it.

I was going to use "Rebase and merge" (which would have been wrong, since I wasn't looking to preserve the _n_ different commits along the way), but luckily I hit an error: "Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch"

If my notes are correct, I eventually accomplished it this way:

```
$ git co master

$ git pull

$ git co -b redo

$ hub pr checkout 103

$ git co redo

$ git reset introduce-impl-and-ci -- $(git diff --name-only master introduce-impl-and-ci)

$ git commit --author="Kengo TODA <skypencil@gmail.com>"
# Fill in commit description.

$ git restore .

$ git co introduce-impl-and-ci

$ git reset --hard redo

$ git b -D redo

$ git push -f
```

It might have been simpler to just tar up the filesystem, `git reset --hard master`, and untar it. (I think I have done this before.) There is surely still a simpler way, even though what I'm doing is probably frowned upon, since it throws away history.

cpovirk:
In the general case, I guess that the `tar` approach (possibly using `git archive`) could require me to then `rm -r` the contents under the new branch if the PR is _removing_ any files. That's a little tricky because I could need to remove even hidden files -- but _not_ `.git`.

Anyway, it sounds like my `reset` + `restore` was just a complicated way of doing a `checkout`. So we can replace those 2 steps with 1. Additionally, there's no need to pass a list of changed files: I can just pass `.`. So that leaves me with:

```
$ git co master

$ git pull

$ git co -b redo

$ hub pr checkout 103

$ git co redo

$ git co introduce-impl-and-ci -- .

$ git commit --author="Kengo TODA <skypencil@gmail.com>"
# Fill in commit description.

$ git co introduce-impl-and-ci

$ git reset --hard redo

$ git b -D redo

$ git push -f
```

But the way that I'd been assuming must exist is more like [Raymond Chen's](https://devblogs.microsoft.com/oldnewthing/20190506-00/?p=102478):

```
$ git co master

$ git pull

$ hub pr checkout 103

$ git reset --hard $(GIT_AUTHOR_NAME='Kengo TODA' GIT_AUTHOR_EMAIL='skypencil@gmail.com' git commit-tree -p master introduce-impl-and-ci^{tree})
# NOTE: This waits for the commit description, but it doesn't tell you that it's doing that :)

$ git push -f
```

cpovirk:
Actually, it sounds like `restore` + `reset` might actually have been better than `checkout` [in the case in which files were removed](https://stackoverflow.com/a/37202812/28465)? But I'm noticing this when trying to do something slightly different, so I may just be confused.

cpovirk:
OK, a one-step `restore` + `reset` that isn't `checkout` might be [`git merge --squash`](https://stackoverflow.com/a/7519701/28465). But I haven't tested it on removing files, so I may still be confused.

```
$ git co master

$ git pull

$ git co -b redo

$ hub pr checkout 103

$ git co redo

$ git merge --squash introduce-impl-and-ci

$ git commit --author="Kengo TODA <skypencil@gmail.com>"
# Fill in commit description.

$ git co introduce-impl-and-ci

$ git reset --hard redo

$ git b -D redo

$ git push -f
```

I imagine there are still simplifications to be made :)

cpovirk:
...and this all raises the question of why I'm not just using "Squash and merge." I think I see the word "merge" and think "that's not what I want," since I don't want to end up with a commit with multiple parents. But "rebase and merge" contains the word "merge" and doesn't do that, either. So probably I should just have done that and avoided all this trouble :) (But I do want to do something much like this but _not_ through the GitHub UI in another repo, so it's good to learn -- though I probably could have learned more efficiently.)

cpovirk:
Since I've already been dumping everything related to Git shenanigans here...

Possibly also helpful in similar situations is `git read-tree`, [as described by mjd](https://blog.plover.com/prog/git-reorder-3.html).

PR #106, opened 2020-04-28
Add a link to the slide

KengoTODA:
This change adds a link to [the slide shared at JVMLS19](https://docs.google.com/presentation/d/1zmblWc_S9LrmKq8VH8p_eb7H31rTYoNFVbutnXPyolM/edit?usp=sharing).
Also applies a tiny fix, to make a header intuitive.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -26,7 +26,7 @@ This missing artifact, with full specifications, is what we are working to creat
 Talks about jspecify
 --------------------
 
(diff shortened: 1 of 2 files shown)


cpovirk:
It doesn't look like I have permission to view the slides unless I use my work email account. That suggests that people who come across our site won't be able to view them, either.

I'm not sure if there's even a way to make the slides publicly visible -- unless @kevinb9n moves the doc to be owned by a general public Google account instead of a work account. (Maybe he'd prefer to make a copy?)

KengoTODA:
I'll reopen this PR when we're ready.

PR #117, opened 2020-07-22
Enhance the build portability

KengoTODA:
I'm using this project as git-subtree (because we haven't publish jar files to Maven Central), then found that I need to apply [some updates in this project](https://gist.github.com/KengoTODA/9c930004e6e2ffa8f34d2a306357e422). Some of them are able to reflect to this repo itself, so I'm going to introduce it to the `master` branch.

It also upgrades Gradle and its plugins.


tbroyer:
> I'm using this project as git-subtree (because we haven't publish jar files to Maven Central), then found that I need to apply [some updates in this project](https://gist.github.com/KengoTODA/9c930004e6e2ffa8f34d2a306357e422).

I think the `$rootDir` wouldn't have been a problem if you had used a [composite build](https://docs.gradle.org/current/userguide/composite_builds.html) rather than (I suppose) integrating JSpecify as a subproject.

KengoTODA:
I tried the composite build and found that [it does not work with `maven-publish` plugin](https://docs.gradle.org/current/userguide/composite_builds.html#included_build_substitution_limitations).
I will keep working with `git-subtree` instead.

KengoTODA:
FYI here is our project, you may refer as an example.
https://github.com/spotbugs/spotbugs-jspecify-plugin

PR #120, opened 2020-08-18
Reduced spec for 0.1.

cpovirk:


docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+=========================
+
+About this document
+-------------------
+
+This document summarizes our current thinking for proposing a set of
+nullness annotations. **Nothing here is “set in stone” yet.**
+
+It will be very helpful to read or refer to the `core
+concepts/glossary <https://docs.google.com/document/d/1KQrBxwaVIPIac_6SCf--w-vZBeHkTvtaqPSU_icIccc/edit>`__
```
kevin1e100, line 11: any chance we could put the glossary into .md form as well?  The docs aren't public, but this page will be publicly visible, IIUC, so links to the docs are probably ill-advised in any case

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+--------
+
+We propose a set of type annotations to specify the intended nullness of
+individual type usages. At a high level, our proposed semantics is
+similar to existing tools’ treatment of nullness type annotations. We
+treat nullness annotations as purely compile-time type refinements
+without effect on runtime semantics. Besides the typical “nullable” and
+“non-null” refinements, we distinguish values of “unspecified nullness”,
+which typically arise from unannotated code, and allow for (with
+optional warnings if desired) unsound subtyping between “unspecified
```
kevin1e100, line 26: We should probably not call it subtyping.  how about "unsound coercions" or "unsound conversions" (as in the proposal doc)?

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+nullness” and known nullness states (i.e., nullable or non-null).
+
+To avoid excessive annotation overhead in hand-written Java code while
+clearly distinguishing annotated from legacy code, we also propose a set
+of explicit defaulting annotations. In the absence of a defaulting
+annotation, type usages will be considered of “unspecified nullness.”
+
+We give `semantics <#semantics>`__ to these annotations independent of
+particular tools. We don’t specify what tools must do with the
+information we provide, nor do we forbid them from adding to it (e.g.,
```
kevin1e100, line 36: "we provide" -> "provided using annotations defined here" or something like that?

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+[`#1 <https://github.com/jspecify/jspecify/issues/1>`__]
+
+We will release one canonical artifact, in which all annotations have
+runtime retention.
+[`#28 <https://github.com/jspecify/jspecify/issues/28>`__]
+
+The type-use annotations
+------------------------
+
+In this release, we will provide a parameterless type-use annotation
```
kevin1e100, line 72: nit: "a parameterless" -> "a single parameterless"

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+Finally, a **type check** (e.g., to determine assignability) for a pair
+of `augmented
+types <https://docs.google.com/document/d/1KQrBxwaVIPIac_6SCf--w-vZBeHkTvtaqPSU_icIccc/edit#bookmark=id.367l48xhsikk>`__
+includes validating **both** (a) a type check of the `base
+types <https://docs.google.com/document/d/1KQrBxwaVIPIac_6SCf--w-vZBeHkTvtaqPSU_icIccc/edit#bookmark=kix.e98jyoqkxvjd>`__,
+handled by the off-the-shelf compiler, and (b) a type check for
+nullability.
+
+-  This specification additionally encourages allowing **unsound
+   nullability subtyping rules** (optionally with warnings, similar to
```
kevin1e100, line 239: as above, don't call this "unsound subtyping".  additionally, the text reverts to using the word "conversion" as in the proposal doc further below, which isn't really explained

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+
+   -  nullable as a subtype of unspecified nullness
+   -  unspecified nullness as a subtype of non-null
+   -  unspecified nullness as a subtype of unspecified nullness
+
+-  For usability reasons, tools that want to generate warnings/errors
+   when applying unsound rules like those above are **strongly**
+   encouraged to permit the suppression of these warnings without
+   suppressing other soundness violations. (In fact, we expect this to
+   be the desirable default behavior.)
```
kevin1e100, line 252: this is probably ambiguously worded in the proposal.  I think what's meant here is that we expect most tools to not even warn about these unsound conversions.

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+corresponding bounds’ types (in addition to base type well-formedness,
+see `JLS
+4.5 <https://docs.oracle.com/javase/specs/jls/se14/html/jls-4.html#jls-4.5>`__).
+As usual, tools may still wish to allow (with warning if desired)
+unsound type arguments involving unspecified nullness.
+
+Specifically, a tool might reject a parameterized type with an explicit
+annotation, such as ``ImmutableList<@Nullable String>``, if
+``ImmutableList``\ ’s type parameter is bounded to be non-null. The
+toool might report an error when encountering this case in source code
```
kevin1e100, line 320: nit: toool -> tool

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+declarations involving unspecified nullness. Note that
+`supermethods <https://docs.google.com/document/d/1KQrBxwaVIPIac_6SCf--w-vZBeHkTvtaqPSU_icIccc/edit#bookmark=id.5nvbughni6vx>`__
+and
+`superparameters <https://docs.google.com/document/d/1KQrBxwaVIPIac_6SCf--w-vZBeHkTvtaqPSU_icIccc/edit#bookmark=id.m2gxs1ddzqwp>`__
+may be defined by members of parameterized supertypes.
+
+Intended “meaning”
+~~~~~~~~~~~~~~~~~~
+
+The broad intention of nullability is that type usages specified as
```
kevin1e100, line 349: we should probably also reference the definitions of nullable and unspecified nullness in the glossary from this section, and/or change the heading to more specifically discuss non-nullness and field initialization

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+   is ignored and might justify a warning. That warning can be avoided
+   by writing ``? extends String`` without changing the nullability:
+   ``String``, being not explicitly annotated, has unspecified
+   nullability, but that’s still superseded by the wildcarded type
+   parameter’s bound, non-null ``Object``.
+-  What’s the nullability of the map’s second wildcard? Also non-null,
+   because the wildcard inherits that bound from the bound of ``V`` in
+   ``ImmutableMap``.
+
+As another example, ``java.util.function.Function`` would be declared as
```
kevin1e100, line 415: maybe we should use guava's Function interface here?

docs/spec.rst
```diff
@@ -0,0 +1,501 @@
...
+   explicit annotation)
+-  It is a mismatch to refer to
+   ``ImmutableMap<@Nullable String, Object>`` because
+   ``@Nullable String`` is outside of ``K``\ ’s bounds.
+
+To illustrate wildcards, consider a method return type
+``ImmutableMap<? extends @Nullable String, ?>`` with no defaulting
+annotation in scope:
+
+-  What’s the map’s nullability? unspecified (since no defaulting
```
kevin1e100, line 400: AFAICT the section that defines nullability is missing, so this discussion seems weird.  maybe add that definition, or otherwise maybe remove some of this text?

docs/spec.rst
```diff
@@ -0,0 +1,506 @@
...
+   [`#33 <https://github.com/jspecify/jspecify/issues/33>`__]:
+
+   -  a nullable type where a subtype of unspecified nullness is
+      required
+   -  a type of unspecified nullness where a subtype of a non-null type
+      is required
+   -  a type of unspecified nullness where a subtype of unspecified
+      nullness is required
+
+-  For usability reasons, many tools will not generate generate
```
wmdietlGC, line 254:
```suggestion
-  For usability reasons, many tools will not generate
```

docs/spec.rst
```diff
@@ -0,0 +1,506 @@
...
+~~~~~~~~~~
+
+If a method overrides other methods according to Java language rules
+(see `JLS
+8.4.8.1 <https://docs.oracle.com/javase/specs/jls/se14/html/jls-8.html#jls-8.4.8.1>`__,
+also cf. \ `JVMS
+5.4.5 <https://docs.oracle.com/javase/specs/jvms/se14/html/jvms-5.html#jvms-5.4.5>`__),
+then tools may wish to check that:
+
+-  The overriding method’s augmented return type is be
```
wmdietlGC, line 340:
```suggestion
-  The overriding method’s augmented return type is
```

docs/spec.rst
cpovirk: That is something we'll need to do. I am hoping to call it out of scope for 0.1: "Fortunately," it still seems unlikely that anyone outside the group will view the spec. Do you think we can get away with that? We can always release an 0.1.1 if we want to update the links to point to somewhere publicly visible for historical interest.

docs/spec.rst
```diff
@@ -0,0 +1,506 @@
...
+   ``@Nullable String`` is outside of ``K``\ ’s bounds.
+
+To illustrate wildcards, consider a method return type
+``ImmutableMap<? extends @Nullable String, ?>`` with no defaulting
+annotation in scope:
+
+-  Can the method return a null map? That is unspecified (since no
+   defaulting annotation is in scope)
+-  Can the map’s keys be null? no. This is because the wildcard’s
+   explicit bound, ``@Nullable String``, is intersected with the
```
wmdietlGC, line 409: Maybe this intersection should be mentioned in the `Type arguments of parameterized types` section? It's a bit surprising to first read about this in the example.

docs/spec.rst
cpovirk: I worried about using "conversions" because the JLS idea of "conversions" is more specific that what we're aiming for. For example, we want for an unannotated `class Foo<T>` to be instantiable as `Foo<@Nullable Object>`. As I understand it from [JLS 4.5.1](https://docs.oracle.com/javase/specs/jls/se14/html/jls-4.html#jls-4.5.1), the rule that's applied there is containment (which applies subtyping), so it doesn't technically involve "conversion."

"Coercion" sounds workable, especially since I don't see that word anywhere in the JLS. Regrettably, I skimmed over that part of your message original, so I went to some contortions instead. Let me know what you think, and I can go with coercion if you prefer that.

docs/spec.rst
cpovirk: Tweaked slightly differently to try to keep the length down. But maybe it's less clear now? I can give it another shot if you think I should.

docs/spec.rst
cpovirk: Done.

docs/spec.rst
cpovirk: I think I have it so that the only references to "conversion" are to the existing rules for base types (unchecked and raw types). If you see any that I missed, please flag them.

docs/spec.rst
cpovirk: I rewrote this to try to clarify the various options. I don't love it, but maybe it's at least clearer about what the options are?

docs/spec.rst
cpovirk: Done.

docs/spec.rst
cpovirk: Oops, I did half of this. I added links to those definitions (though at the top of the doc), but I didn't change the section title. I'll do that now ("Concerns around uninitialized objects") and then push again.

cpovirk: Done.

docs/spec.rst
cpovirk: Done.

docs/spec.rst
cpovirk: Tried to rephrase to keep it intact. My fallback plan is to nuke it :) Let me know.

docs/spec.rst
cpovirk: Oh, yeah, that's in one of the parts I cut out. How about I cut this section down to refer only to the type parameters' bounds (see new version)? Or maybe we'd be better off without the section entirely?

cpovirk:
Hmm, any ideas, @wmdietlGC ? The internal CLA tool tells me that I can accept code from various @google.com email addresses, but it says "no" for yours :(

docs/spec.rst
kevin1e100: SGTM.  maybe put something like "TODO: make these available with spec" into this text then?

docs/spec.rst
```diff
@@ -0,0 +1,498 @@
...
+   -  A module. [*debated*
+      `#34 <https://github.com/jspecify/jspecify/issues/34>`__]
+   -  *Not* a method
+      [`#43 <https://github.com/jspecify/jspecify/issues/43>`__],
+      constructor
+      [`#43 <https://github.com/jspecify/jspecify/issues/43>`__], or
+      field [`#50 <https://github.com/jspecify/jspecify/issues/50>`__].
+      *debated*
+
+-  Applying more than one of these annotations to the same scope (since
```
kevin1e100, line 189: nit: maybe we can lose this rule (for both @Nullable and @DefaultNonNull), since javac will reject duplicate use anyway.

docs/spec.rst
```diff
@@ -0,0 +1,498 @@
...
+In this section, we discuss how type checking should play out.
+
+Type hierarchy
+~~~~~~~~~~~~~~
+
+Our nullability annotations produce the following apparent type
+hierarchy [`#80 <https://github.com/jspecify/jspecify/issues/80>`__]:
+
+| ``Foo`` (written in the context of ``@DefaultNonNull``)
+| ``<: Foo`` (written outside the context of ``@DefaultNonNull``)
```
kevin1e100, line 207: nit: you could use "<=" (or some other symbol) to avoid overloading Java's base subtyping

docs/spec.rst
cpovirk: Done.

docs/spec.rst
cpovirk: Done.

docs/spec.rst
kevin1e100: SGTM.  We can separately try to work on this some more in the underlying doc

docs/spec.rst
cpovirk: I think I made the same comment on the Docs version of this at one point :) As I recall, kevinb took the position that this part of the spec was essentially saying "These annotations will not be `@Repeated`." While it's hard to imagine that anyone would even think to define them that way, I suppose it's technically something that we want to forbid.

PR #121, opened 2020-08-20
Update the target branch name to build

KengoTODA:
We changed the default branch from `master` to `main`, so we need to use `main` in the `docs.yml` file.


Stephan202 (approved):

cpovirk:
Thank you!

PR #122, opened 2020-08-21
Document the format for samples.

cpovirk:


samples/README.md
```diff
@@ -18,3 +22,135 @@ for that code, and they always have the option not to.
...
+Fully qualified class names must be unique across all directories.
+
+> This permits all files to be compiled in a single tool invocation.
+
+TODO: Consider requiring that all individual-file samples be in the top-level
+directory.
+
+TODO: Consider requiring a file's path to match its package and class:
+
+-   individual-file samples: `Foo.java` for `Foo`
```
abreslav, line 43: Do we allow multiple package-local classes in one file? It's usually very convenient for tests

samples/README.md
```diff
@@ -18,3 +22,135 @@ for that code, and they always have the option not to.
...
+-   We may need additional accommodations for JPMS support to demonstrate
+    module-level defaulting.
+
+## Restrictions
+
+Files must be UTF-8 encoded.
+
+Files must contain only printable ASCII characters and `\n`.
+
+Files must be compatible with Java 8.
```
abreslav, line 57: Do you mean that we can't have tests for newer language features? I suspect that local variable type inference will need some test cases

samples/README.md
```diff
@@ -18,3 +22,135 @@ for that code, and they always have the option not to.
...
+-   `jspecify_nullness_mismatch`: for certain instances of case 2
+
+-   `jspecify_nullness_not_enough_information`: for certain instances of case 2
+
+The difference between the "mismatch" and "not enough information" cases arises
+from
+[unspecified nullness](https://docs.google.com/document/d/1KQrBxwaVIPIac_6SCf--w-vZBeHkTvtaqPSU_icIccc/edit#bookmark=id.xb9w6p3ilsq3).
+The rough idea is that a type with unspecified nullness means: "When this code
+is annotated for nullness, either the type should have `@Nullable` or it, or it
+shouldn't."
```
abreslav, line 121: Looks like a typo around "or it, or it shouldn't", but can't decipher what was the intended meaning

abreslav (approved):

samples/README.md
```diff
@@ -18,3 +22,135 @@ for that code, and they always have the option not to.
 The hope is that samples like these may be useful to those who wish to discuss
 the spec or implement tools. To that end, we welcome comments on both the format
 of these files and their content.
+
+## Directory structure
+
+See
+[jspecify: test-data format: Directory structure](https://docs.google.com/document/d/1JVH2p61kReO8bW4AKnbkpybPYlUulVmyNrR1WRIEE_k/edit#bookmark=id.2t1r58i5a03s).
+TODO: Inline that here if Tagir can sign the CLA and contribute it.
```
amaembo, line 30: Looks like I already signed Google CLA:
image

samples/README.md
amaembo: In most of the cases, it's possible to create nested classes. I think it's better (have a single top-level class and place all other classes inside), as makes it easier to avoid the name clash. If some test absolutely requires several top-level classes, it's also should be ok, though probably some naming convention (e.g. common prefix) would be nice.

samples/README.md
amaembo: I think, this should be rectified. Probably we should create separate packages for minimal language levels. I can imagine that some tools may be slower in adopting new language versions. In this case, then still may use part of test-suite to validate.

cpovirk: Makes sense. Should I move the existing samples under a `java8` directory to make space for for future `java10`, etc. directories?

samples/README.md
cpovirk: The first "or" should be an "on."

I'm still not sure if I put this clearly, so let me know.

cpovirk (commented):
Thanks, and sorry that I've left this waiting for so long.

cpovirk:
I opened #130 about having separate directories for "lenient mode" and "strict mode."

I am planning to optimistically merge this PR, since I've heard from at least mernst that it's hard to find when it's a PR instead of an actual README :) I am still happy to make additional changes (and in fact I plan to create another PR shortly for mernst's requested split of the "paradox" cases).

samples/README.md
cpovirk: I thought about the `java8`/`java10` split a little more, since I was also thinking about the strict/lenient split in #130. I think we may want to err on the side of using comment "tags" over directories, but I'm uncertain. I think it should be safe to leave a TODO for now, since sample inputs that use `var` don't yet exist (though we should add them sometime).

cpovirk: https://github.com/jspecify/jspecify/issues/131, https://github.com/jspecify/jspecify/issues/132

samples/README.md
cpovirk: https://github.com/jspecify/jspecify/issues/134

cpovirk:
-> https://github.com/jspecify/jspecify/blob/main/samples/README.md

PR #125, opened 2020-09-12
Link to the new page from the site top

KengoTODA:
[The newly added page](https://jspecify.dev/tsttcpw.html) is not linked from the site top.
This PR suggests to link it from the side menu.

### How the link will be available

Screen Shot 2020-09-12 at 19 31 48

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,10 +1,6 @@
 jspecify: Standard Java annotations for static analysis
 =======================================================
 
-.. toctree::
-   :maxdepth: 2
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks. We're still not sure what we'll be doing with this page:

- Integrate it into the spec?
- Provide it as a "implementation suggestion" for tool authors?
- Delete it entirely?

But it should be fine to link to it so that it's more accessible to project members. We can always make changes as needed.

PR #126, opened 2020-09-12
Minor updates for document

KengoTODA:
* Bump up Sphinx version to the latest one
* Fix two warnings
  1. malformed table that makes [participants in the latest doc](https://jspecify.dev/#participants) invisible
  2. no directory found for static resources
* Add more translation for Japanese language


cpovirk:
Thank you, especially for fixing the participants table! I should have checked it before merging.

Issue #101, opened 2020-04-13
A meta-annotation to express "incompatible with <other annotations>"


kevinb9n:
For example imagine:

```
@IncompatibleWith(Nullable.class)
public @interface NonNull {...}
```

This would express that it is invalid to use the two annotations together in the exact same spot.

By convention we could do it in both directions but only one would be necessary.

kevinb9n:
A compliant checker would not permit the conflicting annotations to appear together in code it is analyzing. But what if they appear together in some dependency of the code being analyzed?

I would suggest that checkers absolutely should not choose based on position (even if that order is preserved in the class file). And trying to specify an order of precedence somehow feels like a big waste of complexity with little to gain. And deciding to just ignore them all won't lead to anything good.

So, we might actually want the `@IncompatibleWith` docs to explicitly condone giving an error even if it's a dependency that contains the violation.

mernst:
If there are annotations that express a subtyping relationship, such as [`@SubtypeOf`](https://checkerframework.org/api/org/checkerframework/framework/qual/SubtypeOf.html), they can serve this purpose.  We might still need `@IncompatibleWith`, but maybe not for type annotations.

cpovirk:
For jspecify, supporting this feels somewhat speculative to me, similar to [how I feel about `@NegatedBy`/`@NegatedOf`](https://github.com/jspecify/jspecify/issues/109#issuecomment-626006545).

The Checker Framework, by contrast, has more reason to support `@SubtypeOf`: It is part of what lets users define arbitrary new type systems.

Kevin has [replied to my concerns on the other thread](https://github.com/jspecify/jspecify/issues/109#issuecomment-626011093). I agree that there's _some_ value to the annotations proposed in both this issue and #109, but both still feel to me to not be part of the minimum viable product.

kevinb9n:
(I can see making the actual release of said annotations lower priority, for sure. I just want us to be thinking ahead a bit, so that in 5 years we have a nice stable of stable annotations, where learning one actually helps you to learn another.)

cpovirk:
One potentially nice thing about meta-annotations in general is that it lets us make changes in the annotations themselves and then have the behavior change automatically picked up by tools. Alternatively, that is a potentially scary thing :) But I'm not sure if it's necessarily any scarier than the idea that we'd make significant changes to existing annotations _period_.

[...]

PR #135, opened 2020-09-29
Split "paradox" into 3 categories.

cpovirk:


cpovirk:
I had a realization about the new categories: Only one is purely "about" nullness:

- `@Nullable int i`

The others are more general:

- `@Nullable @NullnessUnspecified Foo`: This is a general category of "mutually incompatible" annotations. If we provided enough metadata, it could be checked by a tool that knows nothing about nullness (#101).

- `class @Nullable Foo {}`: This is a general category of "unrecognized `TYPE_USE` locations." Again, with enough metadata, it could be checked by a tool that knows nothing about nullness (#136).

Based on that, I have tentatively named those 2 categories "jspecify_multiple_annotations" and "jspecify_unrecognized_location," _without_ putting "nullness" in the name as I've been doing for the others (including new "jspecify_nullness_intrinsically_not_nullable"). @kevin1e100 , I think you may have been the one to suggest including "nullness" in the name of the initial tokens. Does omitting it from those 2 categories make sense to you? (Not that I want to overthink this :))

Once there is agreement on the names, I can update the sample inputs themselves accordingly.

samples/README.md
```diff
@@ -110,9 +110,20 @@ cases above.
 
 A comment on a given line provides information about the following line.
 
-Such a comment contains one of 3 special sequences:
+Such a comment contains one of 5 special sequences. The first 3 cover case 1
+from above:
 
--   `jspecify_nullness_paradox`: for case 1 above
+-   `jspecify_multiple_annotations`: for cases like `@Nullable
```
mernst, line 116: This could also be "conflicting".  When Jspecify expands, it may permit multiple annotations at a single location, such as `int @NonNull @NonEmpty []`.

mernst (approved):
This looks good, thanks.

samples/README.md
cpovirk: Makes sense. Done.

(Mike and I had talked elsewhere about using the term "inconsistent." We set it aside because readers might also conclude that an annotation can be "inconsistent" with a corresponding annotation on a supermethod, for example. While we could have the same concern about "conflicting," I feel that it's clearer somehow. Perhaps it's more natural to assume that a "conflict" occurs between two things that are sitting right next to one another in the source file, whereas "inconsistency" is the kind of thing that you might see even between 2 completely separate codebases.)

I'll leave this PR open for a couple days to collect feedback from @kevin1e100 and other interested parties. After that, I'll make the changes and merge it. As always, we can revisit as needed.

PR #142, opened 2020-10-09
Refactor samples README

mernst:
* Structure comment syntax around normative vs. non-normative
 * Also explain concepts where they are used, not later.
 * Put the most important information first.


cpovirk:
Thanks, and sorry again for leaving this so long.

PR #144, opened 2020-10-09
Add missing errors to DereferenceIntersection

mernst:


cpovirk:
The idea here is that, if an analyzer performs capture conversion, then the return type of `get()` is something like `Object & @NullnessUnspecified Lib` (or `Object & @Nullable Lib` in those cases). In that case, we know the value is not null (since it is a subtype of `Object`), so the call is safe.

IIRC, the Checker Framework does not permit "out of bounds" wildcards in the first place. That is, Java permits things like:

```java
class Foo<S extends String> {}

Foo<? extends CharSequence> foo;
```

But CF forbids (again, IIRC):

```java
class Foo<T extends @NonNull Object> {}

Foo<? extends @Nullable Object> foo;
```

(This may have to do with the Checker Framework's current behavior of not performing capture conversion but rather arriving at the same results in other ways?)

It may be that these tests end up requiring some special tag so that the Checker Framework analyzer's tests can know to expect different behavior for them.

kevinb9n:
(As explained)

PR #148, opened 2020-10-13
Add a few samples based on support of jspecify annotations in Kotlin

PetukhovVictor:
Here is a few tests from Kotlin with rewritten Kotlin use sites to Java ones.

It's quite possible that some annotations are missing or `jspecify_nullness_mismatch` is misused somewhere in accordance with the latest spec, so the review is required. Now inside Kotlin, we can only handle `jspecify_nullness_mismatch` (compare with errors or warnings from the Kotlin type enhancement).

The next step for Kotlin is to take all existing jspecify tests, automatically convert them to Kotlin use sites and support them in our test infrastructure if needed. After that we are going to create a public build with status of those tests passing.


cpovirk:
Thanks. JetBrains has already accepted the corporate CLA, so we need you to do this part:

> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).

PetukhovVictor:
I've found out that JetBrains doesn't restrict employees from participating in Google open source projects so it seems no action needed, right?

cpovirk:
@googlebot rescan

cpovirk:
Thanks again for these, and I'm sorry again that I've taken so long to get to them.

I've pushed several commits to your repo. I hope that's OK.

I have tried to separate out the cleanup changes (like adding licenses and reformatting code) from the functional changes.

Th functional changes are:
- https://github.com/jspecify/jspecify/pull/148/commits/d571282356f315225cc323349ecab013444200b1: Some files had a `Use` class that was not annotated `@DefaultNonNull`. This meant that their method parameters all had unspecified nullness. Thus, many operations on those parameters would need `jspecify_nullness_not_enough_information` comments. To avoid adding those comments, I changed the files to operate on fields or methods that are known to produce non-nullable values.
- https://github.com/jspecify/jspecify/pull/148/commits/6560f3bee65c731632e3c84bf382805a3fd84679: I found some places that appeared to need `jspecify_nullness_mismatch` or `jspecify_nullness_not_enough_information` comments. Most of them were `jspecify_nullness_not_enough_information`. I wouldn't expect Kotlin to issue any warnings or errors for those, so I think the changes are no-ops for you. In some cases, I did add `jspecify_nullness_mismatch`. Usually, that was because a line contained chained method calls, one of which was safe and one of which was not. If you'd prefer to remove the unsafe calls, that would be fine with me.
- https://github.com/jspecify/jspecify/pull/148/commits/3c72e8eebc32706ee3cbcc9fc7823aa77afba9f5: I go into a little detail about this in the commit message. It's about `<?>`.
- https://github.com/jspecify/jspecify/pull/148/commits/89c49e75b586fad757295aadd6803137260c252e: I go into a little detail about this in the TODO I added. It's about usage of type variables when the file does _not_ have `@DefaultNonNull` but the type parameter has a `@Nullable` bound.

Let me know what you think of all this.

One other note: I am going to try to move these files to separate directories and packages. That way, every class has a unique fully qualified name, as covered in [this section of the README](https://github.com/jspecify/jspecify/blob/e1e8a9bd75d4a434134cd788d2f0496b33530abd/samples/README.md#directory-structure). I'll push a commit for that shortly.

PetukhovVictor:
Thank you, @cpovirk!

I think the changes are fine to me. If something diverges from Kotlin side, then after a while I will just create an additional PR.

`jspecify_nullness_not_enough_information` may indeed have been missed in some places, mostly due to the fact that we are not yet checking this jspecify mark on the Kotlin side.

cpovirk:
Thanks again. I've merged this, and one of the new samples already prevented me from introducing a new bug.

PR #156, opened 2020-12-03
add sample with annotated receiver

kevin1e100:
Receiver should be considered categorically non-nullable.

Not super sure I'm using the right error codes here so please review carefully.

### DIFF
--- a/samples/AnnotatedReceiver.java
+++ b/samples/AnnotatedReceiver.java
@@ -19,21 +19,15 @@ import org.jspecify.annotations.Nullable;
 import org.jspecify.annotations.NullnessUnspecified;
 
 @DefaultNonNull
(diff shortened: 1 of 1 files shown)


cpovirk (commented):
Thanks! This raises some interesting issues.

tl;dr: I suggest having only:

```java
void foo(AnnotatedReceiver<T, U> this) {}

// jspecify_nullness_intrinsically_not_nullable
void nullableReceiverInvalid(@Nullable AnnotatedReceiver<T, U> this) {}

// jspecify_nullness_intrinsically_not_nullable
void unspecifiedReceiverInvalid(@NullnessUnspecified AnnotatedReceiver<T, U> this) {}
```

And I am interested in your thoughts on the newly opened issues.

kevin1e100:
removed annotated parameter examples

PR #159, opened 2020-12-30
add samples for some more forbidden type annotation locations

kevin1e100:
Couldn't find these elsewhere but may have missed it in all these samples.


cpovirk:
Thanks! I put it in another file (with a kind of ugly name, since I couldn't think of a good one -- maybe I should have just split it into multiple files?).

PR #163, opened 2021-02-04
Build a multi-release Jar

### LINKED ISSUE #161: Configure build process to create a multi-release jar
...and then create 2 separate copies of [`@DefaultNonNull`](https://github.com/jspecify/jspecify/blob/4fc89fd49663263394bfde630509056df10a816b/src/main/java/org/jspecify/annotations/DefaultNonNull.java), one that includes `MODULE` in its `@Target` values (for `--release 9` and higher) and one that does not.

Here are [some instructions](https://blog.gradle.org/mrjars#how-to-create-a-multi-release-jar-with-gradle) we could use.

Previous discussion: https://github.com/jspecify/jspecify/issues/34#issuecomment-768453364

KengoTODA:
Build a multi-release jar, and run integration-test with Java 8 and Java 9+.
To verify the build result, I will list contents in the jar files as follows:

```sh
$ jar tf build\libs\jspecify-0.1.0-SNAPSHOT.jar
META-INF/
META-INF/MANIFEST.MF
org/
org/jspecify/
org/jspecify/annotations/
org/jspecify/annotations/DefaultNonNull.class
org/jspecify/annotations/Nullable.class
org/jspecify/annotations/NullnessUnspecified.class
module-info.class
META-INF/versions/
META-INF/versions/9/
META-INF/versions/9/org/
META-INF/versions/9/org/jspecify/
META-INF/versions/9/org/jspecify/annotations/
META-INF/versions/9/org/jspecify/annotations/DefaultNonNull.class
```

closes #161


cpovirk:
Thank you! I will probably learn some Gradle one of these years, but until I do, I can't imagine how long I would have to spend to put something like this together.

PR #166, opened 2021-02-07
Remove false positive warnings

mernst:
These test cases mandate a false positive warning in correct code.
We don't want our test cases to contain false positive warnings that not all checkers will issue.

In the future, they could be added with a different (non-normative) tag.  In the meanwhile, I think it is better to remove them rather than have a broken test suite.

### DIFF
--- a/samples/NullCheck.java
+++ b/samples/NullCheck.java
@@ -20,16 +20,6 @@ import org.jspecify.annotations.NullnessUnspecified;
 
 @DefaultNonNull
 class NullCheck {
-  // TODO(cpovirk): Soften README to permit flow-sensitive samples in moderation.
-  Object x0(Object o) {
-    if (o != null) {
-      return o;
-    } else {
-      // jspecify_nullness_mismatch
-      return o;
-    }
-  }
-
   Object x1(@NullnessUnspecified Object o) {
     if (o != null) {
       return o;
@@ -44,15 +34,6 @@ class NullCheck {
       return o;
     } else {
       // jspecify_nullness_mismatch
-      return o;
-    }
-  }
-
-  Object x3(Object o) {
-    if (o == null) {
-      // jspecify_nullness_mismatch
-      return o;
-    } else {
       return o;
     }
   }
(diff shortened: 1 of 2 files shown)


wmdietl:
Could you explain why you consider these false positives?

For the first example that this PR removes:

````java
@DefaultNonNull
class NullCheck {
  // TODO(cpovirk): Soften README to permit flow-sensitive samples in moderation.	
  Object x0(Object o) {	
    if (o != null) {	
      return o;	
    } else {	
      // jspecify_nullness_mismatch	
      return o;	
    }	
  }
````

In the else branch `o` is definitely null, so the mismatch against the non-null return type (note the `@DefaultNonNull` on the class) seems waranted. Similarly for the second example.

wmdietl:
Ah, looking at the examples again I see that the parameters are `Object o` and therefore non-null.
So the null checks are redundant and even in the branch where `o` could be null according to the test, that possibility is excluded by the parameter type.

Would it be better to keep these tests but remove the incorrect expected errors? A comment could point out that tools could have an optional warning on the redundant check.

mernst:
You are right that other alternatives are to keep the tests but remove the false positive errors, or to mark them as optional.

I prefer marking the errors as optional, or removing the test entirely.  Otherwise, the test encodes an assumption about what analyses a tool implements.  That is, I could see JSpecify implementations that do issue errors there, as well as ones that do not.  I think that both implementations ought to pass the test suite.

cpovirk:
As always, I feel obligated to point out to anyone who is reading for the first time (not you two) that these are samples, not a conformance test. Thus it can be OK for tools to disagree about how to handle them. That said, I agree that it's sketchy to even call this example "a mismatch based on the information provided by JSpecify annotations."

In fact, _all_ code in this particular file goes beyond that strict definition, since it relies on flow-sensitive typing. (Hence the TODO at the top of the file for me to revisit the definition.) And that leads to the bigger point, which is: The _real_ goal of the samples is to be useful to tool authors. Even if this sample were The Right Choice for Everyone, that doesn't mean that it's important enough to justify causing difficulty for tool authors. (And yes, as you note, it's not at all clearly the right choice, even on average, let alone for everyone. I can see an argument that it's good if you expect your code to be called by users who do not run a nullness checker and bad if not, for example.)

Anyway, that's all just a bunch of commentary to back up what I've done, which is to merge this (with a couple related changes). Thanks.

PR #168, opened 2021-02-07
Make class name consistent with file name

mernst:


cpovirk:
I am extremely confused about what Git is showing for this PR.

In this review page, I see the expected change.

But after I merged it, it showed up as a no-op: https://github.com/jspecify/jspecify/commit/41961e9c6c26800a64599b35cd57a5b56ea48263

(And the same is true of https://github.com/jspecify/jspecify/commit/3f6c66030351181f05b7c159ed6ee995ea93f0e7 (#167), though I haven't dug further into that.)

My local Git clone has been showing the correct class name, probably all along.

But `git bisect run grep MultipleAnnotations samples/ConflictingAnnotations.java` says that the name was wrong until this commit.

I wonder if something is actually wrong with the repo. If so, I hope it's not severe....

cpovirk:
Oh, it looks like you'd done this back in October in #143.

And similarly for #167, which does the same as #153.

I'd been misreading the `git bisect` output and other history because I'd been confusing the original PRs with these new PRs.

So that makes sense after all.

Issue #49, opened 2019-07-23
Require a finding on a contravariant parameter type? [working decision: no]


dzharkov:
In the draft proposal, there's a part that allows for overrides to have value parameters with contravariant nullability:
> The method’s parameter types may have wider nullability than the corresponding superparameters' types, meaning, it suffices for superparameters to be type-convertible to the overriding method’s parameter types (while base types have to follow the normal rules of the compiler so overriding works as expected).

That might be a problem for the Kotlin compiler since in pure Kotlin value parameters are not contravariant in overrides and we apply the same logic when looking at the annotated Java code.

The reason we don't allow contravariant parameters in override is basically the same as in Java: they might be considered as overloads. In Java, one can not override with a wider-typed parameter (types should be equal). The same applies for Kotlin, but we have nullability as a built-in part of the type system.
```
interface A {
    fun foo(x: String)
}

class B : A {
    override fun foo(x: String) {
        
    }
    
    @JvmName("fooNullable")
    fun foo(x: String?) {} // is a different overload for `foo`
}

fun main() {
    B().foo("") // resolved to the not-nullable
    B().foo(null) // resolved to nullable
}

```

That means that when we see contravariant nullability in Java annotations, such code is assumed to be illegal thus we ignore annotations info for such parts completely as inconsistent:
```
interface A {
    void foo(@NotNull String x);
}

class B implements A {
    @Override
    public void foo(@Nullable String x) {

    }
}
```

`B::foo` has signature like `fun foo(x: String!)`, i.e. it has unknown nullability from Kotlin point-of-view.

But at the same time it's fine transforming from unknown nullness to both `NotNull` and `Nullable`:
```
interface A {
    void foo(String x, String y);
}

class B implements A {
    @Override
    public void foo(@Nullable String x, @NotNull String y) {

    }
}
```

In the above example `B::foo` is perceived by Kotlin as `fun foo(x: String?, y: String)`

mernst:
Formal parameter type contravariance is consistent with Java, so it is natural and strongly desirable.  If we were designing Kspecify rather than Jspecify, we might make a different decision.

I think we should accommodate Kotlin whenever we can, but we should never degrade the behavior of Jspecify on Java code in order to conform to Kotlin's different semantics.

kevinb9n:
I've kept coming back to this in my mind, and am just gonna reopen it for now. It may not be a critical priority right now, but I don't think that's what the open/close status is supposed to show.

[...]

PR #169, opened 2021-02-07
Remove false positive warning

mernst:
Behavioral subtyping permits contravariant formal parameters:  the formal parameter of `m` in a subclass may be wider than the formal parameter of `m` in a superclass.  This pull request removes a mandated warning that unnecessarily forbids such contravariance.  Perhaps it would be better to turn this into an optional (checker-specific) warning, or to remove the method entirely in order that JSpecify does not take a stand.


wmdietl:
This test case corresponds to issue #49 and, if I read the currect status correctly, at the moment we mandate this error.
So we shouldn't change this testcase without updating that issue and the documentation.

mernst:
Thanks for linking the issue -- that is helpful.  I agree with you that this PR, the issue, and the documentation should be kept consistent.  Maybe this can prompt us to revisit that issue.

cpovirk:
I see that you've posted on #49. Thanks. I have some thoughts that I'll put there, as well. (I see that is was actually reopened a while ago.)

If we want, we have some flexibility about what to do here even before that's settled.

On the one hand, the samples README describes `jspecify_nullness_mismatch` as applying in cases in which applying the normal JLS rules to nullness types would reject a piece of code. And it's true that normal JLS rules would reject an override with a different parameter type. So, even without officially writing a rule about this case one way or the other, we _theoretically_ already have an answer: This sample is "correct."

But we need to step back and consider the real point, just as in #166: The purpose of the samples is to be useful to tool authors. If we want to tool authors to be able to permit contravariance in parameter types, then a sample that works against that is at best controversial, and we may be better served by removing it. Granted, that still takes us back to debate of what we _do_ want. But at least it's a potentially easier question: "Do we want to allow tools to support this?" rather than "Do we want to write a part of the spec specifically to cover this?"

Again, though, for the actual heart of the debate, I'll post on #49.

kevinb9n:
Not actionable while #49 is undecided, and will be the obvious consequence if #49 _is_ decided in this direction, so this PR has no current purpose; closing to avoid "PR blindness".

Issue #147, opened 2020-10-10
Samples formatting


mernst:
The samples should use the recommended Java formatting style in which type annotations are written immediately before the type they modify, not on a separate line.  For example, `ClassToObject.java` should contain

```java
  @Nullable Object x6(ClassToObject x) {
    return x;
  }
```

rather than

```java
  @Nullable
  Object x6(ClassToObject x) {
    return x;
  }
```

This is important because we should not distribute sample code that is known to be written in poor style.

The samples could even contain a warning about mal-formatted code like the latter.  That seems more general (less specific to one particular implementation of JSpecify) than some of the current warnings like `jspecify_unrecognized_location`.

cpovirk:
Maybe we should swap out `google-java-format` for something else until they let us hardcode a list of annotations to consider as type annotations. Any suggestions?

mernst:
I think the best solution is to fix https://github.com/google/google-java-format/issues/5.  It requires reading a side file that lists "known type annotations".

Failing that, there is [run-google-java-format](https://github.com/plume-lib/run-google-java-format), which does the same thing, just in a less efficient and more hacky way.

kevinb9n:
Thanks, I've poked at google/google-java-format#5; this samples issue is just the harbinger.

kevinb9n:
Technically this could stay open until we have a solution that _keeps_ the samples formatted right - should we?

PR #170, opened 2021-02-09
Put type annotation before the type, not on its own line

mernst:


cpovirk:
I do agree that what we have is bad style, but I think it's currently more important to support out-of-the-box autoformatting than it is to fix the style. As discussed in #147, we'd be best served by getting google-java-format to do the right thing. In the meantime, if you're up for wiring up run-google-java-format for this project, I'd be on board with using it for our autoformatting.

mernst:
@cpovirk I don't see any autoformatting used on the project.  Can you point me at it if I missed it?

cpovirk:
Sorry, I had seen that Kengo was adding autoformatting in #117, but I hadn't actually merged that change until last week.

mernst:
Thanks for that pointer.  I pulled in master, and this branch still passes, so CI is not checking formatting, nor is `./gradlew build`.
I ran `./gradlew spotlessApply` in this branch and it did not undo the formatting change.
Am I still missing something?

cpovirk:
Hmm. I do not know. Maybe @KengoTODA can explain how to trigger autoformatting.

Setting that aside for a moment: From my perspective, I'm not invested in pre-commit-time or build-time autoformatting. I just want for people to be able to run some command that puts the files in the right format. And if there _is_ some kind of automated process, I don't want it to undo that work :) Given that there's possibly not currently an automated process, I am pretty happy to download `run-google-java-format.py` and run that instead of `google-java-format`. I'll have a look. But I'm first going to respond to the samples discussion and then return to the Javadoc I've been working up for our discussion of `@DefaultNonNull`, so it won't make its way to the top of my list just yet.

cpovirk:
(If there are new merge conflicts in the meantime, I'm happy to rerun `run-google-java-format.py` before merging, so don't feel that you have to keep this PR up to date.)

cpovirk:
Some quick testing suggests that running `run-google-java-format` on the samples takes ~2.5 seconds vs. ~0.5 for plain `google-java-format`. I would assume that could be optimized, but as it is, it's the small drawback that has kept me from pulling the trigger on this: It's nice to be able to run formatting every time I run the tests that use the samples, since that ensures that the special comments won't move later the way that they might if I ran tests first and then ran formatting later as a presubmit.

(Again, though, I think everyone would in principle like for `google-java-format` to just handle this natively. I'm making a note to at least revisit the `google-java-format` feature request when things calm down in the distant future.)

PR #177, opened 2021-02-16
Replace 2020 with 2021 in document and copyright notice

KengoTODA:
Happy new year! 🐮
Hope that you all have a nice days in this year.

### DIFF
--- a/cftojspecify/java/com/google/devtools/javatools/typeannotationrefactoring/CheckerFrameworkToJspecifyRefactoring.java
+++ b/cftojspecify/java/com/google/devtools/javatools/typeannotationrefactoring/CheckerFrameworkToJspecifyRefactoring.java
@@ -1,4 +1,4 @@
-// Copyright 2021 The JSpecify Authors
+// Copyright 2020 The JSpecify Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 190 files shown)


cpovirk:
Thanks, happy new year :)

We have some internal guidance not to update copyright dates in existing source files but instead to keep the year in which the file was created.

I don't know if that is for legal reasons or just to avoid churn in the files. If it is for legal reasons, I don't know if the same reasons apply here as in internal Google code.

I also don't know if the rules are different for documentation. It does seem like it would make sense to at least update `docs/conf.py`, since that gets included in multiple pages, which I assume were not _all_ created in 2020.

Probably it also makes sense to update `gradle/license-header.java`, since I think that may be added to newly created files?

KengoTODA:
@cpovirk Thanks for the review! I reverted the commit and changed just two things:

1. The config for spotless. It will affect only newly created files. Refer to [official doc](https://github.com/diffplug/spotless/tree/9dd69f7ff7feb6b2b7d26285a5b5ea6896a935ad/plugin-gradle#license-header) for detail of its spec.
2. The config for Sphinx. It surely updates the pages created in 2020, I hope that the rule isn't applied to documentation.

When you merge this PR, please squash-and-merge to keep the commit history clear. Then commits regarding the revert won't make the log dirty.

cpovirk:
Oh, nice, `$YEAR` is even better!

PR #185, opened 2021-06-05
build: Update Gradle and Java libraries

KengoTODA:
Gradle has released a new major version 7. Gradle Plugins and test libraries are also updated, so this PR suggests keep them updated.
Also, update the Sphinx to v4, I've confirmed that there is no build error/warning after the update.

_Note for reviewers:_ This PR contains a revert-commit, and its commit message is mistakenly broken ('evert' should be 'revert').
So I suggest applying squash-and-merge to this PR.


cpovirk:
Thanks as usual!

PR #186, opened 2021-06-05
Update the design-overview doc

KengoTODA:
This PR applies the [official note directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-note) to ease translation and improve the layout.

After the change, note will be rendered like below:

Screen Shot 2021-06-15 at 7 41 39

This PR also tries to translate the design-overview to Japanese partially.
I will ask some Japanese speakers to review after the 1.0.0 release, so you don't need to review this at this moment.


docs/design-overview.rst
```diff
@@ -111,7 +113,7 @@ nullability.
    -  a type of unspecified nullness where a subtype of unspecified
       nullness is required
 
--  For usability reasons, many tools will not generate generate
+-  For usability reasons, many tools will not generate
```
KengoTODA, line 116: Also removed a duplicated word here

PR #187, opened 2021-06-09
WIP: Prepare for the deployment onto Sonatype Nexus

KengoTODA:
1. sign files to meet a [requirement from Sonatype](https://central.sonatype.org/publish/requirements/gpg/)
2. deploy to Sonatype Nexus and close its staging repository automatic by [io.github.gradle-nexus.publish-plugin](https://github.com/gradle-nexus/publish-plugin/)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,4 +22,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - run: |
-          ./gradlew spotlessCheck build --no-daemon
+          ./gradlew spotlessCheck build publishToMavenLocal --no-daemon
+        env:
+          ORG_GRADLE_PROJECT_privateKey: ${{ secrets.PGP_ARMORED_PRIVATE_KEY }}
+          ORG_GRADLE_PROJECT_pgpPassword: ${{ secrets.PGP_PASS_PHREASE }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,3 @@ jobs:
             ${{ runner.os }}-gradle-
       - run: |
           ./gradlew spotlessCheck build publishToMavenLocal --no-daemon
-        env:
-          ORG_GRADLE_PROJECT_privateKey: ${{ secrets.PGP_ARMORED_PRIVATE_KEY }}
-          ORG_GRADLE_PROJECT_pgpPassword: ${{ secrets.PGP_PASS_PHREASE }}
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'java-library'
-    id 'maven-publish'
+    id 'io.github.gradle-nexus.publish-plugin' version '1.1.0' apply false
     id 'com.diffplug.spotless' version '5.12.5'
     id 'net.ltgt.errorprone' version '2.0.1'
     id 'org.javamodularity.moduleplugin' version '1.8.7'
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -1,5 +1,9 @@
 // run `./gradlew publishToMavenLocal` to install artifact to your Local Maven Repository
 
+apply plugin: 'maven-publish'
+apply plugin: 'signing'
+apply plugin: 'io.github.gradle-nexus.publish-plugin'
+
 publishing {
     publications {
         mavenJava(MavenPublication) {
@@ -23,6 +27,29 @@ publishing {
                 }
             }
         }
+    }
+}
+
+nexusPublishing {
+    repositories {
+        sonatype {
+            // //only for users registered in Sonatype after 24 Feb 2021
+            nexusUrl = uri("https://s01.oss.sonatype.org/service/local/")
+        }
+    }
+}
+
+def privateKey = findProperty('privateKey')
+def pgpPassword = findProperty('pgpPassword')
+
+signing {
+    useInMemoryPgpKeys(privateKey, pgpPassword)
+    sign publishing.publications.mavenJava
+}
+
+tasks.withType(Sign) {
+    onlyIf {
+        privateKey != null && pgpPassword != null
     }
 }
 
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -49,7 +49,7 @@ signing {
 
 tasks.withType(Sign) {
     onlyIf {
-        privateKey != null && pgpPassword != null
+        privateKey?.trim() && pgpPassword?.trim()
     }
 }
 


KengoTODA:
@cpovirk Are you familiar with [GitHub Secrets](https://docs.github.com/en/actions/reference/encrypted-secrets)?
I want to ask you or other admin of this GitHub repo, to:

1. [Generate pgp keys](https://www.gnupg.org/faq/gnupg-faq.html#starting_out) with passphrase,
2. Export ascii-armored key by `gpg --armor --export-secret-keys <EMAIL>`, 
3. Set exported key to `PGP_ARMORED_PRIVATE_KEY` GitHub Secrets variable,
4. Set passphrase to `PGP_PASS_PHREASE` GitHub Secrets variable, and
5. Rerun the build in this PR

If everything works well, we'll see that the artifact is signed with gpg automatically.

Note that this PR contains nothing that enables deployment to Sonatype Nexus at this moment.  
To enable it, we need to run the `publish` task in the build.

cpovirk:
Thanks! I didn't get a chance to do that today, but I will get to it next week.

cpovirk:
I'm sorry for not getting to this last week. I finally looked at it earlier today.

While I haven't found a direct statement from our security people about this, I have read enough that I'm nervous about uploading the PGP key because of the possibility of a GitHub compromise or other incident. Is it possible to skip signing for snapshot releases? If Sonatype requires signing even for snapshot releases, we could skip snapshots entirely: Our annotations change infrequently, so we can afford to manually push new releases after any major change. And for the final release, I assume we can run the build manually and enter the passphrase when required?

cpovirk:
(It looks like the build script already skips signing if the key is unavailable, so maybe all we'd need is a way to prompt for the passphrase during a release to Central?)

KengoTODA:
>  Is it possible to skip signing for snapshot releases?

I think it's not possible.

I've checked several snapshots at https://s01.oss.sonatype.org/content/repositories/snapshots/ and found that they has `.asc` file in their release. It means that other OSSRH users sign their artifacts before they deploy SNAPSHOT version.

KengoTODA:
> (It looks like the build script already skips signing if the key is unavailable, so maybe all we'd need is a way to prompt for the passphrase during a release to Central?)

No.

What we need to release to Central (via Sonatype Nexus) is just Sonatype user name and password.  
But even though the passphrase is needless at that timing, the artifact needs to be signed before the release process.
So we surely need passphrase to make a release.

So if we don't want to upload the GPG key to GitHub Secrets, we need to handle release procedure manually in your local.

I personally don't prefer this way, because it limits who can release and maintain the project. [SpotBugs team forked FindBugs because the owner was inactive even though he kept release privilege to himself](https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2016-November/004321.html). And each OSS project has a risk to be so.

Of course I can understand your security concerns. Recently we had a [case that leaked env var at Codecov](https://about.codecov.io/security-update/), actually.  
So if we use the manual way, I suggest you to find another and let them to have a release privilege.

cpovirk:
Sorry for dropping this again.

I had forgotten about the history of FindBugs :( We definitely do not want to repeat that.

While I remain nervous about uploading GPG keys, given my lack of security knowledge, I do want to make sure that multiple people can publish. If worst comes to worst, I'd expect that the various people who have write access to this repo can prove ownership to Sonatype through the usual process of pushing files to the repo. But there is no reason to wait for a problem to arise. I had requested a couple other owners when I filed the Sonatype ticket, but it's not clear to me that Sonatype gave permissions to anyone but me. I will post on the ticket to see if they will grant more permissions there or if we need to file new tickets.

In the meantime, this PR looks good to me. I think we could merge it as it is, and that would unblock manual deployment. If we're leaving out the GPG keys, then we could even simplify the PR slightly by omitting the `build.yml` setup for `secrets`.

cpovirk:
Some notes, which we should eventually turn into [documentation for the release process](https://github.com/jspecify/jspecify/issues/305):

The ultimate command line \[edit: after [setting the version number](https://github.com/jspecify/jspecify/commit/b2a10e14bb81c1b7cb6e488c112fe55cb2218d7d) in a detached head to be tagged later, _ideally_ in a fresh clone of the repo\] is something like this \[edit: after https://github.com/jspecify/jspecify/pull/860 changes from the plugin we used for 1.0.0 to the one we used for 1.0.1!\]:

```
JAVA_HOME=$HOME/jdk-21.0.2 ./gradlew clean publishMavenPublicationToMavenCentralRepository -Psigning.secretKeyRingFile=$HOME/.gnupg/secring.gpg -PpgpPassword="$(passphrase)" -PprivateKey="$(gpg --armor --export-secret-keys --passphrase="$(passphrase)" --batch --pinentry-mode=loopback ABCDEF12)"
```

\[edit: For the 1.0.1 release, I had also included `-PmavenCentralUsername="$(username)" -PmavenCentralPassword="$(password)"`. It's likely that this would have been unnecessary after the updated `$HOME/.gradle/gradle.properties` setup discussed below.\]

\[edit: `JAVA_HOME` should be unnecessary after https://github.com/jspecify/jspecify/pull/818 unless your default JDK is so old/new that our current version of Gradle doesn't support it. I successfully omitted it for 1.0.1.\]

\[edit: Once upon a time, we [needed `--no-configuration-cache`](https://github.com/jspecify/jspecify/pull/769). That is being solved by #860.\]

In that command, `ABCDEF12` is your "public key ID" from [these Gradle release docs](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signatory_credentials).

Those instructions also include an `--export-secret-keys` setup step. It looks like I had done that already, presumably for Guava. (Or I once used an old enough version of GPG that I didn't need to do that step?)

Another setup step: Create `$HOME/.gradle/gradle.properties` (with permissions so that no one else can read it) with your _Sonatype_ credentials \[edit: This is probably an alternative to having some configuration in `build.gradle` or a similar "Gradle file", perhaps `username`+`password` [nested under `nexusPublishing` > `repositories` > `sonatype`](https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-ossrh) at one point, and this is separate from some _other_ plugin that uses `maven` somewhere in there.\]:

```
mavenCentralUsername=<username from https://central.sonatype.org/publish/generate-portal-token/, *not* the one you use to log in>
mavenCentralPassword=<likewise from https://central.sonatype.org/publish/generate-portal-token/, *not* the one you use to log in>

# NOT ANYMORE:
#sonatypeUsername=cpovirk
#sonatypePassword=...
```

Note that the command above deploys to and closes the _staging_ repository, but it does not perform a release to Maven Central. If we want, we should be able to perform the release by changing \[edit: the command somehow or other\]. Naturally, I haven't wanted to try that yet :)

To examine the staging repository, you can \[edit: log into https://central.sonatype.com/ and go to "Publish" (near the top right) and then "Deployments."\] Again, there should be no need to look here if we \[edit: someday, maybe!\] have Gradle release the artifacts for us. But it is good to know about for testing \[edit: and for the release process as it stands today\]. Plus, if there are errors (you know, [hypothetically](https://github.com/jspecify/jspecify/pull/202) :)), those errors appear on that site under the Activity tab for a repository. \[edit: Or I assume that they show up there _somewhere_ now; the system has changed since I had problems with the 1.0.0 release, and 1.0.1 didn't hit any problems during this step.\]

General documentation for the Gradle plugin we're using is available at ~https://github.com/gradle-nexus/publish-plugin~ https://vanniktech.github.io/gradle-maven-publish-plugin/

Other notes:

If you get something wrong with the signing `-P` flags, you'll either get a build failure (usually) or a silently un-signed artifact. To see if the artifact was signed, you can look at whether the build created any `build/libs/*.asc` files. (Of course, make sure that you're not looking at `.asc` files from a previous build. This is one of the reasons I suggest `clean` in the command line.) If you get silently un-signed artifacts, the "close" step of the staging deployment should cause the build to fail, too, if you're running the command above (or any other command that performs the deployment step).

Finally, the command above passes your PGP passphrase twice as command-line arguments. I've set mine up so that the passphrase comes from a script, which can be as simple as `echo thepassphrase` or as complicated as prompting the user for the passphrase. Note that this _still_ means that the passphrase is at least temporarily visible to essentially anyone on your machine and anyone who can see the data you get from `ps`. So this is not a good idea for shared machines. The only security I'm getting out of it relative hardcoding the passphrase in the command line is (a) that I can't accidentally copy and paste the passphrase here and (b) that I don't end up with the passphrase in my Bash history, from which it might appear when others are looking over my shoulder in the future.

\[edit: We should also [update documentation for each new release](https://github.com/jspecify/jspecify/pull/857) and of course publish release notes [from a tag we push](https://github.com/jspecify/jspecify/tags). Some of us might also want to quickly adopt the new version in our libraries/tools, perhaps following on testing that we conducted beforehand.\]

\[edit: A release is probably also not a bad time to [update the copies of our annotations in our `jdk` repo](https://github.com/jspecify/jdk/pull/148).\]

cpovirk:
OK, Sonatype said there was some kind of bug that prevented others from being given permissions initially. They've now added the other people, whom I roped in from other projects where we already shared permissions. We could use some non-Google members if you (Kengo) or others are interested.

KengoTODA:
OK, thanks. I've removed GPG related process from the workflow :)

> We could use some non-Google members if you (Kengo) or others are interested.

At this moment I don't need the permission to deploy :) Thanks for your consideration!

cpovirk:
Thanks. I see that this PR is still marked as a draft, but it sounds to me like it's ready, so I'm going to merge it. If I'm wrong about that, do let me know, and we can revert it or fix forward as needed.

PR #192, opened 2021-06-24
Rewrite introduction and first section of JSpecify user guide

mernst:


docs/user-guide.md
```diff
@@ -1,23 +1,26 @@
...
+programmers to express specifications (initially, just nullness properties) in a
+machine-readable way.
+
+JSpecify defines annotations that describe whether a Java type contains the value null.
+Such annotations are useful to:
+ * programmers reading the code,
+ * tools that help developers avoid NullPointerExceptions,
+ * tools that perform run-time checking and test generation,
+ * documentation systems,
+ * and for other purposes.
```
eamonnmcmanus, line 14: Thanks, I was unhappy with this initial section and I think this rewrite is much better. My only comment would be a niggle that the last bullet point isn't grammatical ("useful to and for other purposes"). I think it just needs not to be a bullet.

docs/user-guide.md
```diff
@@ -1,23 +1,26 @@
...
 
 ## Java variables are references
 
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, any non-primitive expression may evaluate to null; for example, the
```
eamonnmcmanus, line 19: Here I'm not sure this change works, at least in isolation. The purpose of this section is partly to set the remainder of the document up, so that we can talk about references that might be null. I do think it's a good idea to introduce the notion of `T` being a subtype of `@Nullable T`. But the aim of this guide is to be relatively informal and to include readers who may not have a very precise notion of language semantics.

eamonnmcmanus: I meant to "start a review" but clicked the wrong button. :-) I'll have more to say...

eamonnmcmanus: I think it's important to keep the text with "The goal of JSpecify" (perhaps qualified since it isn't the _only_ goal). Moving towards a null-annotated world _is_ the immediate goal and readers should understand that up front. 

What would you think of adding text like this:
> Another way to look at this is that the type `@Nullable String` includes the values `{null, "", "a", ...}` (where `"a"` is a reference to a string with that value). The goal is for plain `String` to be a subtype of `@Nullable String` that excludes `null`.

kevinb9n: The current text (in red, here) was actually a part of the guide that I really liked. It seemed just right to me... can you explain what weaknesses you want to address in it?

mernst: There are two separate issues that need to be distinguished in the document:
 * What are the underlying types.  We could call them `@Nullable String` and `@NonNull String`, or `String?` and `String!`, or something else.
 * The surface syntax:  what a user writes.

There is a problem that the current text sometimes uses `String` as one of the underlying types, sometimes as what it means in Java today, and sometimes with respect to the surface syntax.  This is muddled and confusing.  The document should be unambiguous.

@eamonnmcmanus It may be a goal of certain organizations (including yours and mine!) to transition their code to null-annotated.  However, that is *not* a goal of JSpecify, whose goals are laid out in the initial text I wrote.

@kevinb There were other problems with the existing paragraph, too.  Here are a few of them.
 * It said that `int` is a variable, which is not true.
 * It said that `String` means "a reference to `String`", a self-referential definition that does not make sense.
 * It makes a big deal about the distinction between primitives and references.  Even if that distinction is true, it is irrelevant to the document and therefore it distracts readers from the important points and sets readers on the wrong foot.

kevinb9n: > There are two separate issues that need to be distinguished in the document:
> There is a problem that the current text sometimes uses `String` as one of the underlying types, sometimes as what it means in Java today, and sometimes with respect to the surface syntax. This is muddled and confusing. The document should be unambiguous.

Would it help if we:

* reserve code-font capitalized `String` when we are specifically referring to the compile-time type that *appears* exactly like that in source code
* say simply "a reference to a string" when we mean that?


> @kevinb There were other problems with the existing paragraph, too. Here are a few of them.

Acknowledged that your list is not necessarily exhaustive, and that sometimes problems are enough that a rewrite is called for. In this case, they look like spot-fixes to me:

> * It said that `int` is a variable, which is not true.

"variables, except primitives" --> "variables, except those of primitive types"


> * It said that `String` means "a reference to `String`", a self-referential definition that does not make sense.

As above I suggest that "a reference to a string" fixes this.


> * It makes a big deal about the distinction between primitives and references.  Even if that distinction is true, it is irrelevant to the document and therefore it distracts readers from the important points and sets readers on the wrong foot.

This part I don't understand. We don't have anything to say about primitive types, so it seems a lot clearer to just set those aside, and that's all I see the text doing?

kevinb9n: Note also that while _definitions_ are not allowed to be circular and must be precise, a _user guide_ has the goal of creating understanding the audience's mind, and those of us who wish to get that understanding in a very precise mathematical way (a) are unusual and (b) might want to just read the spec instead.

eamonnmcmanus: It looks as if we still have some work to do before converging for this section. If it's not too much hassle, could you make a separate PR with the rewrite of the previous section? I think everyone agrees those are a big improvement.

mernst: There is no need to nitpick every aspect of the old text.  (Except maybe to figure out why the original review missed those issues.)

We have new text, and I see no significant objections to it.  Can we merge it and move on?  A later polish pass could further improve it.

docs/user-guide.md
mernst: Agreed, it bugged me too but I wanted to get the review started.  Thanks for the comment!  I have changed it.

docs/user-guide.md
mernst: Oops, our messages crossed.  What is the specific problem with the new text?  Can you suggest an edit?

kevinb9n: Mike,

We can get through this while also setting a good precedent for how PRs will work in this project.

* Splitting the accepted part from the debated part is an all around good move - please do!
* It works better for the burden of argument to rest on the PR to demonstrate improvement, not on the author to demonstrate harm.

Thanks!

mernst: Was the original pull request subject to those rules?  If not, why the sudden change in policy?

kevinb9n: Yes, we were persuaded that the project having one user guide demonstrated improvement over it having zero user guides.

By this point, our conversation has stopped being constructive. Would the next scheduled group meeting be a good time for you to raise your broader concerns about review process? Then it won't have to feel like this cage match kind of thing - we can get a broader sense of how the group feels. Cool?

PR #193, opened 2021-06-25
Make the doc intuitive even for readers not familiar with xkcd

KengoTODA:
The word "xkcd" is [not so famous in Asia](https://trends.google.com/trends/explore/GEO_MAP/1624582800?hl=en-US&tz=-480&q=%2Fm%2F0fzd3g&sni=3), at least around me.
A hyperlink probably helps readers who don't know what it is :)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,7 +17,7 @@ Finally, it complicates interoperation between Java and other JVM languages, whi
 
 This missing artifact, with full specifications, is what we are working to create. [1]_
 
-.. [1] We have seen the xkcd comic. Please do not send us the xkcd comic. We know about the xkcd comic.
+.. [1] We have seen the `xkcd comic <https://xkcd.com/927/>`_. Please do not send us the xkcd comic. We know about the xkcd comic.
 
 Activity
 --------
--- a/docs/locale/ja/LC_MESSAGES/index.po
+++ b/docs/locale/ja/LC_MESSAGES/index.po
@@ -77,9 +77,9 @@ msgstr "この失われた成果物を正式な仕様と共に作成すること
 
 #: ../../index.rst:20
 msgid ""
-"We have seen the xkcd comic. Please do not send us the xkcd comic. We "
+"We have seen the `xkcd comic <https://xkcd.com/927/>`_. Please do not send us the xkcd comic. We "
 "know about the xkcd comic."
-msgstr "私達はxkcdを読んだことがあります。どうかあのマンガを私達に送らないでください。知ってます。"
+msgstr "私達はあの `xkcdマンガ  <https://xkcd.com/927/>`_ を読んだことがあります。どうかあのマンガを私達に送らないでください。知ってます。"
 
 #: ../../index.rst:23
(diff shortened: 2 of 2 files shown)


kevinb9n:
* I understand why sending this PR makes sense.
* The last thing I want to do is link to that comic, though, because I loathe that comic, or more precisely, I disagree with what 99% of people seem to think it means.
* I feel like the line is only there for the people who know what it means. But if they don't, it doesn't really harm them.
* If really necessary we'll just remove it. But I super-swear there are gonna be comment threads where people think it's so witty to link to that comic, and I enjoy having a little joke on them as someone points them to this line. Maybe this is my warped sense of humor.

kevinb9n:
I'll close this for now but sure, we might want to change this.

For the record: I love most of xkcd, but what this comic says is like this:

* Situation: Stick man is out of shape.
* "You know what? I'm going to work hard and try to run a 5K!"
* Situation: he failed.

How depressing is that? Who isn't aware of the fact that efforts can fail, or that they often _do_ fail?

Moreover, in our case the appropriate number in the first plan is not 14, but 0. There are literally zero competing standards, because no attempt to make a standard ever shipped.

Me, defensive?

mernst:
I think we should remove the text, as both Kengo Todo and kevinb9n say is the right thing to do.  I think that the benefit of "having a little joke on them" does not justify confusion among users.

kevinb9n:
> I think we should remove the text, as both Kengo Todo and kevinb9n say is the right thing to do.

Since your argument is to claim both of us said something that neither of us in fact said, I'm not too concerned about this right now.

It is implausible for a user to get the impression that they can't understand what JSpecify is doing if they don't know the XKCD comic, so I don't buy the claim about user confusion.

Bigger fish to fry.

kevinb9n:
Mike, I think if we spar like this, over an issue this small, we're all gonna be exhausted.

Again: yes, we may well do something about this silly little footnote at some point.

However, I am *not* doing anything in response to a discussion that is more combat than substance. That has never been the way we do things, and I'm too old to learn new tricks. Hence I am re-closing this.

> Kengo Todo asked for the text to be removed, so that is clear.

I explained that I don't see this. Repetition doesn't help; why not point to the conversation you're thinking of?

> You said "The last thing I want to do is link to that comic" and criticized it. That sounds pretty clear to me, but maybe by "The last thing I want to do" you mean something else, such as "I very much want to do it". :-)

Of the three possibilities "link", "remove", "leave alone", I stated which one was my least preference.

> > It is implausible for a user to get the impression that they can't understand what JSpecify is doing if they don't know the XKCD comic, so I don't buy the claim about user confusion.
> 
> No one made this argument. You made it up whole cloth, which is unproductive.

I was never under the impression it was an argument someone had made. I just wanted to reference the important difference between confusion that's obstructive and confusion that isn't. When I use footnotes, they are always nonessential. What I wrote that you quoted here wasn't very well put, and I'm sorry for that.

Mike, are you willing to have a two-way discussion offline about how we can work together more effectively?

Issue #34, opened 2019-06-20
@NullMarked (and its negation) to include module and package targets


kevinb9n:
Filing this so we can punt on it. [edit: haha no]

Trying to accommodate package- or module-level defaults is fraught with issues, so for now, we are going to require an annotation like `@DefaultNotNull` to appear on every class.

We know people aren't going to like this. So this issue is about continuing to try to figure something out eventually.

artempyanykh:
> I think we are pretty certain now that the answer to the question asked by the summary line of this issue "yes, absolutely". Do we need to keep this issue open?

@kevinb9n could you perhaps elaborate on this? From the discussion I see a number of issues mentioned re: package level defaulting annotations but not much about pros besides the implicit "ease of bulk-annotating classes".

cpovirk:
I think the big benefit of package-level defaults is that they make it much more difficult for library owners to forget to annotate a new class. Of course, there are ways to perform similar checks with your build system (or the proposed [`@WantDefaultForEveryClass`](https://github.com/jspecify/jspecify/issues/34#issuecomment-506916714)), but if users have to set up their build systems, many of them won't.

mernst:
I think users have to set up their build systems to run a checker.  Does that obviate the benefits you note here?  Or, could you clarify the benefits?  Thanks.

cpovirk:
I'm thinking of it this way:

1. I have already configured my build to run a checker that gives me errors for "proven dangerous" nullness conversions (like dereferencing a `@Nullable` value).
2. I want to additionally get errors if I forget to apply `@DefaultNonNull` to one of my files (because, without `@DefaultNonNull`, many dangerous operations fall into the category of "possibly dangerous" rather than "proven dangerous").

In my mind, the fundamental question of this issue is whether the same checker than I'm running for (1) is also capable of checking for (2). That "checking" could take either of two forms:

- a way to apply the default at the package/module level (the main proposal on this issue)
- a way to say, at the package/module level, that the checker should require all classes to have `@DefaultNonNull` (`@WantDefaultForEveryClass`)

(Or either of those could be configured with a command-line flag, IDE setting, or other information outside the Java sources.)

[...]

PR #194, opened 2021-06-25
Remove mention of module scope

wmdietl:
`@NullMarked` is currently only for types and packages, not modules.
See https://github.com/jspecify/jspecify/blob/da837510bcea361c41bb8eea458e8de26cbab91b/src/main/java/org/jspecify/nullness/NullMarked.java#L37 and #34.

### DIFF
--- a/src/main/java/org/jspecify/nullness/Nullable.java
+++ b/src/main/java/org/jspecify/nullness/Nullable.java
@@ -24,8 +24,8 @@ import java.lang.annotation.Target;
 
 /**
  * Indicates that the annotated type usage includes {@code null} as a value. To understand the
- * nullness of <i>unannotated</i> type usages, check for {@link NullMarked} on the enclosing class,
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks, I started the process of making a similar edit just yesterday :)

And then I untangled that, in fact, we allow `MODULE` -- albeit only for users of Java 9+, thanks to a multi-release jar. The other copy of `NullMarked` is here:
https://github.com/jspecify/jspecify/blob/da837510bcea361c41bb8eea458e8de26cbab91b/src/java9/java/org/jspecify/nullness/NullMarked.java

#34 did get left in a particularly inconclusive state (like so many of our issues...). That sounds like exactly the kind of thing that we would need to firm up during some kind of finalization push :)

Maybe we just close this and push on #34?

wmdietl:
Ah, sorry, I had missed looking at the Java 9+ version.
I'll close this.

PR #195, opened 2021-06-25
Use jspecify.org

wmdietl:
How do people feel about using `jspecify.org` for links, to be consistent with the package name?
The link will automatically redirect to `jspecify.dev`.
(Also needs to be done in other files.)

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -64,7 +64,7 @@ the following line.
 
 The first three special comments indicate that JSpecify annotations are applied
 in ways that are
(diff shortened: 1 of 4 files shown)


wmdietl:
I made the other replacements. I'm not sure why the other commits also show up in this PR. Will investigate.

wmdietl:
Changing the base branch of the PR fixed the issue and now the correct changes show up.

kevinb9n:
I agree this is the right way, and I'm hoping we can reverse the redirect so that `.org` is what people are seeing in the address bar as well.

PR #196, opened 2021-06-25
Link Nullable from NullMarked.

wmdietl:
There already is a reference from Nullable to NullMarked.
(Java 9+ version also needs this edit. Didn't reformat to highlight change.)

### DIFF
--- a/src/main/java/org/jspecify/nullness/NullMarked.java
+++ b/src/main/java/org/jspecify/nullness/NullMarked.java
@@ -26,7 +26,7 @@ import java.lang.annotation.Target;
 /**
  * Indicates that within the annotated scope (class or package), type usages
  * <i>generally</i> do <i>not</i> include {@code null} as a value, unless they are individually
- * marked otherwise.  Without this annotation, these type usages would instead have <i>unspecified
+ * marked otherwise using {@link Nullable}. Without this annotation, these type usages would instead have <i>unspecified
  * nullness</i>. Several exceptions to this rule and an explanation of unspecified nullness are
  * covered in the <a href="https://jspecify.dev/user-guide">JSpecify User Guide</a>.
  *
--- a/src/main/java/org/jspecify/nullness/NullMarked.java
+++ b/src/main/java/org/jspecify/nullness/NullMarked.java
@@ -25,8 +25,8 @@ import java.lang.annotation.Target;
 
 /**
  * Indicates that within the annotated scope (class or package), type usages <i>generally</i> do
- * <i>not</i> include {@code null} as a value, unless they are individually marked otherwise using {@link Nullable}.
- * Without this annotation, these type usages would instead have <i>unspecified nullness</i>.
+ * <i>not</i> include {@code null} as a value, unless they are individually marked otherwise using {@link Nullable} (or checker-specific other annotations).
+ * Without this annotation, unmarked type usages would instead have <i>unspecified nullness</i>.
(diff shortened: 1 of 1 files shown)


kevinb9n:
Indeed I did feel bad not having a link in this direction. It's only that I think we shouldn't restrict what the meaning of "marked otherwise" can include, because really it could be _any_ annotation that specifies nullness semantics (which might specify `@Implies(org.jspecify.nullness.Nullable)` but might not).

Maybe the fix is as simple as inserting the word "commonly"/"typically"/"presumably"/etc.?  WDYT.

wmdietl:
> Indeed I did feel bad not having a link in this direction. It was only for this reason: I think we shouldn't restrict what the meaning of "marked otherwise" can include, because really it could be _any_ annotation that specifies nullness semantics (which might specify `@Implies(org.jspecify.nullness.Nullable)` but might not).
> 
> Maybe the fix is as simple as inserting the word "commonly"/"typically"/"presumably"/etc.? WDYT.

I saw our goal to have minimal Javadoc for the 0.2.0 release. In that release, Nullable is the only option to mark something, right? (`NullnessUnspecified` won't be included in the release. There is no `Implies` yet.) 

In future releases, we'll need to expand/finalize the Javadoc and can then add other options.
But adding some word (maybe "currently"?) is also fine with me.

Let me know and I'll update and expand to the other copy.

kevinb9n:
Sorry this reply isn't helping move toward resolution yet :-(

> I saw our goal to have minimal Javadoc for the 0.2.0 release. In that release, Nullable is the only option to mark something, right?

I figured we would not want to accidentally imply that using some `@NullAlwaysOk` or `@PolyNull` or even `@other.Nullable` instead is _wrong_. 

(My argument probably means there's a slight problem with the link that goes the other way, too. <sigh>)

But it's also true that saying "typically with Nullable" or "for example with Nullable" makes readers scratch their heads and look for what else we seem to be hinting at.

Maybe something like

", unless marked otherwise (we provide @Nullable for that purpose)."

I don't know if I love that. Also the more we put here the less readers will know what "these types usages" is referring to in the following sentence, maybe.

src/main/java/org/jspecify/nullness/NullMarked.java
```diff
@@ -25,7 +25,7 @@
 
 /**
  * Indicates that within the annotated scope (class or package), type usages <i>generally</i> do
- * <i>not</i> include {@code null} as a value, unless they are individually marked otherwise.
+ * <i>not</i> include {@code null} as a value, unless they are individually marked otherwise using {@link Nullable}.
  * Without this annotation, these type usages would instead have <i>unspecified nullness</i>.
```
wmdietl, line 29:
```suggestion
 * <i>not</i> include {@code null} as a value, unless they are individually marked otherwise using {@link Nullable} (or checker-specific other annotations).
 * Without this annotation, unmarked type usages would instead have <i>unspecified nullness</i>.
```

wmdietl:
I reworded a bit and added `(or checker-specific other annotations)`.
That makes the relationship between the JSpecify annotations clear and highlights that individual checkers might add extra annotations. Is this better?

kevinb9n:
Thanks, I accept this as a step forward and will keep thinking.

kevinb9n:
Whoops, I remember now that you were planning to replicate the change on the other copy.
(Is there some kind of do-not-merge-yet label that we can use to prevent what I just did?)

wmdietl:
I'll open a new PR that reformats and replicates the change on the other copy.

PR #197, opened 2021-06-25
Generalize warning

wmdietl:
Instead of just warning about "libraries", should it be something more general, like "projects"?
Libraries are particularly affected, as others also depend on them. But the warning equally applies to somebody who just uses the annotations in their application.

### DIFF
--- a/src/main/java/org/jspecify/nullness/Nullable.java
+++ b/src/main/java/org/jspecify/nullness/Nullable.java
@@ -28,8 +28,9 @@ import java.lang.annotation.Target;
  * package, or module. See the <a href="https://jspecify.dev/user-guide">JSpecify User Guide</a> for
  * details.
  *
- * <p><b>WARNING: Do not release projects using this annotation at this time.</b> It is under
+ * <p><b>WARNING:</b> This annotation is under
  * development, and <i>any</i> aspect of its naming, location, or design may change before 1.0.
+ * <b>Do not release libraries using this annotation at this time.</b>
  */
 @Documented
 @Target(TYPE_USE)
--- a/src/main/java/org/jspecify/nullness/Nullable.java
+++ b/src/main/java/org/jspecify/nullness/Nullable.java
@@ -28,7 +28,7 @@ import java.lang.annotation.Target;
  * package, or module. See the <a href="https://jspecify.dev/user-guide">JSpecify User Guide</a> for
  * details.
  *
- * <p><b>WARNING: Do not release libraries using this annotation at this time.</b> It is under
+ * <p><b>WARNING: Do not release projects using this annotation at this time.</b> It is under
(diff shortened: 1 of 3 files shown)


kevinb9n:
Well, for me, approximately 99% of my concerns about premature usage come from the library case. An application depending too much on a 0.2.0 release of something is like a "them problem", but a library doing it becomes an "everyone problem". So it was an intentional wording choice; what do you think?

wmdietl:
I agree that "libraries" are probably the larger problem but would still warn everyone that these annotations are not finalized yet.
So how about "Do not release projects, in particular libraries, using this annotation..."?

kevinb9n:
So, what the current text is *trying* to do is to *both* "warn everyone that these annotations are not finalized yet", and also give a clear strong demand to *library* owners to never *release* dependencies on it. The trouble might just be that I made the second part the boldfaced headline, and then maybe non-libraries wouldn't read the part after it?

I estimated that released libraries cause 99% of the problem, but. I was rounding down from 99.99%. I just feel weird about a statement like "Be very careful any time you touch this gun at all, *in particular* loading it, pointing it at your foot, and puling the trigger." It's just so disproportionate.

mernst:
> So, what the current text is _trying_ to do is to _both_ "warn everyone that these annotations are not finalized yet", and also give a clear strong demand to _library_ owners to never _release_ dependencies on it.

How about just saying that?

kevinb9n:
> How about just saying that?

In my view, it literally does just say that.

We're discussing whether it says it effectively or not. It may have camouflaged one of the two points. Maybe "Warnings:" followed by items 1 and 2 is better?

I also still feel okay with changing nothing.

src/main/java/org/jspecify/nullness/Nullable.java
```diff
@@ -28,7 +28,7 @@
  * package, or module. See the <a href="https://jspecify.dev/user-guide">JSpecify User Guide</a> for
  * details.
  *
- * <p><b>WARNING: Do not release libraries using this annotation at this time.</b> It is under
+ * <p><b>WARNING: Do not release projects using this annotation at this time.</b> It is under
  * development, and <i>any</i> aspect of its naming, location, or design may change before 1.0.
```
wmdietl, line 32:
```suggestion
 * <p><b>WARNING:</b> This annotation is under
 * development, and <i>any</i> aspect of its naming, location, or design may change before 1.0.
 * <b>Do not release libraries using this annotation at this time.</b>
```

wmdietl:
I've reordered the warning.
I'm fine with just closing this PR if everyone prefers the current warning.

kevinb9n:
Love it. Thanks for bearing with this conversation.

kevinb9n (approved):

kevinb9n:
(please merge if the checks are happy)

PR #201, opened 2021-07-02
Comments on user guide

wmdietl:
DO NOT MERGE. Marker edits for discussion items.

I've made a pass through the User Guide and only have a few small comments. I'll add more comments to some edits.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
 
 ## Types and nullness
 
-In the JSpecify world, most references have one of three possible properties
+In the JSpecify world, most TODO references have one of three possible properties
 regarding nullness:
 1. They can be null.
 2. They can't be null.
 3. We don't know if they can be null or not.
 
-If `x` is a reference of a certain type, then if `x` can be null, `x.toString()` is
+If `x` is a reference of a certain type, then if `x` can be null, the dereference `x.toString()` is
 unsafe because it could produce a NullPointerException. If `x` can't be null,
 `x.toString()` can never produce a NullPointerException. And if we don't know
 whether `x` can be null or not, `x.toString()` might or might not be unsafe.
 
 There are two JSpecify annotations that indicate these properties:
 * `@Nullable` applied to a type denotes a reference of that type that can be null.
-* `@NullMarked` applied to a module, class, or package means that a reference in
+* `@NullMarked` applied to a class, package, or module means that a reference in
 that scope can't be null unless its type is explicitly marked `@Nullable`. (Below
 we will see that there are some exceptions to this for [local
 variables](#local-variables) and [type variables](#defining-generics).)
 
 If a type is not explicitly marked `@Nullable` and is not inside a `@NullMarked`
-class or package, then we don't know if references of that type can be null or
+class, package, or module, then we don't know if references of that type can be null or
 not.
 
 The notion of "can't be null" should really be read with a footnote that says
@@ -59,7 +59,7 @@ includes references that can be null. Code that deals with those references must
 be able to deal with the null case.
 
 ```java
-  public static void print(@Nullable String x) {
+  void print(@Nullable String x) {
     System.out.println(String.valueOf(x));
   }
 ```
@@ -96,6 +96,7 @@ The `@NullMarked` annotation indicates that references can't be null in its scop
 unless their types are explicitly marked `@Nullable`. If applied to a package then
 its scope is all the code in the package. If applied to a class or interface
 then its scope is all the code in that class or interface.
+TODO: mention modules. First use of `interface`, before `class` subsumed all type declarations.
 
 ```java
 @NullMarked
@@ -114,8 +115,9 @@ In this example, both methods are in the scope of `@NullMarked`, so plain `Strin
 means "a reference to a `String` that can't be null". `@Nullable String` continues
 to mean "a reference to a `String` that can be null". Tools should warn you if you
 try to pass a "reference to a `String` that can be null" to `spaceIndex`, since its
-argument can't be null, and indeed it will throw NullPointerException if given a
+argument can't be null, and indeed it will throw a NullPointerException if given a
 null argument.
+TODO: warning only expected if call is also in `@NullMarked` context, per earlier comment?
 
 As mentioned above, there are some exceptions to this interpretation for local
 variables (as we'll see next) and [type variables](#defining-generics).
@@ -131,8 +133,8 @@ the values that are assigned to the variable. For example:
 @NullMarked
 public class MyClass {
   void myMethod(@Nullable String one, String two) {
-    String copyOfOne = one;
-    String copyOfTwo = two;
+    String aliasOfOne = one;
+    String aliasOfTwo = two;
     String oneOrTwo = random() ? one : two;
     String twoOrNull = Strings.emptyToNull(two);
     ...
@@ -140,10 +142,10 @@ public class MyClass {
 }
 ```
 
-Analysis can tell that all of these variables except `copyOfTwo` can be null.
-`copyOfTwo` can't be null since `two` can't be null: it is not `@Nullable` and it is
-inside the scope of `@NullMarked`. `copyOfOne` can be null since it is a copy of a
-`@Nullable` parameter. `oneOrTwo` can be null because it may be a copy of a
+Analysis can tell that all of these variables except `aliasOfTwo` can be null.
+`aliasOfTwo` can't be null since `two` can't be null: it is not `@Nullable` and it is
+inside the scope of `@NullMarked`. `aliasOfOne` can be null since it is an alias of a
+`@Nullable` parameter. `oneOrTwo` can be null because it may be an alias of a
 `@Nullable` parameter. And `twoOrNull` can be null because its value comes from a
 method that returns `@Nullable String`.
 
@@ -198,7 +200,7 @@ public interface List<E extends @Nullable Object> {...}
 
 If it were `interface List<E>` rather than `interface List<E extends @Nullable
 Object>` then `NumberList<E extends @Nullable Number> extends List<E>` would not be
-legal. That's because `List<E>` is short for `List<E extends Object>`. Inside
+legal. That's because `interface List<E>` is short for `interface List<E extends Object>`. Inside
 `@NullMarked`, plain `Object` means "`Object` reference that can't be null". The `<E
 extends @Nullable Number>` from NumberList would not be compatible with that.
 
@@ -341,6 +343,7 @@ won't need to know.
 
 There are a couple of places where the syntax of type annotations like `@Nullable`
 is counterintuitive.
+TODO: make the principle clear: type use annotation immediately before the annotated type.
 
 1. For a nested type like `Map.Entry`, if you want to say that the value can be null
 then the syntax is `Map.@Nullable Entry`. You can often avoid dealing with this by
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -24,7 +24,7 @@ that might be null then you write `@Nullable String x;`.
 
 ## Types and nullness
 
-In the JSpecify world, most TODO references have one of three possible properties
+In the JSpecify world, references have one of three possible properties
 regarding nullness:
 1. They can be null.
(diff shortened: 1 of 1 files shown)


docs/user-guide.md
```diff
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
 
 ## Types and nullness
 
-In the JSpecify world, most references have one of three possible properties
+In the JSpecify world, most TODO references have one of three possible properties
```
wmdietl, line 27: The `most` is never explained. Maybe a forward reference here, or a backward reference from generics that explains the "most" aspect would be nice.

docs/user-guide.md
```diff
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
...
 regarding nullness:
 1. They can be null.
 2. They can't be null.
 3. We don't know if they can be null or not.
 
-If `x` is a reference of a certain type, then if `x` can be null, `x.toString()` is
+If `x` is a reference of a certain type, then if `x` can be null, the dereference `x.toString()` is
```
wmdietl, line 33: Maybe this is too nit-picky, but the guarantee is only about the dereference of `x` itself, not about what the method itself might do.

docs/user-guide.md
```diff
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
...
 unsafe because it could produce a NullPointerException. If `x` can't be null,
 `x.toString()` can never produce a NullPointerException. And if we don't know
 whether `x` can be null or not, `x.toString()` might or might not be unsafe.
 
 There are two JSpecify annotations that indicate these properties:
 * `@Nullable` applied to a type denotes a reference of that type that can be null.
-* `@NullMarked` applied to a module, class, or package means that a reference in
+* `@NullMarked` applied to a class, package, or module means that a reference in
```
wmdietl, line 40: Reordered to be consistent.

docs/user-guide.md
```diff
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
...
 that scope can't be null unless its type is explicitly marked `@Nullable`. (Below
 we will see that there are some exceptions to this for [local
 variables](#local-variables) and [type variables](#defining-generics).)
 
 If a type is not explicitly marked `@Nullable` and is not inside a `@NullMarked`
-class or package, then we don't know if references of that type can be null or
+class, package, or module, then we don't know if references of that type can be null or
```
wmdietl, line 46: Module was missing here and order inconsistent.

docs/user-guide.md
```diff
@@ -59,7 +59,7 @@ includes references that can be null. Code that deals with those references must
 be able to deal with the null case.
 
 ```java
-  public static void print(@Nullable String x) {
+  void print(@Nullable String x) {
```
wmdietl, line 62: I'm wondering whether we should remove some of the boilerplate that isn't strictly needed.
Some methods are later called, so e.g. they would need to remain static. But maybe this could be simplified.

docs/user-guide.md
```diff
@@ -96,6 +96,7 @@ The `@NullMarked` annotation indicates that references can't be null in its scop
 unless their types are explicitly marked `@Nullable`. If applied to a package then
 its scope is all the code in the package. If applied to a class or interface
 then its scope is all the code in that class or interface.
+TODO: mention modules. First use of `interface`, before `class` subsumed all type declarations.
```
wmdietl, line 99: Maybe just say package/module to also cover modules.
The extra mention of `interface` here is a bit surprising, as before (and according to our glossary), `class` covers all type declarations.

docs/user-guide.md
```diff
@@ -114,8 +115,9 @@ In this example, both methods are in the scope of `@NullMarked`, so plain `Strin
 means "a reference to a `String` that can't be null". `@Nullable String` continues
 to mean "a reference to a `String` that can be null". Tools should warn you if you
 try to pass a "reference to a `String` that can be null" to `spaceIndex`, since its
-argument can't be null, and indeed it will throw NullPointerException if given a
+argument can't be null, and indeed it will throw a NullPointerException if given a
 null argument.
+TODO: warning only expected if call is also in `@NullMarked` context, per earlier comment?
```
wmdietl, line 120: An earlier comment (at the end of `Types and nullness`) called out calls from non-NullMarked code. So for the hypothetical call of `spaceIndex`, we could make the two cases explicit here.

docs/user-guide.md
```diff
@@ -131,19 +133,19 @@ the values that are assigned to the variable. For example:
 @NullMarked
 public class MyClass {
   void myMethod(@Nullable String one, String two) {
-    String copyOfOne = one;
-    String copyOfTwo = two;
+    String aliasOfOne = one;
```
wmdietl, line 136: Nothing is copied in these examples, so `alias` seems like a better term.

docs/user-guide.md
```diff
@@ -198,7 +200,7 @@ public interface List<E extends @Nullable Object> {...}
 
 If it were `interface List<E>` rather than `interface List<E extends @Nullable
 Object>` then `NumberList<E extends @Nullable Number> extends List<E>` would not be
-legal. That's because `List<E>` is short for `List<E extends Object>`. Inside
+legal. That's because `interface List<E>` is short for `interface List<E extends Object>`. Inside
```
wmdietl, line 203: On first reading, it wasn't clear whether this refers to the `extends List<E>` part. Adding `interface` might be one option to fix this. Or instead of reusing `E` we could use different type variable names.

docs/user-guide.md
```diff
@@ -341,6 +343,7 @@ won't need to know.
 
 There are a couple of places where the syntax of type annotations like `@Nullable`
 is counterintuitive.
+TODO: make the principle clear: type use annotation immediately before the annotated type.
```
wmdietl, line 346: I'm not sure I agree with the `counterintuitive` part... we need to make the principle clear: a type use annotation is immediately before/to the left of the annotated type. Then the two examples should be clear.

docs/user-guide.md
```diff
@@ -341,6 +343,7 @@ won't need to know.
...
 
 1. For a nested type like `Map.Entry`, if you want to say that the value can be null
```
wmdietl, line 348: Actually for the array case, but I can't comment there: Should we call out the difference between declaration and type use annotations here? That is a frequent point of confusion.

docs/user-guide.md
```diff
@@ -341,6 +343,7 @@ won't need to know.
```
wmdietl, line 343: I think `Type annotation` hasn't been introduced before, but maybe that's intuitive enough.
And maybe "Type use annotations` to be more precise?

docs/user-guide.md
kevinb9n: (How would standardizing it on the opposite order be? The wider the scope the better.... e.g. I think we may be adding "method", but I wouldn't want it to be highly emphasized.)

docs/user-guide.md
kevinb9n: or maybe just `anotherOne`

docs/user-guide.md
kevinb9n: That's a good point for us to emphasize.

The counterintuitiveness was there already, because `String[]` (array of string) is backwards from `List<String>` (list of string), and because in `Map.Entry`, despite completely resembling `OuterType.InnerType`, `Map` is only a name qualifier and not a type usage (and only checking for `static` at the nested declaration would tell you).

docs/user-guide.md
kevinb9n: Sadly we can never get away with saying "type annotation" since ElementType uses `TYPE` (completely improperly!) just to mean `CLASS`.

I would prefer "type-usage annotation" even though it doesn't roll of the tongue... because "type use" is just wrong. (An example "use" of `boolean` is "to indicate whether a mutating method actually mutated", and an example "usage" of `boolean` is "as the return type of `Collection.add`".)

docs/user-guide.md
kevinb9n: Yeah... saying anything about the exceptions would be a harmful distraction right now. Can we say "all" but then "(We'll cover a few exceptions later.)"?

docs/user-guide.md
eamonnmcmanus: I have to say that it's not clear to me that "immediately to the left of the annotated type" helps explain the placement of `@Nullable` in `String @Nullable []`. `[]` isn't a type. Things get even worse with multi-dimensional arrays. I'm not sure there's a simple principle to state here, and I think just showing examples remains the way to go.

docs/user-guide.md
```diff
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
 
 ## Types and nullness
 
-In the JSpecify world, most references have one of three possible properties
+In the JSpecify world, references have one of three possible properties
```
wmdietl, line 27: Follow-up to https://github.com/jspecify/jspecify/pull/201#discussion_r663097830

Maybe not saying anything works well enough. The other case we're concerned about is "parametric nullness", which we'll discuss later with generics/wildcards. But one could see that as another case where we don't know whether it can be null or not (if the bound is nullable) or that they can't be null (if the bound is non-null).
Or am I missing what other case "most" was referring to?

docs/user-guide.md
```diff
@@ -24,26 +24,26 @@ that might be null then you write `@Nullable String x;`.
...
-If `x` is a reference of a certain type, then if `x` can be null, `x.toString()` is
+If `x` is a reference of a certain type, then if `x` can be null, the dereference `x.toString()` is
 unsafe because it could produce a NullPointerException. If `x` can't be null,
 `x.toString()` can never produce a NullPointerException. And if we don't know
 whether `x` can be null or not, `x.toString()` might or might not be unsafe.
 
 There are two JSpecify annotations that indicate these properties:
 * `@Nullable` applied to a type denotes a reference of that type that can be null.
-* `@NullMarked` applied to a module, class, or package means that a reference in
+* `@NullMarked` applied to a module, package, or class means that a reference in
```
wmdietl, line 40: Follow-up to https://github.com/jspecify/jspecify/pull/201#discussion_r663064798

I switched the order in this document. Should we do the same throughout the javadoc, e.g. here: https://github.com/jspecify/jspecify/blob/8797899cff85804cd3cf049854e498805b122104/src/java9/java/org/jspecify/nullness/NullMarked.java#L28

wmdietl (commented):
I've addressed a few of the comments.
Do let me know if I should break out any of these changes into smaller PRs or if #201 will address those in separate PRs.

docs/user-guide.md
kevinb9n: I think that'll get fixed naturally when we come by and add "method" (which I seem to recall looking like what we want to do, when last I saw the issue).

docs/user-guide.md
eamonnmcmanus: I propose to change the example to `x.getClass()`. I think it will be clearer then that the only thing that can provoke NPE is if `x` is null.

docs/user-guide.md
eamonnmcmanus: OK, we can certainly remove many `public` and/or `static` modifiers.

docs/user-guide.md
eamonnmcmanus: The annotations are called `TYPE_USE` in the `java.lang.annotation` API so I think we can reasonably call them "type-use annotations" and be understood. I actually think "type-usage annotations" would be slightly harder to understand.

docs/user-guide.md
eamonnmcmanus: I actually don't remember why I wrote "most" here and I think it can reasonably be omitted.

docs/user-guide.md
kevinb9n: I can live with that.

kevinb9n:
What's the path forward for this PR? Is #207 addressing some of the same issues?

eamonnmcmanus:
> What's the path forward for this PR? Is #207 addressing some of the same issues?

I tried to incorporate everything from this PR in #207. Let me know if I missed anything.

kevinb9n:
Ok I'm gonna proactively close this, but Werner, feel welcome to re-raise any things that were missed.

PR #202, opened 2021-07-02
Add a `developer` to the project metadata.

cpovirk:
Instructions:
https://docs.gradle.org/current/userguide/publishing_maven.html#sec:modifying_the_generated_pom

Without this, we can't close the Sonatype repository: We get the error
"Developer information missing."

It's not clear who exactly we should list here. Probably we'd prefer
that people contact us on GitHub or our mailing list (if we've opened up
posting permission to everyone). For the moment, I've added only kevinb,
since he has plans to do some outreach to the wider community. If anyone
would like to be added, feel free to add yourself.

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -27,6 +27,13 @@ publishing {
                         name = 'Kevin Bourrillion'
                         email = 'kevinb@google.com'
                     }
+                    developer {
+                        id = 'KengoTODA'
+                        name = 'Kengo TODA'
+                        email = 'skypencil@gmail.com'
+                        url = 'https://github.com/KengoTODA/'
+                        timezone = '+8'
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


kevinb9n:
Oh dang, I did look at this and I went down a rabbit-hole of trying to understand what this is supposed to mean, what the requirement is, etc., and I drifted away.

(And I still don't know how "make sure no jspecify PRs are waiting on me" fits into my daily workflow.)

Could we have a comment like

// These are here only because we have to have *someone*.
// Any project member is welcome to add themselves if they want to.

cpovirk:
Thanks, Kengo and Kevin. This is ready for review again.

kevinb9n:
Sorry for the delay, I don't yet know how to integrate "check for PRs waiting on me" into my daily workflow.

Issue #203, opened 2021-07-07
User Guide improvements


eamonnmcmanus:
The initial draft of the User Guide has a number of issues that have been pointed out. I'm trying to summarize the points made in #192 and #201, as well as discussion elsewhere.

* The guide talks about "unspecified nullness", but really it should say that the rules it describes apply to code within `@NullMarked` and that how that code interacts with non-`@NullMarked` code is unspecified. Maybe also saying that individual tools will typically document their own rules for that interaction.

* In an attempt to stay informal, it uses (for example) `String` to mean a number of different things, such as the literal text `String` in a program, the type `String` as opposed to `@Nullable String`, the type of an object that a `String` variable references, etc. We should try to find ways to be more precise without making the guide too formal.

* It should state explicitly early on that `String` is a subtype of `@Nullable String` and what that means.

* It's inconsistent about the list of things that can be `@NullMarked`. It should say "module, package, or class" everywhere.

* The method examples include variables like `copyOfOne` that are just copies of _references_. Another name might be better, such as `aliasOfOne` or `anotherOne`.

* It should be clearer about the distinction between declaration and type-use annotations. It should probably avoid talking about "type annotations" when it specifically means type-use annotations. Possibly "type-usage annotations" for maximal accuracy.

There are also a number of more minor concerns mentioned in those PRs, which I think we can just go ahead and fix.

There are probably other concerns too, which people should feel free to raise either here or in separate GItHub issues.

Issue #39, opened 2019-06-24
@NotNull field semantics


donnerpeter:
Basic question: when a field is marked as `@NotNull`, when can it be safely assumed to be non-null?

If a field is final, then it probably becomes notnull after its initialization. But before that, other methods can be called, also via superclass constructor, and it seems an undecidable problem in general to determine if a specific field reference is done after its initialization or before it. IntelliJ does some guessing here to avoid marking `if (field == null)` as "always true" in such semi-initialized places, but to avoid false positives, it'd also skip useful warnings sometimes.

If a field is non-final, things get even more complicated, not even compiler complains about obvious field dereferences before its initialization. Should we even allow notnull fields not initialized before or inside all constructors? AFAIK Eclipse doesn't. IntelliJ has an option for this, and some people don't like it because they have non-null fields initialized in test set-up, or by dependency injection.

kevinb9n:
Above I said, "Non-final fields are the source of most problems" and I fantasized about handwaving them away. But yeah, maybe it's not so bad.

There's no problem with `public @Nullable Type field`; you can assign anything to it, but any read from it is nullable.

As for `public @NonNull Type field`, first, this would prevent assigning a nullable type to it, which is all good. For reads, it documents the owner's promise that no one outside the class or its subclasses should ever observe a null value. The owner could fail to meet that promise, but it can also fail to meet its promise for a nullable method return type or anything else too.

kevinb9n:
Here's a broader proposal for how we might look at non-final fields.

First, we could stick to my beloved "all we're doing is making enhanced types" design anchor, but in this case it doesn't get us far. Fields go through changes; types are unchangeable. Once a field gets marked nullable then a checker has little opportunity left to do anything smart about it.

So: class `Foo` has non-final field `bar`.

First, I want to talk about `Foo`'s *responsibilities* toward `bar`, *then* get into "how might a tool *help* `Foo`'s author get those responsibilities right."

1. `Foo` is responsible for *initializing* `bar` (making it able to serve its purpose) _sometime_ before any code outside the class might read its value, directly or indirectly (with a few more coping strategies available if the latter).
2. If `Foo` wants to *invalidate* `bar` (making it unable to serve its purpose), then `Foo` is likewise responsible for ensuring that outside callers won't access it anymore, directly or indirectly (with a few more coping strategies available if the latter).

None of this is new or specific to nullness checking, it's just a way to write less-buggy code in general.

[...]

PR #207, opened 2021-07-12
User guide changes to address the issues in #203.

eamonnmcmanus:
Several other issues have also been addressed. Here the main changes:

* Avoided using `String` to mean both a reference of that type and an object of that type.
  Now we say "variable of type `String`" and we say "string object" when we mean an object of that type.

* Reworded text that was implicitly talking about unspecified nullness so that it explicitly
  says that JSpecify doesn't say anything about the possible nullness of a non-`@Nullable` type
  outside `@NullMarked`.

* Removed claim that null-annotating everything is a "goal" of JSpecify in favour of saying that
  JSpecify can help with it.

* Explicitly introduced the notion that `String` is a subtype of `@Nullable String` early on.

* Removed some `public` and `static` keywords from examples where they were basically noise.

* Changed an example from `x.toString()` to `x.getClass()` to make it more obvious that the
  NPE we are talking about is one that comes from dereferencing `x`, not from something its
  `toString()` method might do.

* Explicitly listed module, package, and class, in every case and in that order, as the things
  that `@NullMarked` can apply to.

* Renamed `copyOfOne` etc to `anotherOne` to avoid confusion about what "copy" might mean.

* Spelled out `interface List<E>` rather than just `List<E>` at a point where it might
  otherwise not have been clear whether we were talking about a definition or a use of `E`.

* Revisited the section on type-use annotation syntax to reflect comments received.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -104,12 +104,11 @@ might be null).
 ```java
 @NullMarked
 class Example {
-  private String nonNull;
-  private @Nullable String nullable;
-
-  void example() {
-	nullable = nonNull; // JSpecify allows this
-	nonNull = nullable; // JSpecify doesn't allow this
+  void useNullable(@Nullable String x) {...}
+  void useNonNull(String x) {...}
+  void example(@Nullable String nullable, String nonNull) {
+	useNullable(nonNull); // JSpecify allows this
+	useNonNull(nullable); // JSpecify doesn't allow this
   }
 }
 ```
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -107,8 +107,8 @@ class Example {
   void useNullable(@Nullable String x) {...}
   void useNonNull(String x) {...}
   void example(@Nullable String nullable, String nonNull) {
-	useNullable(nonNull); // JSpecify allows this
-	useNonNull(nullable); // JSpecify doesn't allow this
+    useNullable(nonNull); // JSpecify allows this
+    useNonNull(nullable); // JSpecify doesn't allow this
   }
 }
 ```
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -17,10 +17,11 @@ Such annotations are useful to (for example):
 
 In Java, all non-primitive variables are references. We often think of `String
 x;` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_
-to an object of type `String`". One important difference between these two views
-is that `x` can also be null.  JSpecify can help move towards a world where
-`String x;` means "`x` is a reference to an actual string", and if you want a
-reference that might be null then you write `@Nullable String x;`.
+to a string or it is null". JSpecify includes a notion of "null-marked code"
+where `String x;` means "`x` is a reference to an actual string", and if you
+want a reference that might be null then you write `@Nullable String
+x;`. Null-marked code is simply code that is covered by the `@NullMarked`
+annotation.
 
 ## Types and nullness
 
@@ -88,7 +89,7 @@ Tools could then use the `@Nullable` information to determine that the first use
 is safe but the second is not.
 
 As far as JSpecify is concerned, `String` and `@Nullable String` are _different_
-types. A variable of type `String` can reference any non-null string. A variable
+types. A variable of type `String` can reference any string. A variable
 of type `@Nullable String` can too, but it can also be null. This means that
 `String` is a _subtype_ of `@Nullable String`, in the same way that `Integer` is
 a subtype of `Number`. One way to look at this is that a subtype narrows the
@@ -134,9 +135,9 @@ public class Strings {
 ```
 
 In this example, both methods are in the scope of `@NullMarked`, so plain `String`
-means "a string reference that can't be null". `@Nullable String` continues
-to mean "a string reference that can be null". Tools should warn you if you
-try to pass a "string reference that can be null" to `spaceIndex`, since its
+means "a reference to a string, not null". `@Nullable String` continues
+to mean "a reference to a string, or null". Tools should warn you if you
+try to pass a "reference to a string, or null" to `spaceIndex`, since its
 argument can't be null, and indeed it will throw NullPointerException if given a
 null argument.
 
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -101,12 +101,15 @@ other subtype). In the same way, a `@Nullable String` can be assigned from a
 might be null).
 
 ```java
-void good(String nonNull, @Nullable String nullable) {
-  nullable = nonNull; // JSpecify allows this
-}
+@NullMarked
+class Example {
+  private String nonNull;
+  private @Nullable String nullable;
 
-void bad(String nonNull, @Nullable String nullable) {
-  nonNull = nullable; // JSpecify doesn't allow this
+  void example() {
+	nullable = nonNull; // JSpecify allows this
+	nonNull = nullable; // JSpecify doesn't allow this
+  }
 }
 ```
 
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -16,29 +16,30 @@ Such annotations are useful to (for example):
 ## Java variables are references
 
 In Java, all non-primitive variables are references. We often think of `String
-x;` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_
-to a string or it is null". JSpecify includes a notion of "null-marked code"
-where `String x;` means "`x` is a reference to an actual string", and if you
-want a reference that might be null then you write `@Nullable String
-x;`. Null-marked code is simply code that is covered by the `@NullMarked`
-annotation.
+x` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_,
+either null or a reference to a string".
+
+JSpecify includes a `@NullMarked` annotation. In code covered by that
+annotation, `String x` means "`x` is a reference to an actual string", and
+`@Nullable String x` means "`x` is either null or a reference to an actual
+string."
 
 ## Types and nullness
 
-In the JSpecify world, references have one of three possible properties
-regarding nullness:
-1. They can be null.
-2. They can't be null.
-3. JSpecify annotations haven't indicated whether they can be null.
+Each reference can have one of three possible properties regarding nullness:
 
-If `x` is a reference of a certain type, then if `x` can be null, `x.getClass()`
-is unsafe because it could produce a NullPointerException. If `x` can't be null,
-`x.getClass()` can never produce a NullPointerException. If JSpecify annotations
-haven't said whether `x` can be null or not, we don't know whether `x.getClass()`
-is safe (at least as far as JSpecify is concerned).
+1. JSpecify annotations indicate that it can be null.
+2. JSpecify annotations indicate that it can't be null.
+3. JSpecify annotations don't indicate whether it can be null.
+
+For a given reference `x`, if `x` can be null then `x.getClass()` is unsafe because
+it could produce a NullPointerException. If `x` can't be null, `x.getClass()`
+can never produce a NullPointerException. If JSpecify annotations haven't said
+whether `x` can be null or not, we don't know whether `x.getClass()` is safe (at
+least as far as JSpecify is concerned).
 
 There are two JSpecify annotations that indicate these properties:
-* `@Nullable` applied to a type denotes a reference of that type that can be null.
+* `@Nullable` applied to a type means a reference of that type that can be null.
 * `@NullMarked` applied to a module, package, or class means that a reference in
 that scope can't be null unless its type is explicitly marked `@Nullable`. (Below
 we will see that there are some exceptions to this for [local
@@ -96,7 +97,7 @@ a subtype of `Number`. One way to look at this is that a subtype narrows the
 range of possibilities. A `Number` variable can be assigned from an `Integer`
 but it can also be assigned from a `Long`. Meanwhile an `Integer` variable
 can't be assigned from a `Number` (since that `Number` might be a `Long` or some
-other subtype). In the same way, a `@Nullable String` can be assigned from a
+other subtype). Likewise, a `@Nullable String` can be assigned from a
 `String` but a `String` can't be assigned from a `@Nullable String` (since that
 might be null).
 
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -15,36 +15,33 @@ Such annotations are useful to (for example):
 
 ## Java variables are references
 
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, all non-primitive variables are references. We often think of `String
+x;` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_
+to an object of type `String`". One important difference between these two views
+is that `x` can also be null.  JSpecify can help move towards a world where
+`String x;` means "`x` is a reference to an actual string", and if you want a
+reference that might be null then you write `@Nullable String x;`.
 
 ## Types and nullness
 
-In the JSpecify world, most references have one of three possible properties
+In the JSpecify world, references have one of three possible properties
 regarding nullness:
 1. They can be null.
 2. They can't be null.
-3. We don't know if they can be null or not.
+3. JSpecify annotations haven't indicated whether they can be null.
 
-If `x` is a reference of a certain type, then if `x` can be null, `x.toString()` is
-unsafe because it could produce a NullPointerException. If `x` can't be null,
-`x.toString()` can never produce a NullPointerException. And if we don't know
-whether `x` can be null or not, `x.toString()` might or might not be unsafe.
+If `x` is a reference of a certain type, then if `x` can be null, `x.getClass()`
+is unsafe because it could produce a NullPointerException. If `x` can't be null,
+`x.getClass()` can never produce a NullPointerException. If JSpecify annotations
+haven't said whether `x` can be null or not, we don't know whether `x.getClass()`
+is safe (at least as far as JSpecify is concerned).
 
 There are two JSpecify annotations that indicate these properties:
 * `@Nullable` applied to a type denotes a reference of that type that can be null.
-* `@NullMarked` applied to a module, class, or package means that a reference in
+* `@NullMarked` applied to a module, package, or class means that a reference in
 that scope can't be null unless its type is explicitly marked `@Nullable`. (Below
 we will see that there are some exceptions to this for [local
 variables](#local-variables) and [type variables](#defining-generics).)
-
-If a type is not explicitly marked `@Nullable` and is not inside a `@NullMarked`
-class or package, then we don't know if references of that type can be null or
-not.
 
 The notion of "can't be null" should really be read with a footnote that says
 "if all the code in question is `@NullMarked`". For example, if you have some code
@@ -59,7 +56,7 @@ includes references that can be null. Code that deals with those references must
 be able to deal with the null case.
 
 ```java
-  public static void print(@Nullable String x) {
+  static void print(@Nullable String x) {
     System.out.println(String.valueOf(x));
   }
 ```
@@ -69,7 +66,7 @@ call. The body of the `print` method does not do anything with `x` that would
 provoke a NullPointerException so this method is safe.
 
 ```java
-  public static @Nullable String emptyToNull(@Nullable String x) {
+  static @Nullable String emptyToNull(@Nullable String x) {
     return (x == null || x.isEmpty()) ? null : x;
   }
 ```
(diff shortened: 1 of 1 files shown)


docs/user-guide.md
```diff
@@ -111,9 +134,9 @@ public class Strings {
 ```
 
 In this example, both methods are in the scope of `@NullMarked`, so plain `String`
-means "a reference to a `String` that can't be null". `@Nullable String` continues
-to mean "a reference to a `String` that can be null". Tools should warn you if you
-try to pass a "reference to a `String` that can be null" to `spaceIndex`, since its
+means "a string reference that can't be null". `@Nullable String` continues
```
kevinb9n, line 137: Here again, I really don't like condoning the idea that null IS a "string reference" at all.

kevinb9n (changes requested):

docs/user-guide.md
```diff
@@ -90,13 +87,39 @@ be too. You might use this method like this:
...
+range of possibilities. A `Number` variable can be assigned from an `Integer`
+but it can also be assigned from a `Long`. Meanwhile an `Integer` variable
+can't be assigned from a `Number` (since that `Number` might be a `Long` or some
+other subtype). In the same way, a `@Nullable String` can be assigned from a
+`String` but a `String` can't be assigned from a `@Nullable String` (since that
+might be null).
+
+```java
+void good(String nonNull, @Nullable String nullable) {
+  nullable = nonNull; // JSpecify allows this
```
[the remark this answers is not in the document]
eamonnmcmanus: We can't, because tools don't have to allow `@Nullable` on local variables.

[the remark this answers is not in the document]
kevinb9n: Ugh sorry. As everyone here knows by now, sometimes my brain just does not work.

Let me start over.

Maybe a cleaner way to handle this is to show one method that returns nullable, and one that accepts non-nullable. Then calling code that passes the result of the first method into the second.
Did I brain well this time?

There is also this lingering trouble of "JSpecify allows".  What we want to say is "Presumably any tool worth the bytes it occupies would highlight a problem here", but that is not a very ergonomic phrase.

docs/user-guide.md
```diff
@@ -15,37 +15,34 @@ Such annotations are useful to (for example):
...
 ## Java variables are references
 
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, all non-primitive variables are references. We often think of `String
+x;` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_
```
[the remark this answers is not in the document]
eamonnmcmanus: OK. I did actually take note of your suggestion to use "string" rather than "object of type `String`" everywhere else but I left this one in. But I agree that we can be more precise.

docs/user-guide.md
```diff
@@ -15,37 +15,34 @@ Such annotations are useful to (for example):
...
+to an object of type `String`". One important difference between these two views
+is that `x` can also be null.  JSpecify can help move towards a world where
+`String x;` means "`x` is a reference to an actual string", and if you want a
+reference that might be null then you write `@Nullable String x;`.
 
 ## Types and nullness
 
-In the JSpecify world, most references have one of three possible properties
+In the JSpecify world, references have one of three possible properties
```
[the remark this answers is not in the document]
eamonnmcmanus: OK, I'll reword to the best of my understanding. I think your key point is that some of the existing wording seems to be talking about references to null, when it should be saying null references.

Regarding variables, I think it is easier for readers to understand the concrete notion of a variable that has a type, and that they will naturally extend that to other type uses. Talking about types directly risks being a bit too abstract. But yes, if you want to tackle that later we can discuss it then.

docs/user-guide.md
eamonnmcmanus: OK, will reword.

docs/user-guide.md
```diff
@@ -15,37 +15,34 @@ Such annotations are useful to (for example):
...
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, all non-primitive variables are references. We often think of `String
+x;` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_
+to an object of type `String`". One important difference between these two views
+is that `x` can also be null.  JSpecify can help move towards a world where
```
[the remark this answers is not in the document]
eamonnmcmanus: I think we can introduce the idea of "null-marked code", which then is obviously code that is `@NullMarked`.

docs/user-guide.md
```diff
@@ -15,37 +15,34 @@ Such annotations are useful to (for example):
...
+3. JSpecify annotations haven't indicated whether they can be null.
 
-If `x` is a reference of a certain type, then if `x` can be null, `x.toString()` is
-unsafe because it could produce a NullPointerException. If `x` can't be null,
-`x.toString()` can never produce a NullPointerException. And if we don't know
-whether `x` can be null or not, `x.toString()` might or might not be unsafe.
+If `x` is a reference of a certain type, then if `x` can be null, `x.getClass()`
+is unsafe because it could produce a NullPointerException. If `x` can't be null,
+`x.getClass()` can never produce a NullPointerException. If JSpecify annotations
+haven't said whether `x` can be null or not, we don't know whether `x.getClass()`
```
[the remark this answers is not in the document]
eamonnmcmanus: OK, we can revisit this.

eamonnmcmanus:
PTAL

docs/user-guide.md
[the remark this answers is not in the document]
eamonnmcmanus: Sorry, I missed this earlier. I agree that it's kind of ugly to use parameter assignment for the examples. What about fields? That feels a bit more direct than using method calls. This was the best I could do with those:

```java
@NullMarked
class Example {
  void useNonNull(String nonNull) {...}
  void useNullable(@Nullable String nullable) {...}
  void example(String nonNull, @Nullable String nullable) {
    useNullable(nonNull); // JSpecify allows this
    useNonNull(nullable); // JSpecify doesn't allow this
  }
}    
```

docs/user-guide.md
```diff
@@ -15,37 +15,35 @@ Such annotations are useful to (for example):
...
 
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, all non-primitive variables are references. We often think of `String
+x;` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_
+to a string or it is null". JSpecify includes a notion of "null-marked code"
```
netdpb, line 20: This might be slightly clearer by using "either": "either x is a reference to a string or it is null".

docs/user-guide.md
```diff
@@ -15,37 +15,35 @@ Such annotations are useful to (for example):
...
+where `String x;` means "`x` is a reference to an actual string", and if you
```
netdpb, line 21: To emphasize the distinction, consider changing ", and" to a simple semicolon.

Or maybe refactor to something like:

JSpecify includes a `@NullMarked` annotation. In code covered by that annotation, `String x` means "`x` is a reference to an actual string", and `@Nullable String x` means "`x` is either null or a reference to an actual string."

docs/user-guide.md
```diff
@@ -15,37 +15,35 @@ Such annotations are useful to (for example):
...
 
 ## Types and nullness
 
-In the JSpecify world, most references have one of three possible properties
+In the JSpecify world, references have one of three possible properties
 regarding nullness:
 1. They can be null.
 2. They can't be null.
-3. We don't know if they can be null or not.
+3. JSpecify annotations haven't indicated whether they can be null.
```
netdpb, line 32: Is it simpler to remove the reference to "the JSpecify world" and say something like:

Each reference can have one of three possible properties regarding nullness:
1. JSpecify annotations indicate that it can be null.
2. JSpecify annotations indicate that it can't be null.
3. JSpecify annotations don't indicate whether it can be null.

docs/user-guide.md
```diff
@@ -15,37 +15,35 @@ Such annotations are useful to (for example):
...
 
-If `x` is a reference of a certain type, then if `x` can be null, `x.toString()` is
-unsafe because it could produce a NullPointerException. If `x` can't be null,
-`x.toString()` can never produce a NullPointerException. And if we don't know
-whether `x` can be null or not, `x.toString()` might or might not be unsafe.
+If `x` is a reference of a certain type, then if `x` can be null, `x.getClass()`
```
netdpb, line 34: Do we need "of a certain type" here? How about:

For a given reference `x`, if `x` can be null....

docs/user-guide.md
```diff
@@ -15,37 +15,35 @@ Such annotations are useful to (for example):
...
+is unsafe because it could produce a NullPointerException. If `x` can't be null,
+`x.getClass()` can never produce a NullPointerException. If JSpecify annotations
+haven't said whether `x` can be null or not, we don't know whether `x.getClass()`
+is safe (at least as far as JSpecify is concerned).
 
 There are two JSpecify annotations that indicate these properties:
 * `@Nullable` applied to a type denotes a reference of that type that can be null.
```
netdpb, line 41: Can we change "denotes" to means?

I think "denotes" requires an explicit "that", and it's more formal.

docs/user-guide.md
```diff
@@ -90,13 +88,42 @@ be too. You might use this method like this:
 Tools could then use the `@Nullable` information to determine that the first use
 is safe but the second is not.
 
+As far as JSpecify is concerned, `String` and `@Nullable String` are _different_
+types. A variable of type `String` can reference any string. A variable
+of type `@Nullable String` can too, but it can also be null. This means that
+`String` is a _subtype_ of `@Nullable String`, in the same way that `Integer` is
```
netdpb, line 94: The comma isn't necessary here.

docs/user-guide.md
```diff
@@ -90,13 +88,42 @@ be too. You might use this method like this:
...
+a subtype of `Number`. One way to look at this is that a subtype narrows the
+range of possibilities. A `Number` variable can be assigned from an `Integer`
+but it can also be assigned from a `Long`. Meanwhile an `Integer` variable
+can't be assigned from a `Number` (since that `Number` might be a `Long` or some
+other subtype). In the same way, a `@Nullable String` can be assigned from a
+`String` but a `String` can't be assigned from a `@Nullable String` (since that
+might be null).
```
netdpb, line 101: Can this be simplified to say that a subtype contains fewer values than a supertype? And the examples can refer to actual values, like "some Numbers are not Integers, such as 3.2".

Although maybe a domain that isn't boxed primitives would be better?

docs/user-guide.md
```diff
@@ -90,13 +88,42 @@ be too. You might use this method like this:
...
+```
+
 ## `@NullMarked`
 
 The `@NullMarked` annotation indicates that references can't be null in its scope,
 unless their types are explicitly marked `@Nullable`. If applied to a package then
 its scope is all the code in the package. If applied to a class or interface
 then its scope is all the code in that class or interface.
 
+Outside `@NullMarked`, `@Nullable String` still means a reference that can be
```
netdpb, line 123: Should this be `@Nullable String x` to be consistent?

docs/user-guide.md
```diff
@@ -90,13 +88,42 @@ be too. You might use this method like this:
...
+null, but JSpecify doesn't have anything to say about whether plain `String` can
```
netdpb, line 124: and `String x` here

docs/user-guide.md
```diff
@@ -15,37 +15,35 @@ Such annotations are useful to (for example):
 
 ## Java variables are references
 
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, all non-primitive variables are references. We often think of `String
```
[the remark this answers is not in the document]
eamonnmcmanus: I am fine with removing the semicolon. I think people will understand just as well without it, and as you point out we don't want to seem to be excluding parameter declarations.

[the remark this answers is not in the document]
eamonnmcmanus: I feel like this one sentence has generated more discussion than everything else. :-) I've piled on yet let more words in an attempt to make the point clear, because I do think it is an important point that sets up the rest of the document.

docs/user-guide.md
eamonnmcmanus: I feel like this one sentence has generated more discussion than everything else. :-) I've piled on yet let more words in an attempt to make the point clear, because I do think it is an important point that sets up the rest of the document.

docs/user-guide.md
eamonnmcmanus: OK, I think that works well. I've reworded the earlier text to be more consistent with it.

docs/user-guide.md
eamonnmcmanus: Thanks, I think that's better.

docs/user-guide.md
eamonnmcmanus: Not sure about the "that", but I agree that "means" is simpler and clearer.

docs/user-guide.md
eamonnmcmanus: I think it helps, though. If I were saying this aloud I would pause. I'm not saying "`String` is a subtype of `@Nullable String` in the same way that...", as opposed to some other way. I'm saying "`String` is a subtype of `@Nullable String`", and here's an example that shows what that means. However I notice that I say "in the same way" twice in rapid succession so some editing is needed. :-)

netdpb: Fair enough.

docs/user-guide.md
eamonnmcmanus: I'm inclined to say no. You can have a return value of `@Nullable String` for example. I introduced the notion that the type `String` is a reference type by talking about variables, because I think that level of concreteness is good. Maybe at some point we should explicitly make the step from rvalue variables to expressions and from lvalue variables to method arguments. I'm not sure that's really necessary, though; readers can probably generalize on their own.

netdpb: But all your previous comments about `@Nullable String x` also apply to return values. What's different here? (and by removing the semicolon, who's to say that the very next character after `x` isn't `(`?)

docs/user-guide.md
eamonnmcmanus: This is tough. I'm not particularly happy with the current text, but I want to avoid making it harder to understand. The example of `Number` with subtypes `Integer` and `Long` is very familiar and I'm not sure there's a ton of other examples that would be equally familiar. We could say that `String` and `StringBuilder` are both subtypes of `CharSequence` but I think that would confusing, because then we're talking about `String` being a subtype of two different things (`CharSequence` and `@Nullable String`) in two different ways (in the language vs in JSpecify).

I don't want to get into boxing, so I don't want to say `3.2`, any more than I wanted to say `"a"` as an example of something that a `String` variable can reference. These are things that are treated specially by the language (string literals and autoboxing) and I want to make it clear that exactly the same considerations would apply to `Object` or `Thread`, for example.

Maybe a picture would help, either instead of or as well as all this text?

docs/user-guide.md
eamonnmcmanus: Yes, I think that's better.

docs/user-guide.md
netdpb: Nah, I think what you have here is okay, at least for now.

docs/user-guide.md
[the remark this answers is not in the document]
kevinb9n: This works better for me.
We can change "JSpecify allows" to something more appropriate in a future change.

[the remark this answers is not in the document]
eamonnmcmanus: Are you saying the version with fields works better for you or the one with methods?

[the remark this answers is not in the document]
kevinb9n: I'm sorry again. I meant to express that the one with methods, that you're showing inline here in your comment, appeals to me more than the others.

I feel the meaning of an annotated parameter type or return type is much clearer and simpler than the field case, where there will always be more confusion like #39.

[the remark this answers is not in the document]
eamonnmcmanus: OK, I switched to something very like the method call example I suggested earlier.

docs/user-guide.md
```diff
@@ -90,13 +89,41 @@ be too. You might use this method like this:
...
+    useNullable(nonNull); // JSpecify allows this
+    useNonNull(nullable); // JSpecify doesn't allow this
+  }
+}
+```
+
 ## `@NullMarked`
 
 The `@NullMarked` annotation indicates that references can't be null in its scope,
 unless their types are explicitly marked `@Nullable`. If applied to a package then
```
wmdietl, line 119: This doesn't list `module` and is the only place that calls out `interface` separately from `class`.

docs/user-guide.md
```diff
@@ -111,9 +138,9 @@ public class Strings {
 ```
 
 In this example, both methods are in the scope of `@NullMarked`, so plain `String`
-means "a reference to a `String` that can't be null". `@Nullable String` continues
-to mean "a reference to a `String` that can be null". Tools should warn you if you
-try to pass a "reference to a `String` that can be null" to `spaceIndex`, since its
+means "a reference to a string, not null". `@Nullable String` continues
+to mean "a reference to a string, or null". Tools should warn you if you
```
wmdietl, line 142: Earlier https://github.com/jspecify/jspecify/pull/201/files#r662668740
Would making the distinction between the two kinds of calls explicit be useful?

docs/user-guide.md
```diff
@@ -129,20 +156,20 @@ the values that are assigned to the variable. For example:
...
   }
 }
 ```
 
-Analysis can tell that all of these variables except `copyOfTwo` can be null.
-`copyOfTwo` can't be null since `two` can't be null: it is not `@Nullable` and it is
-inside the scope of `@NullMarked`. `copyOfOne` can be null since it is a copy of a
+Analysis can tell that all of these variables except `anotherTwo` can be null.
+`anotherTwo` can't be null since `two` can't be null: it is not `@Nullable` and it is
+inside the scope of `@NullMarked`. `anotherOne` can be null since it is a copy of a
```
wmdietl, line 172: Instead of `copy of` I would say `alias of` or `reference to`, same in the next sentence.

wmdietl (commented):
A few small comments, some from #201.

docs/user-guide.md
```diff
@@ -15,37 +15,36 @@ Such annotations are useful to (for example):
...
 
-In Java, all variables, except primitives like `int`, are references. We often
-think of `String x;` as meaning "`x` is a `String`", but actually it means "`x` is a
-_reference_ to a `String`". One important difference between these two views is that
-`x` can also be null. The goal of JSpecify is to move towards a world where `String
-x;` means "`x` is a reference to an actual string", and if you want a reference
-that might be null then you write `@Nullable String x;`.
+In Java, all non-primitive variables are references. We often think of `String
+x` as meaning "`x` is a `String`", but actually it means "`x` is a _reference_,
+either null or a reference to a string".
```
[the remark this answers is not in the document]
eamonnmcmanus: I think "string object" is good, and I changed "string" and "actual string" to "string object" throughout the document.

docs/user-guide.md
eamonnmcmanus: I added the `module` case. I think it's OK that we say "class or interface" here when we just said "class" when we were introducing `@NullMarked` earlier. That was a quick intro, this is more detailed. I don't think there are any other places where we say "class" where we could say "class or interface".

docs/user-guide.md
eamonnmcmanus: After the discussion in our last meeting, I've tried to make it so that we say as little as possible about code that is not `@NullMarked`. We dispensed with non-`@NullMarked` code above and I think we can leave it at that.

docs/user-guide.md
eamonnmcmanus: I changed it to "assigned from".

docs/user-guide.md
```diff
@@ -337,21 +364,28 @@ The previous sections cover 99% of everything you need to know to be able to use
...
+`import java.util.Map.Entry` might be undesirable because `Entry` is such a
+common type name.
 
 1. For an array type, if you want to say that the _elements_ of the array can be null
 then the syntax is `@Nullable String[]`. If you want to say that the _array itself_
 can be null then the syntax is `String @Nullable []`. And if both the elements and
 the array itself can be null, the syntax is `@Nullable String @Nullable []`.
 
+A good way to remember this is that it is the thing right after `@Nullable` that
+can be null. It is the `Entry` objects that can be null in `Map.@Nullable
```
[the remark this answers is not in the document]
eamonnmcmanus: Yes, "objects" was clearly a mistake. I think it is fine just to say "It is the `Entry` that can be null.

eamonnmcmanus (commented):
Thanks Werner!

kevinb9n (approved):

kevinb9n:
For any new or remaining concerns, let's continue fixing-forward.

PR #211, opened 2021-09-03
Remove build warnings introduced by Gradle 7.2

KengoTODA:
Now builds in `main` print several warnings caused by the Gradle Modules plugin. It was introduced by Gradle v7.2 introduced by 6c3f28650dcb3843965f9691f05aa624ec300e56. So I've investigated and found a solution: java9-modularity/gradle-modules-plugin#201

The [updated version](https://github.com/java9-modularity/gradle-modules-plugin/releases/tag/v1.8.9) has been released yesterday, so I'm suggesting this PR that applies the released version to remove warnings. I've confirmed that this change has removed warnings during the build in [this workflow run](https://github.com/jspecify/jspecify/runs/3501654400).


cpovirk:
Thanks for noticing the problem and for doing all that to fix it

PR #224, opened 2022-01-01
Minor updates for welcoming 2022

KengoTODA:
* Update documents to bump up 2021 to 2022
* Update Gradle and plugins to the latest release
* Update JUnit5 to the latest release
* Update Errorprone to the latest release


cpovirk:
Thanks! As in #177, let's keep the copyright data on existing Java files. (I've pushed a commit to revert those changes.) But updating the template makes sense, and of course bumping dependencies is always good.

Issue #245, opened 2022-04-21
Make sure that our multi-release jar actually lets people use module-scope @NullMarked


cpovirk:
I had found that [it works for plain javac invocations](https://github.com/jspecify/jspecify/issues/34#issuecomment-768529601), and I believe eamonnmcmanus was the one who [confirmed for us that it works with NetBeans](https://github.com/jspecify/jspecify/issues/34#issuecomment-877496867). But we had seen [a report that it doesn't work with IntelliJ](https://github.com/jspecify/jspecify/issues/34#issuecomment-876595861), and [we haven't tried Eclipse](https://github.com/jspecify/jspecify/issues/34#issuecomment-877496867).

And now... I tried to use module-scope `@NullMarked` in a Gradle project that uses `--release 11`, and:

```
error: annotation type not applicable to this kind of declaration
```

That's happening even though I see a `jspecify-0.2.0.jar` (with what looks like the right contents) on the `--module-path` when I pass `-d` to Gradle.

cpovirk:
On the Gradle front, I found a bug that sound potentially related: https://github.com/gradle/gradle/issues/19587. I also wasn't surprised to find that the Gradle people [recommend](https://github.com/gradle/gradle/issues/10046#issuecomment-514119882) that library authors instead use [their mechanism for separate jars](https://blog.gradle.org/mrjars), the thing that I continually feel guilty for [not enabling for Guava](https://github.com/google/guava/pull/3683). Separate jars is certainly a possible fallback option, though Guava users are familiar with some of the resulting pain, and Gradle's work to eliminate that pain won't help users of other build systems.

Also, to be clear: @KengoTODA had [added some tests](https://github.com/jspecify/jspecify/commit/1b044838727efdd134df96d234e0d13b52fb4d3f#diff-ab088f59da4188c8a5d2295914afdc261699f7d7896e7c1fa4d85a2cde8c1018R72) to make sure that the right version of `@NullMarked` is selected _at runtime_. The problem I was seeing was _at compile time_.

And that problem was:

```
$ ./gradlew compileJava
...
/usr/local/google/home/cpovirk/clients/caffeine-black/caffeine/caffeine/src/main/java/module-info.java:1: error: annotation type not applicable to this kind of declaration
@org.jspecify.nullness.NullMarked
^
1 error
...
```

...when I tried to make [this change to Caffeine](https://github.com/cpovirk/caffeine/commit/e9ff9b354ce863a99930bb725e098fd49e889a56).

[...]

Issue #172, opened 2021-02-09
Move module-info.java to multi-release jar Java 9+ directory


cpovirk:
Now that we have a multi-release jar, we could do this.

I _think_ I heard this pattern suggested somewhere, but I'd want to confirm. But it's the end of the day, so I'm filing this issue and walking away.

(Note to self: If we do this, then I can simplify our internal Java 8 setup to ignore only `java9` instead of both `java9` and `module-info`.)

cpovirk:
A couple suggestions of this pattern:

- https://github.com/google/guava/issues/2970#issuecomment-626684027
- https://openjdk.java.net/jeps/238#Modular-multi-release-JAR-files (despite some confusing wording that initially suggests that the module-info _must_ appear at the top level of the jar)

cpovirk:
It's possible that this would eliminate the need for `modularity.mixedJavaRelease`...
https://github.com/jspecify/jspecify/blob/19961fa5ec1ad822106f52e6a1b427c8c3a111da/gradle/publish.gradle#L31
...which nearly got me into trouble [in 8135398](https://github.com/jspecify/jspecify/commit/81353983747a772e930890f24f72f83edc00c466#diff-d6dec2a6144229896b7142578a46e5aede6004e37b4fbde49f9cd2a129adef54R37), as noted in the description of https://github.com/jspecify/jspecify/commit/6fae6f081e170285eb8decae19249396eab0262c.

kevinb9n:
Is this an issue purely of internal organization or does it affect the product?

cpovirk:
I don't know a ton about this, but my impression is that it _could_ affect the product -- but I'm fuzzy on the details, _and_ we could do it in a patch release anytime.

One thing I've heard in the past is that some build tools blindly scan jars for `.class` files, and some such tools used to get confused by `module-info.class`, if only because it's necessarily built with `-source 9` or higher. Putting `module-info.class` under `META-INF` (as this feature request proposes) typically keeps those tools from discovering it.

That said, I would _hope_ that `module-info`-ignorant tools would be rare nowadays.

_That_ said, I don't know if there are _other_ reasons to move the `module-info.class`.

I'm highly comfortable calling this post-1.0.

PR #246, opened 2022-05-10
mrjar: Set `@Target({..., MODULE})` in the root, and move `module-info` out.

cpovirk:
Fixes https://github.com/jspecify/jspecify/issues/245. But introduces a compile warning for users targeting old versions of Java, and introduces [runtime problems](https://bugs.openjdk.java.net/browse/JDK-8247797) if a user tries to read _the annotations on `@NullMarked` itself_.

Fixes https://github.com/jspecify/jspecify/issues/172. I didn't really need to do this in the same commit. I just happened to get things working that way first, and we've talked about at least trying this change, so here goes....

It's clear that there's no great option for `MODULE`. But given that we've already encountered problems in the wild with our old approach, let's try this new approach instead.


build.gradle
```diff
@@ -3,7 +3,6 @@ plugins {
     id 'io.github.gradle-nexus.publish-plugin' version '1.1.0' apply false
     id 'com.diffplug.spotless' version '6.1.0'
     id 'net.ltgt.errorprone' version '2.0.2'
-    id 'org.javamodularity.moduleplugin' version '1.8.9'
```
KengoTODA, line 6: :+1: 
Better to stop depending on this plugin, it seems their development is not so active recently. I sent a PR two months ago and there is still no reaction 😂

gradle/mrjar.gradle
```diff
@@ -27,14 +33,16 @@ tasks.named('compileJava9Java', JavaCompile).configure {
...
 jar {
     into('META-INF/versions/9') {
-        from sourceSets.java9.output
+        from(sourceSets.java9.output) {
+            /*
+             * As discussed above, we compiled all the classes again. But all
+             * that we want to include under META-INF/versions/9 is the
+             * module-info.
+             */
+            include 'module-info.class'
```
KengoTODA, line 42: :+1:
Great comment, it will help us to understand the intention of this build script.

KengoTODA (approved):
LGTM, I've learned a lot from related discussions.
Also thank you making the build script more easy-to-understand!

build.gradle
cpovirk: Thanks, that is good to know :)

PR #247, opened 2022-06-10
Update EISOP entry

wmdietl:


cpovirk:
@kevinb9n in case there's some significant reason that you did it the way you did it, but I imagine we'd generally prefer to let the participants supply their own labels.

Issue #215, opened 2021-09-24
@Implies syntax (array or repeatable?)


cpovirk:
I'd like to propose that #35 and #213 be focused on requirements questions. That way, we can break off smaller issues for details -- such as this one.

Between those two threads, we have a few options for how to declare the one annotation implies another:

1. `@Implies({Foo.class, Bar.class}) @interface MyAnno {}`
2. `@Implies(Foo.class) @Implies(Bar.class) @interface MyAnno {}`
3. `[@MetaAnnotated] @Foo @Bar @interface MyAnno {}`

(This issue isn't intended to be about _names_, just the choices between the 3 patterns above.)

The main choice is probably between option 1/2 and option 3. (Or we could support both.)

A possible advantage to option 3 is that is works naturally with annotations that have elements. That is, you can define an annotation that implies `@Foo(bar = "baz")`. Or maybe annotation elements are tricky no matter what. (Thankfully, we've mostly been leaning toward avoiding them.) For example, maybe it's _better_ to write `@Implies(Foo.class) @interface MyAnno { ... }` because then it makes at least _some_ sense to support `@MyAnno(bar = "baz")` as meaning the same as `@Foo(bar = "baz")`.

A possible disadvantage to option 3 is that it's easy to imagine automatically applying an annotation like `@NullMarked` or `@NullUnmarked` to every class in a project, only to discover that you've accidentally applied that annotation's semantics to all usages of your project's annotation in other projects! (Or similarly, imagine the "no-op" change of moving in either direction between package- or module-level `@NullMarked` and per-class `@NullMarked`.) This danger could be reduced by requiring a `@MetaAnnotated` annotation -- but not eliminated, since some annotations will legitimately have `@MetaAnnotated` to begin with.

I know that we've had some good success with option 3 for some annotations inside Google, but it scares me as a much more widely used, general-purpose approach that we'd encourage people to apply "by default" to new code and migrate existing code to use.

kevin1e100:
I guess option 3 requires allowing JSpecify annotations to be placed on
annotation classes, while that's not needed for 1/2--there, JSpecify
annotations become attribute values of another annotation instead.

Not sure that difference is an advantage or disadvantage for option 3
though ;).

cpovirk:
[...]

PR #258, opened 2022-08-08
Change `@Implies` from a repeatable annotation to one with an array of values

netdpb:
Addresses question in #215.

### DIFF
--- a/src/main/java/org/jspecify/meta/Implies.java
+++ b/src/main/java/org/jspecify/meta/Implies.java
@@ -20,40 +20,23 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 import java.lang.annotation.Annotation;
 import java.lang.annotation.Documented;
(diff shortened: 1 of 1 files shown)


cpovirk (approved):
Looks like a good implementation of the change. Maybe sit on this until we have more time to mull over https://github.com/jspecify/jspecify/issues/215 and figure out whether to make the change or not? (I have been pretty strongly in favor of the change, but my response has been a little too reflexively "of course we wouldn't use `@Repeatable` unless we had to," so I wouldn't mind some time to think it over. Not that we can't revert any change that we want to, anyway....)

netdpb:
Yes, this was in anticipation. Will not merge until the decision is made on that issue.

kevinb9n:
The docs I have ready to merge in have already been adjusted for the array. I see no problem merging this; if we need to undo it there's the extra step of changing the docs back anyway.

PR #264, opened 2022-08-16
Set javadoc encoding

wmdietl:
Recently, Checker Framework builds started failing with an error about encodings in JSpecify files:

````
> Task :javadoc FAILED
/__w/1/jspecify/src/main/java/org/jspecify/nullness/NullUnmarked.java:44: error: unmappable character (0xE2) for encoding US-ASCII
 * <p>Otherwise it is considered null-unmarked: whether that???s because it is more narrowly enclosed
                                                            ^
/__w/1/jspecify/src/main/java/org/jspecify/nullness/NullUnmarked.java:44: error: unmappable character (0x80) for encoding US-ASCII
 * <p>Otherwise it is considered null-unmarked: whether that???s because it is more narrowly enclosed
                                                             ^
/__w/1/jspecify/src/main/java/org/jspecify/nullness/NullUnmarked.java:44: error: unmappable character (0x99) for encoding US-ASCII
 * <p>Otherwise it is considered null-unmarked: whether that???s because it is more narrowly enclosed
````

I can only reproduce the issue in the CI system, not locally.

This PR explicitly sets the javadoc encoding to UTF-8, which fixes the issue, see the corresponding EISOP PR:
https://github.com/eisop/checker-framework/pull/323

Alternatively, we could replace the offending character in `jspecify/src/main/java/org/jspecify/nullness/NullUnmarked.java`.

Another alternative might be to change the CF docker images in some way, but I didn't see anything related to encodings in them.

Typetools [disabled their JSpecify dependency](https://github.com/typetools/checker-framework/commit/803bc76c71944cf53956771b669c570dc2ea3517), which I can't do in EISOP.

(Note that I didn't look at the Javadoc HTML output before or after this change, so I don't know whether anything changes.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,4 +34,8 @@ java {
     }
 }
 
+javadoc {
+    options.encoding = 'UTF-8'
+}
+
 defaultTasks 'spotlessApply', 'build'


cpovirk:
Thank you!

I don't know if @kevinb9n wants to try to use smart quotes in the docs normally: Clearly we should be _able_ to (as this PR makes possible), but presumably we'd want to be consistent about it. We should merge this in any case, but I'd be interested in what he thinks.

(Any chance that setting up CI for multiple Java versions is something that's trivially straightforward for you or someone else who knows Gradle or maybe more so GitHub Actions? We've [done it for Guava](https://github.com/google/guava/blob/5c8719e28880a0f942272bdd57d9a194a2d6226c/.github/workflows/ci.yml#L21), so presumably it's not hard.)

msridhar:
> (Any chance that setting up CI for multiple Java versions is something that's trivially straightforward for you or someone else who knows Gradle or maybe more so GitHub Actions? We've [done it for Guava](https://github.com/google/guava/blob/5c8719e28880a0f942272bdd57d9a194a2d6226c/.github/workflows/ci.yml#L21), so presumably it's not hard.)

I could take a crack at this; which Java versions do we want to run CI with?

cpovirk:
Thanks! That is another question I would be happy for help in answering :) [typetools](https://github.com/typetools/checker-framework/blob/a6241349d5537aec32212391682b702230f272ff/.travis.yml#L21)+[eisop](https://github.com/eisop/checker-framework/blob/4e483296fe6dc8397e669497a9e9a04fe25cb113/.travis.yml#L21) are the only projects I know of that _build_/built JSpecify as part of their builds, and they target 8 (which we don't support), 11, and 17. So that seems like a good place to start. Presumably we'll be able to add versions easily once the infrastructure is in place.

wmdietl:
Thanks for merging this!
Testing everything with multiple Java versions is definitely good, but I'm not sure this is a Javadoc 17 issue. I couldn't reproduce the failure locally with any version.

cpovirk:
Understood, thanks. There was an additional issue with newer patch versions of Java 17+ or so, which I think I addressed with https://github.com/jspecify/jspecify/commit/6a1837b417e07beeb4d907995d7e1cbe73c6617d

PR #265, opened 2022-08-16
Run CI jobs on multiple JDK versions

msridhar:
I chose 11, 17, 18, and 19-ea, but this can easily be tweaked.

I also updated to `actions/setup-java@v3` and switched to the maintained Temurin JDK distribution.


msridhar:
Cool.  I think when 19 is released, we could shift to running on 19 and 20-ea, and so on.

PR #266, opened 2022-08-16
Bump Gradle to 7.5.1

msridhar:
This just keeps us up to date.


cpovirk:
(I should probably ask about norms for GitHub projects that actually merge commits directly, rather than the way we do for projects like Guava.... Is it preferable for me to merge or for me to approve and leave the merge to you? Are there standard practices for this, or do different projects do it differently?)

PR #270, opened 2022-08-22
Test that NullMarked.class.getAnnotations() really does fail under JDK8.

cpovirk:


cpovirk:
OK, I was wondering if this couldn't actually work right in CI, since we [don't request JDK 8](https://github.com/jspecify/jspecify/blob/4242439295d8f75e69c72f012bbde14dda2120d7/.github/workflows/build.yml#L19). But it looks like [we request the latest Ubuntu release](https://github.com/jspecify/jspecify/blob/4242439295d8f75e69c72f012bbde14dda2120d7/.github/workflows/build.yml#L8), which [currently is 22.04](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#preinstalled-software), which [has JDK 8 by default](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#java).

It would be more hygienic for us to request it explicitly. But for now, everything works.

msridhar:
It's not clear to me this new test is even running in CI, since we never use JDK 8.  Do you want to do a limited set of tests on JDK 8 in CI?

cpovirk:
Thanks, I'll investigate further, probably by sending a PR that should fall the JDK 8 test. (I do see the Java 8 integration test show up in the Gradle output, but I don't know if the toolchains magic responds to a missing JDK by quietly substituting another JDK or by failing the build. I'd aimed to test that with this PR, but I didn't think the "quiet substitution" scenario through: In that case, I think we'd quietly run the Java 9+ tests twice and the Java 8 tests zero times. So this PR doesn't distinguish between that and actually testing insert JDK 8.)

cpovirk:
Hooray for the [failure](https://github.com/jspecify/jspecify/runs/7975368500) that [I was aiming for](https://github.com/jspecify/jspecify/pull/272/files#diff-6a22a6ede57062df5613483ebcc3fd278316078440bc35c7bf253ef49e89ee12R57):

```
@NullMarked > with Java 8 > basicReflectionDoesNotThrowException() FAILED
    org.opentest4j.AssertionFailedError at NullMarkedTest.java:66
```

msridhar:
Sorry for the confusion on my end!

cpovirk:
No problem at all. It was the kind of test I'd been _thinking_ I'd done originally, but I hadn't!

Issue #271, opened 2022-08-22
The build vs. JDK versions


kevinb9n:
I don't understand build stuff, but I ?think? that currently:

* Most gradle commands respect the "environmental" JDK
* The main compile uses `-source 8 -target 8` to make sure the JAR is 8-compatible
* Our github actions use this to test out several different JDKs

But isn't the only thing that we really need to splay out to multiple JDKs for tests? More to the point: since we will only be releasing one artifact, shouldn't these should all test against the exact same single jar -- the one we will actually publish?

So I'm wondering if for purposes of compilation we can/should hardcode a JDK version? And if we do the same thing for the javadoc generation (setting it to say 17) it would fix the current issue that it's failing on JDK 11.

As for running tests though, I suppose it's okay to keep working as it has been? I don't know what the tradeoffs would be of moving to one distinct more-hermetic gradle target for each.

JakeWharton:
Yes!

I meant to comment on the original PR but it landed before I had a chance. I wrote about this a bit [here](https://jakewharton.com/build-on-latest-java-test-through-lowest-java/). Gradle makes this super easy. Set your toolchain to 18 (19 in a few weeks!) and create a test task for each JDK in your supported range.

kevinb9n:
Whoa thanks, that's perfectly helpful!

PR #273, opened 2022-08-23
Mark 19-ea as experimental.

cpovirk:
I'm hoping that this will address [a sporadic(?)
failure](https://github.com/cpovirk/jspecify/runs/7975247419):

```
Could not open settings generic class cache for settings file '/home/runner/work/jspecify/jspecify/settings.gradle' (/home/runner/.gradle/caches/7.5.1/scripts/8fxsx7a5az16qabak9o2v29h1).
> BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 63
```

Gradle doesn't clearly support Java 19 yet:
https://docs.gradle.org/current/userguide/compatibility.html

I'm basing this config on [similar configuration in Error
Prone](https://github.com/google/error-prone/blob/f94a04d407583bc9f85847089d72298deecf6fb0/.github/workflows/ci.yml#L23).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ jobs:
     strategy:
       matrix:
         java_version: [11, 17, 18]
+        experimental: [false]
         include:
           - java_version: 19-ea
             experimental: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java_version: [11, 17, 18, 19-ea]
+        java_version: [11, 17, 18]
+        include:
+          - java_version: 19-ea
+            experimental: true
     steps:
       - uses: actions/checkout@v2
       - uses: gradle/wrapper-validation-action@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,7 @@ jobs:
         include:
           - java_version: 19-ea
             experimental: true
+    continue-on-error: ${{ matrix.experimental }}
     steps:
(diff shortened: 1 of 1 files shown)


cpovirk:
Maybe this will become obsolete after https://github.com/jspecify/jspecify/issues/271#issuecomment-1223320796, but for now, I'm hoping that it eliminates any flakiness in the CI.

.github/workflows/build.yml
```diff
@@ -8,7 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java_version: [11, 17, 18, 19-ea]
+        java_version: [11, 17, 18]
```
msridhar, line 11:
```suggestion
        java_version: [11, 17, 18]
        experimental: [false]
```
To set the flag for other builds

.github/workflows/build.yml
```diff
@@ -8,7 +8,10 @@ jobs:
...
+        include:
+          - java_version: 19-ea
+            experimental: true
```
msridhar, line 14:
```suggestion
            experimental: true
    continue-on-error: ${{ matrix.experimental }}
```
I think you need that line to allow the overall build to succeed even if `19-ea` fails.  See, e.g., here: https://github.com/google/error-prone/blob/f94a04d407583bc9f85847089d72298deecf6fb0/.github/workflows/ci.yml#L34

msridhar (changes requested):

cpovirk:
Thank you! I should have pulled up https://github.com/google/error-prone/commit/50e2e7b54215f8fc0e2b6d82e3667a12f0f6e530 when I was writing this PR, but obviously I did not.

PR #282, opened 2022-09-07
docs: update README

scolsen:
Make the warning about changes pre-1.0 more prominent.

FWIW, this change isn't strictly necessary since we don't have anyone saying they *didn't* realize it was pre-1.0 and subject to change, however, I think it's important to make caveats like this hard to miss.


scolsen:
oh. looks like vim squashed this to an 80 col limit. sorry for the diff noise

PR #286, opened 2022-09-13
Replace leading `@` in code snippets in Javadoc

netdpb:
Use `&commat;` so that the Javadoc processor doesn't try to interpret them as block tags.


netdpb:
We could use `&64;` instead.

cpovirk:
> That was discussed more somewhere else, possibly somewhere in #269.

OK, that actually wasn't as hard to find as I feared it would be:
https://github.com/jspecify/jspecify/pull/269#issuecomment-1223096045

kevinb9n:
If there IS a fix that only requires ~6 little hacks like this and completely works I'd be okay with merging it.

netdpb:
Closed as unnecessary after #290.

PR #297, opened 2022-10-03
Merge.

cpovirk:


cpovirk:
(created from the "Create pull request" button on https://github.com/jspecify/jspecify/compare/main...samples-google-prototype, which is accessible from the "72 commits ahead, 2 commits behind" links at https://github.com/jspecify/jspecify/tree/samples-google-prototype)

netdpb (approved):

cpovirk:
Thanks. Here goes nothing! I am encouraged by the "command line instructions" link, which appears to be performing a `git merge`.

cpovirk:
That does not look like a merge commit to me :\

"1 parent": https://github.com/jspecify/jspecify/commit/cd35d7281e829f03a3ef64af38c0bdadb8cb1686

Contrast "2 parents": https://github.com/jspecify/jspecify/commit/47f68d8cb4218ad89ae7e057f19b5edbfd08c5d9

And `git merge main` (after I fetched the new pointer for each branch) tries to do stuff. Granted, it produces an empty commit, but it's a sign that we didn't get a true merge.

So I just did a `git push -f` of https://github.com/jspecify/jspecify/commit/7f7d67fde5bdb000c2ac933810797414422e5165, whose contents are identical to the original PR. In the future, it looks like we may want to perform merges by using the command-line instructions, or maybe we'll want to reenable "Create a merge commit" and see if that helps. I'll try to remember to try that next time.

PR #319, opened 2022-11-04
Trying moving the javadoc where we want it

### LINKED ISSUE #277: Decide javadoc permalink
Decide javadoc permalink. Until then it is at mumblemumble-"nolink".

kevinb9n:
I'm not really set up to test this yet; easier to just see if it works.

Should resolve #277


cpovirk (approved):

.idea/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Default ignored files
```
scolsen, line 1: could this just be `.gitignore` at root with:

```
.idea/
```

scolsen (approved):

.idea/.gitignore
kevinb9n: sigh, I forgot that would be in here.

But at some point I want to figure out which is the actual right set of IDEA files to share, so I'll optimize this then.

PR #321, opened 2022-11-04
Smallspecupdates

kevinb9n:
* I didn't get far
* I know there are bigger changes that need to be made, which possibly I might be able to help with too
* Some are obviously not totally right yet
* We don't have to merge this; we could batch up more and change the spec at once


kevinb9n:
Sorry, of course I missed some step to tell it that the 2 commits that were *already* pulled don't need to show up here.

docs/docs/spec.md
```diff
@@ -146,16 +141,18 @@ A nullness operator is one of 4 values:
 >     -   One way to conceptualize this is that `String NO_CHANGE` means
 >         "non-null `String`" but that `T NO_CHANGE` means "nullness comes from
 >         the value of `T`."
-> -   `UNSPECIFIED`: This is the operator produced by not putting `@Nullable` on
->     a type usage *in code that is outside a [null-marked scope]*. Roughly, it
->     is the operator assigned to "completely unannotated code."
+> -   `UNSPECIFIED`: This is the operator produced for a type usage *outside* of
```
cpovirk, line 144: I'm kind of so-so on this change. There was some parallelism to `NO_CHANGE`, though I admit it wasn't great. If not for the "FIX THAT" part, I'd probably say to go for it, but maybe just revert so that we can land the rest of the PR?

docs/docs/spec.md
```diff
@@ -146,16 +141,18 @@ A nullness operator is one of 4 values:
...
+>     [null-marked scope]*, where nullness type-use annotations are applicable
+>     **FIX THAT?**
+>     yet none appears. Roughly, it is the operator assigned to "completely
+>     unannotated code."
 >     -   The type usage `String UNSPECIFIED` includes `"a"`, `"b"`, `"ab"`,
->         etc., but the developer did not specify whether to include `null`.
+>         etc., but it has been left ambiguous whether to include `null`.
```
cpovirk, line 150: I like the "developer did not specify" phrasing for its precision. But presumably my approach of beating them over the head with variants of the word "specified" is passing up the opportunity to add _something_, so you're presumably aiming to do that with "ambiguous?"

docs/docs/spec.md
```diff
@@ -281,8 +278,8 @@ Additionally, any location above is unrecognized if it makes up *any
 
 -   a local variable type
 -   an exception parameter
--   the type in a cast expression
--   an array or object creation expression
+-   the type in a cast or `instanceof` expression
+-   an array or object creation expression (including via a member reference)
```
cpovirk, line 282: (I am getting myself confused as I try to differentiate parts of this bullet from the following bullet. But I'm going to say to just go ahead with this change: This whole section will be blown up and rewritten to accommodate our changing perspective toward implementation code, presumably following the structure of your Wiki page.)

docs/docs/spec.md
```diff
@@ -334,7 +331,7 @@ locations listed below:
 
 To determine whether a type usage appears in a null-marked scope:
 
-Look for an `@org.jspecify.nullness.NullMarked` annotation on any of the scopes
+Look for a `@NullMarked` annotation on any of the scopes
```
cpovirk, line 334: I think there is value is fully specifying the name to be clear that we're not matching on simple name. Ditto in a couple places below.

Alternatively, I can see an argument that we could define somewhere that, when we say "`@NullMarked," etc., we _always_ mean that specific class.

But also, for fun, I'm choosing to interpret this comment as another look of disapproval toward the current package name :)

docs/docs/spec.md
```diff
@@ -1,37 +1,31 @@
...
-:::note Temporary advice to readers (non-normative)
+:::note Advice to readers (non-normative)
 
-For someone new to our nullness annotations, this document does not make a good
-introduction. This document is targeted more at tool authors or advanced users.
-New users will prefer to start with our [User Guide]. We are working on further
-user documentation, including Javadoc.
+The primary audience for this document is the authors of analysis tools. Some
+very advanced users might find it interesting. But it would make a very poor
+introduction for anyone else; instead see the [User Guide] and [Javadoc].
```
cpovirk, line 14: 🎉

cpovirk (commented):
Thanks!

docs/docs/spec.md
```diff
@@ -1,37 +1,31 @@
...
---------------------------------------------------------------------------------
-
-:::note Temporary advice to readers (non-normative)
+:::note Advice to readers (non-normative)
 
-For someone new to our nullness annotations, this document does not make a good
-introduction. This document is targeted more at tool authors or advanced users.
-New users will prefer to start with our [User Guide]. We are working on further
-user documentation, including Javadoc.
+The primary audience for this document is the authors of analysis tools. Some
```
scolsen, line 12: Might read better flip flopped:

"Analysis tool owners are the primary audience for this document."

kevinb9n: For some reason that reads weirder to me. So uh... sword fight to the death? :-)

docs/docs/spec.md
kevinb9n: sounds amazing, will do

docs/docs/spec.md
kevinb9n: I could be okay with it, but it felt strange for a spec like this to personify "the developer" like this. All it really knows is what's in the code.

They generally say to avoid the passive-voice-with-unspecified-subject, but I always wondered at that advice, because sometimes the point is that the subject absolutely doesn't matter!

docs/docs/spec.md
kevinb9n: Ah, I read "details common to all annotations" as defining just that. Could we tweak it so it accomplishes that more clearly? I think the FQN is obtrusive and of course is just one more place to fix on rename (which obvs I expect to happen SEVERAL MORE TIMES)

kevinb9n (commented):
Don't PTAL yet, just continuing discussion

docs/docs/spec.md
scolsen: An unnecessarily long argument for the suggestion:

Argument from authority? See bullet 24. in section 1. https://jmlr.csail.mit.edu/reviewing-papers/knuth_mathematical_writing.pdf Technically, the advice is scoped to opening paragraphs and you could argue this is a slightly different case, but I feel it generalizes. The form `Predicate ... Subject` is more awkward in English than `Subject .... Predicate`. Subject-Verb-Object order is the default, after all. I think this reads better when you take the more specific class as the subject of the sentence when it comes to conveying intention. In this case I think the predication is the relation of some class of things (analysis tool owners) to the document, and not the relation of an abstract concept (the document audience) to a concrete class of things (analysis tool owners). I think we should try to make the concrete aspects of meaning the "centrifugal force" of the sentence; when you place the concrete portion in the S position of SVO, it adds emphasis and will tend to parse better. I take `analysis tool authors` as the subject since it's the answer to the question form: `who is the primary audience of this document?`. Existential statements like this one are a bit fuzzier though, so I do think it's more debatable than other clear-cut cases of using passive voice when active voice ought to be preferred.

Also, I didn't mean to rewrite your wording, I commented too quickly:

`Analysis tool authors are the primary audience for this document.`

You could also use `of` instead of `for`.

At any rate, I don't think it will matter much in the grand scheme of things, so I'm fine if you stick with your original wording.

kevinb9n: You're giving me the "I'm fine" out, so I'm gonna take it. :-)

This is really not an "opening paragraph", and if we need to set it off in a colored box or something to make that clear we can do it.

Knuth's examples both sound better the way he rewrote them. That's argument enough for me. This one, to me, really doesn't.

Part of it is my informal sorta "ADHD writing principles", which is that readers' minds are eager to tune out your words as much as possible. So I try to order sentences and paragraphs in the order "who cares about this / when they care about it --> what they care about". Here, if I start reading "Analysis tool authors", it's a subtle clue that I don't have to care about the rest because that's not me.

docs/docs/spec.md
kevinb9n: I've tweaked this. Can go back to "the developer" if you prefer (I just don't, that's all).

docs/docs/spec.md
kevinb9n: tweaked that section.

kevinb9n:
Ok I think this is "PTAL" status now.

kevinb9n:
(Oh, except I still don't know the way to make the unrelated already-pulled changes disappear from here.)

docs/docs/spec.md
cpovirk: Thanks, I think that helps.

docs/docs/spec.md
cpovirk: I agree with Kevin's information ordering (and his dispute-resolution mechanism :))

The much too long article I've cited for this before is [this one](https://www.e-education.psu.edu/styleforstudents/c10_p6.html). It advocates for putting the "old" information at the end of the sentence and the "new" at the end. (Here's [a shorter page that makes a similar argument](https://www.student.unsw.edu.au/using-old-new-information-order-sentence.), which I just pulled from a search now without reading in any detail.)

Here, I see the old information as "We're talking about this doc" and the new information as "The doc we're talking about? It's useful for tool authors."

I think this fits well with Kevin's argument that, if the paragraph starts off by mentioning "tool authors," other people will tune out because the paragraph doesn't seem to be "about" the audience; it appears to be "about" tool authors.

(I was a huge fan of the article I linked above when I first found it, so I have surprisingly strong opinions on this :))

scolsen: Thanks! I think the information order argument is a good one, people certainly skim a lot, so it makes sense. The grammatical structure still sounds odd and robotic to me for some reason, but I think the arguments you both give are more compelling in this case.

PR #323, opened 2022-11-07
various small web site updates

kevinb9n:
I'm not using docusaurus (yet?)... just trying to tweak text in ways that I hope won't upset the programming.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -7,8 +7,8 @@ const darkCodeTheme = require('prism-react-renderer/themes/dracula');
 /** @type {import('@docusaurus/types').Config} */
 const config = {
   title: 'JSpecify',
-  tagline: 'Standard Java Annotations for Static Analysis',
-  url: 'https://jspecify.dev/',
+  tagline: 'Tool-Independent Annotations for Java Static Analysis',
+  url: 'http://jspecify.org/',
   baseUrl: '/',
   onBrokenLinks: 'throw',
   onBrokenMarkdownLinks: 'warn',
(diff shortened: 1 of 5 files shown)


docs/docusaurus.config.js
```diff
@@ -81,13 +81,21 @@ const config = {
             title: 'Docs',
             items: [
               {
-                label: 'User Guide',
+                label: 'Draft User Guide',
```
netdpb, line 84: Isn't everything a draft at this point?

docs/docusaurus.config.js
```diff
@@ -81,13 +81,21 @@ const config = {
...
                 to: '/docs/user-guide',
               },
               {
-                label: 'Specification',
+                label: 'Current Javadoc',
```
netdpb, line 88:
```suggestion
                label: 'Javadoc',
```

docs/docusaurus.config.js
```diff
@@ -81,13 +81,21 @@ const config = {
...
+                to: '/docs/api/org/jspecify/nullness/package-summary.html',
+              },
+              {
+                label: 'Draft Spec',
                 to: '/docs/spec',
               },
+              {
+                label: 'JSpecify Wiki',
```
netdpb, line 96:
```suggestion
                label: 'Wiki',
```

otherwise "JSpecify" goes everywhere.

netdpb (approved):

docs/docusaurus.config.js
kevinb9n: Yes; I think it probably benefits from being labeled as such as much as possible.

kevinb9n:
Merged. Javadoc link is right but not working; I think the f/w expects it to be an "internal" link and I have to make it like the github link.

Issue #267, opened 2022-08-16
Can we switch the `.dev` and `.org`?


kevinb9n:
Right now, when looking at the JSpecify site, you always see a `.dev` URL. That is, `jspecify.org` only redirects.

It would be great if we could reverse that arrangement, since we ask users to write `import org.jspecify` in their code.

How feasible is that?

kevinb9n:
Making matters a bit worse, any deep links such as I am putting into javadoc are having to link to the `.dev` site, because going to `.org` gives a scary SSL warning (which, if you click through it, redirects just fine). It feels problematic to sanctify the `.dev` address that way.

cpovirk:
(I think you can link to http://jspecify.org/whatever fine: The problem is using https without SSL in place. (I can't remember if I had any idea how easy it was to get SSL in place for the .org site. Maybe it's easy to get in place for the "main" site??)

wmdietl:
Earlier discussion: #195.

kevinb9n:
Not sure what it will take to fix this, but in the meantime I am always linking to `.org` as much as possible.

cgdecker:
I'm working on getting this addressed so that `.org` is the canonical domain and `.dev` redirects to that.

kevinb9n:
Would appreciate this very much. There was a very small burst of discussion about JSpecify recently, and every comment that referred to the site always called it `jspecify.dev`. It's not a disaster, but it would be much nicer for them to see `.org` as the real domain.

sdeleuze:
Current user experience is really suboptimal as users trying to reach the domain used in the package see that:

![Image](https://github.com/user-attachments/assets/e18ba4b5-e4e3-4efb-a6b9-beb42e9d4519)

I think we should:
 - Switch the default domain to `jspecify.org`
 - Redirect `jspecify.dev/*` to `jspecify.org/*` with a permanent redirect to transfer the pagerank and avoid breaking links.
 - Ensure that both `jspecify.dev` and `jspecify.org` have their own HTTPS certificate to avoid the problem shown in my screenshot
 - Replace any occurence of `jspecify.dev` by `jspecify.org` in our documentations.

PR #324, opened 2022-11-07
Try to fix javadoc link from site

kevinb9n:
probably needs to be an absolute link

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,7 +1,8 @@
 # Default ignored files
 /shelf/
-/workspace.xml
 /compiler.xml
 /jarRepositories.xml
+/markdown.xml
 /misc.xml
 /vcs.xml
+/workspace.xml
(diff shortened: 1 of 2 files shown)


kevinb9n:
(I hope we are willing to let small .gitignore tweaks ride along with real changes)

netdpb:
I think the problem was just a caching one. If I refresh the 404 page I get the Javadoc.

netdpb:
Also, your URL redirects to https://jspecify.dev/docs/api/org/jspecify/nullness/package-summary.html.

kevinb9n:
> I think the problem was just a caching one. If I refresh the 404 page I get the Javadoc.

Same here, but if I go back and do it again, I get the 404 again. I think it's because the javascript updates the browser to believe it's at that URL without actually loading that URL, which is what it would want to do for "internal" pages. Except it's weird that it even follows the redirect, then.

> Also, your URL redirects to https://jspecify.dev/docs/api/org/jspecify/nullness/package-summary.html.

Correct. See #267

kevinb9n:
Since this *ought* to fix a broken link I think I'll submit it and handle any ongoing concerns in another PR if you have them.

PR #326, opened 2022-11-08
first round of javadoc revisions

kevinb9n:
I'm still going to be revising this numerous times. I'm aware it has plenty of issues; the question is which really need to be fixed for 0.3.

### DIFF
--- a/src/main/java/org/jspecify/nullness/NonNull.java
+++ b/src/main/java/org/jspecify/nullness/NonNull.java
@@ -24,7 +24,7 @@ import java.lang.annotation.Target;
 
 /**
  * Indicates that the annotated <a href="http://github.com/jspecify/jspecify/wiki/type-usages">type
- * usage</a> (commonly a variable type or return type) is considered to <i>exclude</i> {@code null}
+ * usage</a> (commonly a parameter type or return type) is considered to <i>exclude</i> {@code null}
  * as a value; rarely needed within {@linkplain NullMarked null-marked} code.
  *
  * <p>This annotation serves two primary purposes:
--- a/src/main/java/org/jspecify/nullness/NonNull.java
+++ b/src/main/java/org/jspecify/nullness/NonNull.java
@@ -23,7 +23,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 /**
- * Indicates that the annotated <a href="http://github.com/jspecify/jspecify/wiki/type-usages">type
+ * Indicates that the annotated <a href="https://github.com/jspecify/jspecify/wiki/type-usages">type
  * usage</a> (commonly a parameter type or return type) is considered to <i>exclude</i> {@code null}
  * as a value; rarely needed within {@linkplain NullMarked null-marked} code.
  *
--- a/src/main/java/org/jspecify/nullness/NonNull.java
+++ b/src/main/java/org/jspecify/nullness/NonNull.java
@@ -23,10 +23,11 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
 /**
- * Indicates that the annotated type usage is considered to exclude {@code null} as a value; rarely
- * needed within {@linkplain NullMarked null-marked} code.
+ * Indicates that the annotated <a href="http://github.com/jspecify/jspecify/wiki/type-usages">type
+ * usage</a> (commonly a variable type or return type) is considered to <i>exclude</i> {@code null}
+ * as a value; rarely needed within {@linkplain NullMarked null-marked} code.
  *
- * <p>This annotation has two primary purposes:
+ * <p>This annotation serves two primary purposes:
  *
  * <ul>
  *   <li>To mark any sporadic non-null type usages inside a scope that is not ready to be fully
@@ -64,10 +65,11 @@ import java.lang.annotation.Target;
  *
  * <p>Because {@code MyOptional} accepts only non-null type arguments, we need both {@code
  * maybeNulls.firstNonNull()} and {@code nonNulls.firstNonNull()} to produce the same return type:
- * {@code MyOptional!<String!>} (see <a href="Nullable.java#notation">notation</a>). However, as
- * specified above, they won't do that. In fact, there is a problem with the {@code firstNonNull()}
- * signature, since the type argument {@code String?} would not meet the requirements of {@code
- * MyOptional}'s type parameter.
+ * {@code MyOptional!<String!>} (see <a
+ * href="https://github.com/jspecify/jspecify/wiki/notation#shorthand-notation">notation</a>).
+ * However, as specified above, they won't do that. In fact, there is a problem with the {@code
+ * firstNonNull} signature, since the type argument {@code String?} would not meet the requirements
+ * of {@code MyOptional}'s type parameter.
  *
  * <p>The solution is to <b>project</b> the type argument to its non-null counterpart:
  *
@@ -90,7 +92,7 @@ import java.lang.annotation.Target;
  *
  * <h2>Where it is not applicable</h2>
  *
- * <p>{@code @NonNull} is not applicable in all the <a href="Nullable.html#applicability">same
+ * <p>{@code @NonNull} is inapplicable in all the <a href="Nullable.html#applicability">same
  * locations</a> as {@link Nullable}.
  */
 @Documented
--- a/src/main/java/org/jspecify/nullness/NullMarked.java
+++ b/src/main/java/org/jspecify/nullness/NullMarked.java
@@ -73,13 +73,14 @@ import java.lang.annotation.Target;
  *             type argument in {@code MyList<@Nullable Foo>} is out-of-bounds, so the list elements
  *             are always non-null. (<a href="https://bit.ly/3ppb8ZC">Why?</a>)
  *       </ul>
- *   <li>When a type variable has a nullable upper bound, such as the {@code E} in {@code class Foo<E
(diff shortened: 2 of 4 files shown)


src/main/java/org/jspecify/nullness/NonNull.java
```diff
@@ -23,10 +23,11 @@
 import java.lang.annotation.Target;
 
 /**
- * Indicates that the annotated type usage is considered to exclude {@code null} as a value; rarely
- * needed within {@linkplain NullMarked null-marked} code.
+ * Indicates that the annotated <a href="http://github.com/jspecify/jspecify/wiki/type-usages">type
+ * usage</a> (commonly a variable type or return type) is considered to <i>exclude</i> {@code null}
```
cpovirk, line 27: (We might have had this conversation already?)

It's certainly defensible to refer to fields and parameters as "variables" (e.g., [`VariableElement`](https://docs.oracle.com/en/java/javase/19/docs/api/java.compiler/javax/lang/model/element/VariableElement.html)). However, I think most people are likely to _initially_ read "variable" as referring to a local variable and in particular to the "top-level" type of that local variable. We of course explain elsewhere that that's _not_ where `@NonNull` can be used, but I still feel that we're getting off on the wrong foot.

I suggest changing "variable" to "parameter" or changing the whole parenthetical to something like "for example, a return type."

(Obviously, to the degree that you buy that, it applies to `@Nullable`, too.)

src/main/java/org/jspecify/nullness/Nullable.java
```diff
@@ -46,124 +47,61 @@
...
- *       commonly useful).
+ *   <li>On a <b>type argument</b>: A type usage of "nullable string" appears <i>within</i> the
+ *       compound type {@code List<@Nullable String>}. No matter how this type is used (return type,
+ *       etc.), this means the same thing: every appearance of {@code E} in {@code List}'s member
+ *       signatures will be considered nullable. For a list, this means it may contain null
+ *       <i>elements</i>. If the list reference itself might be null as well, we can write
+ *       {@code @Nullable List<@Nullable String>}, a "nullable list of nullable strings".
+ *   <li>On the upper bound of a <b>type parameter</b>: For example, as seen in {@code class List<E
+ *       extends @Nullable Object>}. This means that a <i>type argument</i> supplied for that type
+ *       parameter is permitted to be nullable if desired: {@code List<@Nullable String>}. A
```
cpovirk, line 77: Maybe cut "if desired?"

src/main/java/org/jspecify/nullness/Nullable.java
```diff
@@ -46,124 +47,61 @@
...
+ *       non-null type argument (as in {@code List<String>}) is always acceptable.
```
cpovirk, line 78: Maybe it's worth making the contrast here stronger? It's "always" acceptable not just in the situation we described but even in other situations. Which, yeah, is what "always" means... and maybe I'm advocating for a slippery slope, at the end of which the `@Nullable` Javadoc describes both what `@Nullable` means and what not-`@Nullable` means....

"`It is also still permitted to be non-null (as in {@code List<String>}), since non-null type arguments are permitted for every type parameter?`"

But again, I'm not sure if there's even a problem with the text as you have it.

src/main/java/org/jspecify/nullness/NonNull.java
```diff
@@ -23,10 +23,11 @@
 import java.lang.annotation.Target;
 
 /**
- * Indicates that the annotated type usage is considered to exclude {@code null} as a value; rarely
- * needed within {@linkplain NullMarked null-marked} code.
+ * Indicates that the annotated <a href="http://github.com/jspecify/jspecify/wiki/type-usages">type
```
cpovirk, line 26: While I've programmed you to avoid https for jspecify.dev links, we can use it for github.com links throughout.

cpovirk (commented):
Looks good. My only significant comment is on the term "variable."

src/main/java/org/jspecify/nullness/NonNull.java
scolsen: I agree.

src/main/java/org/jspecify/nullness/NullMarked.java
```diff
@@ -73,13 +73,13 @@
...
  *             are always non-null. (<a href="https://bit.ly/3ppb8ZC">Why?</a>)
  *       </ul>
- *   <li>A usage of a <b>nullable-bounded type variable</b> (such as the {@code E} in {@code class
- *       Foo<E extends @Nullable Bar>}) is not considered nullable, non-null, or even of
- *       "unspecified" nullness. Rather it has <b>parametric nullness</b>. In order to support both
- *       nullable and non-null type arguments safely, the {@code E} type itself must be handled
- *       <i>strictly</i>: as if nullable when "read from", but as if non-null when "written to".
- *       (Contrast with {@code class Foo<E extends Bar>}, where usages of {@code E} are simply
- *       non-null, just like usages of {@code String}.)
+ *   <li>When a type variable has nullable upper bound, such as the {@code E} in {@code class Foo<E
```
scolsen, line 76: nit: "nullable upper bound" --> "a nullable upper bound"

src/main/java/org/jspecify/nullness/NullMarked.java
```diff
@@ -73,13 +73,13 @@
...
+ *       extends @Nullable Bar>}), an unannotated usage of this type variable is not considered
+ *       nullable, non-null, or even of "unspecified" nullness. Rather it has <b>parametric
+ *       nullness</b>. In order to support both nullable and non-null type arguments safely, the
```
scolsen, line 79: wdyt about using the notion of "type variance" here instead of parametricity? It feels closer to me to what we're talking about. When I hear parametric I think "can be passed as argument" or at least, "determined by argument" which is not really true here, based on what the text says.

src/main/java/org/jspecify/nullness/NullUnmarked.java
```diff
@@ -49,10 +49,9 @@
  * <h2>Null-marked and null-unmarked code</h2>
  *
  * <p>{@link NullMarked} and this annotation work as a pair to include and exclude sections of code
- * from null-marked status.
- *
- * <p>Code is considered null-marked if the most narrowly enclosing element annotated with either of
- * these two annotations is annotated with {@code @NullMarked}.
+ * from null-marked status (respectively). Specifically, code is considered null-marked if the most
+ * narrowly enclosing element annotated with either of these two annotations is annotated with
+ * {@code @NullMarked}.
```
scolsen, line 54: I find this sentence vague and confusing. What about:

```
Specifically, code is considered *null-marked* if the most narrowly enclosing element is annotated with `@NullMarked`, otherwise, it is considered *null-unmarked*.
```

This would require moving the first sentence of the next paragraph into this sentence.

scolsen (approved):

src/main/java/org/jspecify/nullness/NonNull.java
kevinb9n: Yeah great point, done x2

src/main/java/org/jspecify/nullness/Nullable.java
kevinb9n: If this is a "maybe" I'll lean toward keeping it for emphasis.

src/main/java/org/jspecify/nullness/Nullable.java
kevinb9n: I made a slight rewording that might make it a tiny bit better.

src/main/java/org/jspecify/nullness/NullMarked.java
kevinb9n: fixed

src/main/java/org/jspecify/nullness/NullMarked.java
kevinb9n: this'll have to be a broader discussion than just this PR.

src/main/java/org/jspecify/nullness/NullUnmarked.java
kevinb9n: (For starters, I made the current wording slightly more accurate and *more* awkward.)

I think the reword doesn't mean the same thing. The most narrowly enclosing element might not be annotated. Also the "code" we're talking about could be a method signature itself, which might be annotated `@NullUnmarked`.

Frankly, trying to talk about "code" might just not be working.

Here's a totally different approach:

"The signature and body of a `@NullMarked` element is considered null-marked code, and the signature and body of a `@NullUnmarked` element is considered null-unmarked code, where more narrowly enclosing elements have precedence over those at a broader scope like a module. Finally, the signature and body of an element having neither annotation nor being enclosed by any element having either annotation is considered null-unmarked."

Or we could write it more algorithmically, saying "to determine whether code is null-marked, <take these steps>".

it's crazy how much harder this is to write than to understand. It's just inclusion/exclusion.

src/main/java/org/jspecify/nullness/NonNull.java
kevinb9n: Fixed

kevinb9n:
I think this is ready for another look as of now?

src/main/java/org/jspecify/nullness/NullUnmarked.java
kevinb9n: I've left a TODO comment about improving this, and am gonna go ahead and merge this now -- still happy to make more changes tomorrow.

kevinb9n:
Seems okay to submit this and keep iterating on it.

PR #327, opened 2022-11-11
Switch back to Standard (from Tool-Independent) in titles

kevinb9n:
I'm still using tool-independent in a couple places where it seems like it wants to be more descriptive.


artempyanykh (approved):
Thanks @kevinb9n!

PR #329, opened 2022-11-14
Tighten wording to similar length as the other two.

kevinb9n:


kevinb9n:
Still postponing installing and figuring out docusaurus so I'm only guessing that this will get the text to fit on 3 lines. Thoughts on the wording?

kevinb9n:
Once again I can't see this as worth waiting for review on. These things we can always fix forward.

PR #331, opened 2022-11-15
more small tweaks to home page

kevinb9n:


kevinb9n:
Happy to handle post-merge comments.

Issue #260, opened 2022-08-09
final package names/locations


kevinb9n:
# edit: This issue will now be decided in [this document](https://docs.google.com/document/d/13i7hlSF80EuY2IdR6BpBtIWPVJgE2V29wh6PapC1hKQ/edit#)

Sincere apologies to the public who cannot access that at this time. Feel free to append your arguments here and we will factor them in.

---

Currently the four nullness annotations are in `org.jspecify.nullness`. Where does `@Implies` go?

* `org.jspecify`
* `org.jspecify.nullness`
* `org.jspecify.meta`
* `org.jspecify.shared`
* `org.jspecify.implies`
* `org.jspecify.annotations` and move nullness in there too
* other

kevinb9n:
I feel we will create constant headaches for ourselves by trying to aggressively categorize all our annotations, and having `org.jspecify.nullness` (as we "working-decided" in #160) is putting us on that path. Could we stomach a move to `org.jspecify.annotations`?

cpovirk:
Thanks for opening this. The thought had crossed by mind during the review for https://github.com/jspecify/jspecify/pull/256, but it immediately left.

The question of the best package structure is still one that I don't feel I have a firm opinion on. (History is available in #160, as linked above, as well as #1, which is linked from #160. See also [this thread](https://groups.google.com/g/jspecify-dev/c/w-oOKwR6tjk) and [these meeting notes](https://docs.google.com/document/d/1hr8Ld9cRMFksvA_Uzh61o4bdiVMXl084AR5QAkokEn0/edit#heading=h.lt5u79bikyk9).)

The main claim I would make on that front is that it's at least not obviously a terrible user experience for us to have any of the following scenarios:
- all the annotations in one package
- all the annotations in subpackages (`nullness`, `meta`, `resultuse`(???), ...)—though I do see how this could get tricky to name
- a mix of annotations in subpackages (`nullness`, `meta`, ...) and annotations in the "everything else" superpackage (or `misc`-like subpackage)

[...]

PR #335, opened 2022-11-16
Move annotations from `.nullness` to `.annotations`, leaving deprecated versions behind

kevinb9n:


kevinb9n:
Oh: decision was in #260

src/main/java/org/jspecify/nullness/package-info.java
```diff
@@ -14,42 +14,5 @@
...
- * another might likewise determine that a declaratively nullable expression can take on only
- * non-null values. In both cases the declarative knowledge is <i>correct</i>, but the inferred
- * knowledge is both correct and more specific.
- *
- * <p>On the other end, the tools might even enforce <i>nothing at all</i>. In particular, your
- * annotated code (or other code dependent on its annotated APIs) might be compiled and run without
- * any appropriate tool even in use. Therefore adopting JSpecify annotations is not a replacement
- * for <b>explicitly checking arguments</b> at runtime.
- */
+/** Old location of {@link org.jspecify.nullness}. */
```
cpovirk, line 17:
```suggestion
/** Old location of nullness annotations, which are now in {@link org.jspecify.annotations}. */
```

cpovirk (approved):

Stephan202:
I filed google/google-java-format#869 so that the annotations in the new package are formatted like the ones in the old package.

kevinb9n:
Thank you! Liam should be able to get that in and I will sit here and decide how embarrassed to be about it.

PR #336, opened 2022-11-17
docs: Add a missing svg asset

artempyanykh:
This is the only asset that was not svg. The svgs under docs/static/img can be imported into many vector graphics editors (e.g. Figma) and should be reasonably easy to tweak as they have all the original shapes and whatnot unlike pngs/jpegs.


kevinb9n (commented):
Thanks!

PR #337, opened 2022-11-17
Fix references to o.j.nullness -> o.j.annotations

kevinb9n:
I believe this gets them all, from this tree.


cpovirk:
The diff shows changes from your previous PR, but I'm optimistic that merging is going to Just Work.

Issue #307, opened 2022-10-27
Postpone implication to post-1.0


Mooninaut:
* Implication / inheritance is an extremely complicated and subtle problem with myriad pitfalls
* JSpecify 1.0 should be simple and usable to encourage uptake
* When all you have is annotations, everything looks like an annotation
* There are other, potentially better ways than meta-annotations to define cross-annotation and cross-annotation-library semantic hierarchies that have not been sufficiently explored
  * Define relationships somewhere else (such as tool config files)
  * Single-inheritance or no inheritance instead of unrestricted implication

Allowing annotations to define relationships to other annotations is inflexible and users will need to be able to override those relationships in the tools anyway, so why define them in the annotations in the first place?

kevinb9n:
Thanks for this, am working on response.

kevinb9n:
We would love nothing more, provided we can accomplish our goals without it.

You deserve responses to your points, but I want to stress this: you’re far more likely to find a way forward on this through engaging with my summation of the issue in the next comment. The particular mini-conversations in this comment are more of a sideshow.


> * Implication / inheritance is an extremely complicated and subtle problem with myriad pitfalls
> * JSpecify 1.0 should be simple and usable to encourage uptake
> * When all you have is annotations, everything looks like an annotation

Your points are so general I see nothing to confirm or refute. Of course we prefer simple and usable over complicated and subtle. The whole game is to show why the alternatives before us are like the one or the other.


> * There are other, potentially better ways than meta-annotations to define cross-annotation and cross-annotation-library semantic hierarchies that have not been sufficiently explored
>   
>   * Define relationships somewhere else (such as tool config files)

My next comment will argue that this doesn't solve the problem `@Implies` exists to solve.


>   * Single-inheritance or non-recursive grouping instead of unrestricted implication

Any further discussion of this should go to [#237](237) with the rest. But I doubt it would be productive. The conversation ran its course; there were ample opportunities to make the case, and in the end we simply don’t agree that that case has been made.


> Allowing annotations to define relationships to other annotations is inflexible 

Also too general to respond to.


> and users will need to be able to override those relationships in the tools anyway, so why define them in the annotations in the first place?

Our job is to define annotations that capture unambiguous information that enough developers want to express and that enough tools want to know. If we’re on net balance failing at that, we want to change course. If we’re on net succeeding, it's okay that some tools might ignore the information some of the time.

kevinb9n:
[...]

PR #339, opened 2022-11-17
Goodbye, @Implies

kevinb9n:
#307 is the relevant bug, though I might recast rather than close it.


cpovirk (approved):
Do you normally do Javadoc regeneration in the same PR (until we get it automated)? If so, you'll presumably want to do it here. If not... you'll presumably want to push the "Squash and merge" button now :)

kevinb9n:
derp! pushed.

PR #343, opened 2022-11-22
move /start-here to /docs/start-here

kevinb9n:
... since it's permanently useful and this makes it appear in the sidebar.


docs/docs/start-here.md
```diff
@@ -0,0 +1,106 @@
...
+* The [Javadoc], which is reasonably thorough and specific, though not an
+  organized walkthrough
+* The current [User Guide] draft, which is a better choice for reading
+  top-to-bottom, but is currently a bit outdated and in need of improvement.
+  It will change quite a bit before 1.0.
+
+### Then if you're *really* interested
+
+* The [specification], written to be understood by owners of compilers and
+  static analysis tools, and currently a bit outdated
```
scolsen, line 48: Probably better to mention that it's outdated only on the spec itself and just point to it here. If we say that it's outdated here we'll need to remember to go back and also update this doc when it's no longer outdated.

docs/docs/start-here.md
```diff
@@ -0,0 +1,106 @@
...
+* Do you use any libraries or tools that you think should use/support JSpecify
+  annotations? Please tell them about us!
+* Give some thought to what factors would make your own projects more or less
+  likely to adopt, and let us know.
+* [File an issue] to request a feature or if something is wrong.
+* Don't forget to Star and Watch our [github repo]!
+
+(TODO: code of conduct link.)
+
+[file an issue]: https://github.com/jspecify/jspecify/issues/new
```
scolsen, line 95: I personally feel this style of reference link is very brittle. Mind giving these identifiers, numeric or otherwise? e.g.:

inline in the paragraphs
[github repo][1]

[1]: https://github.com.... 

or give them stable semantic names:

[github repo][gh-repo]
[gh-repo]: https://....

docs/docs/start-here.md
kevinb9n: Fair point, but... I do feel better warning about this right here. So I filed #344 with a reminder about this.

PR #349, opened 2022-11-29
healthy round of javadoc updates

kevinb9n:
Some cosmetic changes, but also conveying some design aspect that were missing. (Not finished with that whole audit, so expect more changes.)


src/main/java/org/jspecify/annotations/NullMarked.java
```diff
@@ -41,30 +41,27 @@
...
- *       bounds with no further restriction. A <b>lower-bounded wildcard</b> ({@code List<? super
- *       String>}) similarly has no <i>upper</i> bound of its own. Therefore, since <i>nullable</i>
- *       {@code Object} is the new "<a href="Nullable.html#subtypes">top type</a>", either list
- *       might have null elements, even though the word {@code @Nullable} is nowhere in sight. On
- *       the other hand, in {@code List<? extends Object>} a bound <i>is</i> provided, and as usual,
- *       being unannotated it is considered non-null. This means that while {@code List<?>} and
- *       {@code List<? extends Object>} have always been identical as base types, they are no longer
- *       identical as <a href="Nullable.html#augmented-types">augmented types</a>. (<a
+ *       {@code extends} or {@code super}) will represent <i>any</i> type within the type
+ *       parameter's bounds, with no further restriction. A <b>lower-bounded wildcard</b> ({@code
```
eamonnmcmanus, line 57: This is a fairly confusing area (I know I was confused by it), so I think it might be worth spelling it out in even more detail. You might add text like this, for example:

> ...with no further restriction. So if {@code List} is declared as {@code interface List<E extends @Nullable Object>}
> then {@code List<?>} is equivalent to {@code List<? extends @Nullable Object>}.
> If {@code List} is {@code interface List<E>} or {@code List<E extends Object>}
> then {@code List<?>} is equivalent to {@code List<? extends Object>}.

eamonnmcmanus (commented):
LGTM, except one thing that is only tangentially related to these changes...

src/main/java/org/jspecify/annotations/NullMarked.java
kevinb9n: Well, this prompted me to attempt a rewording, see what you think.

src/main/java/org/jspecify/annotations/NullMarked.java
```diff
@@ -41,30 +41,27 @@
...
+ *       List<? super String>}) similarly has no <i>upper</i> bound of its own. Therefore, since
+ *       <i>nullable</i> {@code Object} is the new "<a href="Nullable.html#subtypes">top type</a>",
+ *       either list might have null elements, even though the word {@code @Nullable} is nowhere in
```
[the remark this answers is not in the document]
kevinb9n: This text is gone.

kevinb9n:
Sometime I will work on an article that tries to really explain The Deal With Generics And Nullness. Then this can link to that instead of straight to the design faq.

eamonnmcmanus (approved):
LGTM

Issue #348, opened 2022-11-29
Remove `@Target(MODULE)` from `@NullUnmarked`?


kevinb9n:
Seems like it can't serve any purpose and might confuse someone (who rightly thinks "but I thought it was the default").

cpovirk:
That sounds sensible, and I like it a lot for an additional reason:

Removing `MODULE` would at least marginally decrease the chances that users would encounter the crash discussed in #302. In fact, if we imagine a world in which users use `@NullMarked` only at the class, package, and module levels, while perhaps using `@NullUnmarked` on individual methods, then removing `MODULE` would _eliminate_ #302. In practice, I'm sure it wouldn't _totally_ eliminate it. But it might help a lot.

The main counterargument I can see is that an enterprising organization might try to "flip the default" for now code so that _all_ code is considered null-marked unless otherwise annotated :) Then they might theoretically want to unmark their already existing modules. But I'm inclined not to worry about that, thanks to some combination of:

- For that to work, every tool that the organization uses would need to support "null-marked by default."
- Some people don't make use of modules.
- It doesn't seem that hard to put in place enforcement that all _new_ modules are `@NullMarked`.
- We might even look askance at the idea that code without `@NullMarked` should be treated as null-marked in general.

kevinb9n:
This is my askance look.

cpovirk:
https://github.com/jspecify/jspecify/pull/351 did this and updated docs accordingly.

tbroyer:
> Arguably, my use case can be solved differently:
> 
> * By adding a `package-info.java` in the test scope, annotated with `@NullUnmarked`
> * By configuring ErrorProne with `-XepExcludedPaths:.*/src/test/java/.*`

or by disabling NullAway when compiling your test code

<details>
<summary>With Maven</summary>

(untested)

```xml
<plugin>
  <artifactId>maven-compiler-plugin</artifactId>
  <configuration>
    <compilerArgs combine.children="append">
      <arg>-XDcompilePolicy=simple</arg>
      <arg>--should-stop=ifError=FLOW</arg>
    </compilerArgs>
    <executions>
      <execution>
        <id>default-compile</id>
        <configuration>
          <compilerArgs>
            <arg>-Xplugin:ErrorProne -Xep:NullAway:ERROR</arg>
          </compilerArgs>
        </configuration>
      </execution>
      <execution>
        <id>default-testCompile</id>
        <configuration>
          <compilerArgs>
[...]

PR #351, opened 2022-11-30
Remove MODULE from NullUnmarked (per #348)

kevinb9n:


src/main/java/org/jspecify/annotations/NullMarked.java
```diff
@@ -104,7 +104,8 @@
  *   <li>Applying this annotation to an instance <b>method</b> of a <i>generic</i> class is
  *       acceptable, but is not recommended because it can lead to some confusing situations.
  *   <li>An advantage of Java modules is that you can make a lot of code null-marked with just a
- *       single annotation (before the {@code module} keyword).
+ *       single annotation (before the {@code module} keyword). {@link NullUnmarked} is not
```
cpovirk, line 107: The text above the list still suggests that NullUnmarked is usable on a module.

"and {@link NullUnmarked}, with one exception" or something?

And then you get to regenerate the Javadoc :)

cpovirk (approved):

Issue #350, opened 2022-11-30
Delete old annotations after -alpha2


kevinb9n:
Delete old annotations after -alpha2

PR #353, opened 2022-12-01
Export both old and new packages from the module.

cpovirk:
We'll want to clean this up as part of
https://github.com/jspecify/jspecify/issues/350, too.

### DIFF
--- a/src/java9/java/module-info.java
+++ b/src/java9/java/module-info.java
@@ -15,4 +15,5 @@
  */
 module org.jspecify {
   exports org.jspecify.annotations;
+  exports org.jspecify.nullness;
 }


cpovirk:
As always, it's possible that no one cares about this. But it's easy enough to release 0.3.0-alpha-3 with this fix, so I figure we might as well do it?

(Related: I notice that we never included an `exports` line for `org.jspecify.meta` back when it was present in 0.3.0-alpha-1 :))

cpovirk:
> We'll want to clean this up as part of https://github.com/jspecify/jspecify/issues/350, too.

I just checked, and javac will catch us if we forget :)

kevinb9n:
I should probably learn about modules one day.

PR #366, opened 2023-02-01
Sample for an annotation on a multidimensional primitive array.

cpovirk:
This demonstrates the problem that
https://github.com/jspecify/nullness-checker-for-checker-framework/pull/53
fixes.


wmdietl (approved):
Change itself looks good, thanks for adding the test case!

One general question: do you want to keep the copyright years up on line 2 up-to-date and change this to `2020-2023`?

cpovirk:
Thanks. The guidelines we have from Google say to leave the oldest date in place. I don't know exactly why. If you want to read more about how I don't know exactly why, see [here](https://github.com/jspecify/jspecify/pull/177#issuecomment-780163773) :)

PR #371, opened 2023-02-28
Run reference checker tests on PRs

### LINKED ISSUE #370: Run reference checker tests on each PR
For each PR (and post-merge), add a job to run tests (maybe just conformance tests?) for [the reference checker](https//github.com/jspecify/nullness-checker-for-checker-framework).

That way we can know whether the reference checker needs to be updated for any given change in this repo.

netdpb:
Fixes #370.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,8 +22,9 @@ jobs:
           cache: gradle
       - run: |
           ./gradlew spotlessCheck build publishToMavenLocal --no-daemon
-  conformance-tests:
-    name: Run Conformance Tests
+  reference-checker-tests:
+    if: github.base_ref == 'main' || github.base_ref == 'samples-google-prototype'
+    name: Run Reference Checker Tests
     runs-on: ubuntu-latest
     steps:
       - name: Check out the code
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -22,3 +22,34 @@ jobs:
...
+        with:
+          distribution: temurin
+          java-version: 17
+      - name: Run Conformance Tests
+        if: github.base_ref == 'main'
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: ConformanceTest
+          build-root-directory: nullness-checker-for-checker-framework
+      - name: Run Samples Tests
```
scolsen, line 50: Should we make this bit and the conformance tests separate jobs so that we can run them in parallel?

netdpb: Which bit? The "samples" tests? We're only running the samples tests for PRs to the samples-google-prototype branch, and we're only running the conformance tests for PRs to the main branch.

scolsen: Yeah that's the one I meant--github's contextualization of comments is so-so. And ah, yeah the conditional takes care of that. Still, it might be more modular to factor these out to separate jobs—then again, we'd just be doing extra work.

PR #381, opened 2023-05-03
Update spotless version, format annotations, and apply to all .java files

wmdietl:
When I saw [this comment](https://github.com/jspecify/jspecify/pull/378#discussion_r1178158343) I was wondering why spotless didn't do this automatically.

The first commit in this PR updates the configuration and the second commit adds the resulting changes.


wmdietl:
Related PR to fix the reference checker tests: https://github.com/jspecify/nullness-checker-for-checker-framework/pull/72

PR #383, opened 2023-05-05
Fill in a couple license years that had been autogenerated.

cpovirk:


cpovirk:
The other thing I notice is that `TypeVariableUnspecToSelfUnspec.java` is formatted to >100 lines in one place. My wild guess is that `addTypeAnnotation` is implemented by transforming `@NullnessUnspecified` into `@Nullable`, formatting, and transforming it back :) I can live with that—I just thought it was interesting.

wmdietl (approved):
I looked at the history of these two files and confirmed that these are the years those files appeared in the repo.

Regarding the long line: Spotless [knows about `NullnessUnspecified`](https://github.com/jspecify/jspecify/blob/cfb1fb279e57ff79dc78cd4c5c438e1b70f4a376/gradle/format.gradle#L8) and should format it like other type annotations.
So this must be an issue with how Spotless post-processes Google Java Format to put type annotations on the same line as their types.

cpovirk:
Thanks. Yep, it looks like this is the Spotless post-processing logic: https://github.com/diffplug/spotless/blob/5e75cb166ed97abc32d2337099465828e5797602/lib/src/main/java/com/diffplug/spotless/java/FormatAnnotationsStep.java

I'm going to not worry about it.

PR #384, opened 2023-05-09
Trigger conformance test report update

netdpb:


netdpb:
Since this updates workflows to use repository variables and secrets, and since those are ["not passed to workflows that are triggered by a pull request from a fork"](https://github.com/jspecify/jspecify/settings/secrets/actions), I guess we'll test after merging.

cpovirk:
Hmm, actually, we might have the option of having you push your changes to a _branch_ of _this repo_ and then create a PR based on that. Are you interested in giving that a try, or should we just merge this and figure that we can always try that approach for any necessary future PRs?

netdpb:
See https://github.com/jspecify/jspecify/pull/386.

PR #387, opened 2023-05-10
Reformat new samples to fix the build.

cpovirk:


netdpb (approved):

cpovirk:
I've been having this problem today of seeing failed checks, like "Build / JDK 17 (push)" above, but then seeing no information when I click through, only "This job failed"....

PR #388, opened 2023-05-12
start-here page: link to reference-checker and portray it as ready

kevinb9n:
Plus random other rewordings

### DIFF
--- a/docs/docs/start-here.md
+++ b/docs/docs/start-here.md
@@ -10,7 +10,7 @@ Quick links: [Release] | [GitHub] | [UserGuide] | [Javadoc] | [Spec] | [Wiki] | 
 
 [A group of organizations](/about) are working together to define a common set
 of annotation types for use in JVM languages, to improve static analysis and
-language interoperation. We've been focusing on nullness analysis first.
+language interoperation. Our initial focus is on nullness analysis.
 
 This includes providing both an artifact of annotation types your code can
 depend on, and *precise* specifications of their semantics.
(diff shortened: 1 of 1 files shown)


docs/docs/start-here.md
```diff
@@ -56,35 +55,36 @@ then [send us a mail](mailto:jspecify-discuss@googlegroups.com)
...
+  annoying our current design choices are for you (which you should let us
+  know!). However, this tool is still a work in progress, and is not at *full*
+  conformance with our own specification quite yet.
 * **Version 1.0** will commit **permanently** to the annotation APIs, and they
-  will be safe to depend on. From that point, no incompatible changes from a
-  compilation perspective will happen, and any further specification adjustments
-  will not break null-checking for you or your clients except to improve rare
-  corner cases. We have no prediction for when this release will be. We hope to
-  get as much of your feedback as we can before then.
+  will be safe to depend on. From that point, no compile-breaking changes will
```
cpovirk, line 66: minor and preexisting point:

I wonder if users will in fact understand "compile-breaking" to mean "no breakages from the stock javac" as opposed to "no breakages produced by nullness checkers in Java or Kotlin clients."

Is it worth saying something like "We will maintain source and binary compatibility" instead?

cpovirk (approved):

docs/docs/start-here.md
kevinb9n: I've spent some hems and haws over this. I think I'm gonna leave it as a future improvement for now, thanks for approving. :-)

PR #394, opened 2023-06-20
Make a section in `@NonNull` match the one in `@Nullable`.

cpovirk:


cpovirk:
I'm wondering if the difference is just historical. (But I haven't checked.) Alternatively, there may be a reason behind it that didn't occur to me in all 5 seconds of thought that I gave it.

kevinb9n (approved):
We actually use the terms "applicable" and "non-applic".

Thanks!

PR #409, opened 2023-08-18
Add a few sink-type assertions to prove they work.

netdpb:


conformance/Basic.java
```diff
@@ -20,13 +20,24 @@
...
-    // test:cannot-convert:Object? to Object!
     // test:expression-type:Object?:nullable
+    // test:sink-type:Object!:cannotConvertNullableToNonNull#return
+    // test:cannot-convert:Object? to Object!
     return nullable;
   }
 
   @Nullable Object canConvertNonNullToNullable(@NonNull Object nonNull) {
     // test:expression-type:Object!:nonNull
+    // test:sink-type:Object?:canConvertNonNullToNullable#return
```
netdpb, line 31: We could skip the name of the function here, since it should be unambiguous.

```suggestion
    // test:sink-type:Object?:return
```

cpovirk: Skipping SGTM.

cpovirk (approved):

Issue #411, opened 2023-08-22
Weird behavior Javadoc link


Ao-senXiong:
Hi,

When I first click on the Javadoc on this page: https://jspecify.dev/docs/start-here, it shows me the following page
<img width="847" alt="image" src="https://github.com/jspecify/jspecify/assets/82676488/3036dbcf-5633-4fdb-962d-19e7f394d1ce">
but I believe the link is right when I copy this link https://jspecify.dev/docs/api/org/jspecify/annotations/package-summary.html to a new tab of my browser, the Javadoc webpage appears correctly.

cpovirk:
Hmm, thanks. This was supposed to be fixed by https://github.com/jspecify/jspecify/pull/392. Supposedly our "Build Documentation" task for that commit was [successful](https://github.com/jspecify/jspecify/actions/runs/5203720414/job/14084848783), but I see no commits to https://github.com/jspecify/jspecify/tree/gh-pages since February.

Ah, it looks like https://github.com/jspecify/jspecify/pull/369 split "build" and "deploy" into separate steps, and it looks like the "deploy" step has been getting skipped. I think we've had trouble before with getting comparisons like [`github.ref == 'refs/head/main'`](https://github.com/jspecify/jspecify/commit/370977917111f8ae2aac8d5c5465408068671315#diff-9cf2000c53760d837a449f874e53f792819108d3a4bf346336d0f7d082deae2cR33) right. Maybe @scolsen or @netdpb knows offhand?

scolsen:
Looks like the appropriate reference is `refs/heads/main` not *head*. Barring that, might have something to do with the event choice (push) which I generally find difficult to reason about. Let's start w/ fixing the reference name.

scolsen:
https://github.com/jspecify/jspecify/pull/413 will possibly get us to the right place? 🤞

cpovirk:
Fixed by @scolsen in a series of PRs culminating in https://github.com/jspecify/jspecify/pull/416

PR #412, opened 2023-08-22
workflow: fix head ref identifier

scolsen:
It's refs/heads/main, not refs/head/main.

Attempt to fix the continuous deployment issue raised in https://github.com/jspecify/jspecify/issues/411#issuecomment-1688377040


cpovirk (approved):
Thanks!

PR #413, opened 2023-08-22
workflow: fix build and deploy docs workflow

scolsen:
Use upload/download artifact to share built doc results across jobs.

Each job runs in a separate image, so the deploy job wasn't able to use build results. Following https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts, we use the artifact rules to persist the build output across the two jobs. This consumes some storage quota, but since we were already (incorrectly?) using upload artifact, I stuck with it.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,6 +27,11 @@ jobs:
         run: npm run build
         working-directory: docs
 
+      - uses: actions/upload-artifact@v2
+        with:
+          name: document
+          path: docs/build
+
   # Deploys the documentation if this is a push to the main branch.
   deploy:
     name: Deploy Documentation
@@ -34,10 +39,9 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/upload-artifact@v2
-        with:
+       - uses: actions/download-artifact@v2
+         with:
           name: document
-          path: docs/build
 
       - uses: JamesIves/github-pages-deploy-action@4.1.5
         with:


scolsen:
unhelpfully, the workflow does not seem to be triggering now...

cpovirk:
Oh, interesting.

As for the "not running" part.... I think it ran on past PRs, right? Could it be that the `uses` sections all have to come before the "real" `steps` elements?

scolsen:
doesn't seem to be a requirement based on the examples here: https://github.com/jspecify/jspecify/pull/413/files

maybe it's a permissions issue? it would make sense to not allow random users to trigger an action that consumes quota/uploads an artifact....maybe we should just remerge all this into a single action after all....

cpovirk (approved):
It looks like other CI tasks ran, so I'm not sure why only this one would require an approval. But maybe it does. And even if it doesn't, we can merge this and see what happens.

cpovirk:
Still no results, but here goes...

cpovirk:
"You have an [error](https://github.com/jspecify/jspecify/actions/runs/5943136336/workflow) in your yaml syntax on [line 36](https://github.com/jspecify/jspecify/blob/40433ac2d4c4d0ab97bc569f28722c738b01f31f/.github/workflows/docs.yml#L36)"

scolsen:
ahhhhhh thanks! (the error is not actually on line 36; a few lines down I missed a space) https://github.com/jspecify/jspecify/pull/414

PR #420, opened 2023-11-21
Produce conformance test artifacts

netdpb:
The conformance tests are now a subproject, which produces two artifacts:

* `org.jspecify.conformance:conformance-tests-<version>-assertion-sources.jar` containing test sources.
* `org.jspecify.conformance:conformance-tests-<version>-deps.jar` containing compiled dependencies of those test sources, which will transitively require dependencies like the JSpecify annotations themselves.

The conformance test files are still also in their old location so as not to break the reference checker's build. Once that build is changed to depend on these artifacts (https://github.com/jspecify/jspecify-reference-checker/pull/112), the `conformance` directory can be deleted.

Part of https://github.com/jspecify/jspecify-reference-checker/issues/107.


cpovirk (approved):
I think I understand this one somewhat acceptably well. I'll keep trying to get there on the other one.

PR #428, opened 2023-12-11
add OSGi metadata and reproducible build

### LINKED ISSUE #362: Provide OSGI support
JSpecify annotations could also be used in OSGI environments if `MANIFEST.MF` would have bundle entries like e.g.
```
Bundle-ManifestVersion: 2
Bundle-Name: JSpecify annotations
Bundle-SymbolicName: org.jspecify.annotations
Bundle-Version: 0.3.0
Export-Package: org.jspecify.annotations
[...]
```
Please provide a solution and JSpecify annotations could be used out of the box.

chrisrueger:
Closes #362

This PR generates OSGi metadata during the build the following META-INF/MANIFEST.MF during a gradle build and should address #362 . In other words it creates a proper OSGi bundle.
It uses the bndtools Gradle Plugin (see [here](https://github.com/bndtools/bnd/blob/master/gradle-plugins/README.md)  and [here](https://plugins.gradle.org/plugin/biz.aQute.bnd.builder))

An additional side effect of this PR is that the build is made [reproducible](https://reproducible-builds.org/).

## Why?

Simplifies adoption for people using an OSGi environment. 

image

```
Manifest-Version: 1.0
Bnd-LastModified: 1702332812312
Bundle-Description: An artifact of well-specified annotations to power
  static analysis checks and JVM language interop.
Bundle-DocURL: https://jspecify.dev/docs/start-here
Bundle-ManifestVersion: 2
Bundle-Name: jspecify
Bundle-SymbolicName: org.jspecify.jspecify
Bundle-Vendor: jspecify.dev
Bundle-Version: 0.0.0.SNAPSHOT
Created-By: 17.0.9 (Eclipse Adoptium)
Export-Package: org.jspecify.annotations;version="0.0.0.SNAPSHOT"
Import-Package: java.lang,java.lang.annotation
Multi-Release: true
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Tool: Bnd-7.0.0.202310060912
```

More Detailed View using bndtools Jar-Viewer:

```
[MANIFEST]

Bnd-LastModified                        1702332812312
Bundle-Description                      An artifact of well-specified annotations to power static analysis checks and JVM language interop.
Bundle-DocURL                           https://jspecify.dev/docs/start-here
Bundle-ManifestVersion                  2
Bundle-Name                             jspecify
Bundle-SymbolicName                     org.jspecify.jspecify
Bundle-Vendor                           jspecify.dev
Bundle-Version                          0.0.0.SNAPSHOT
Created-By                              17.0.9 (Eclipse Adoptium)
Export-Package                          org.jspecify.annotations;version="0.0.0.SNAPSHOT"
Import-Package                          java.lang
                                        java.lang.annotation
Manifest-Version                        1.0
Multi-Release                           true
Require-Capability                      osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Tool                                    Bnd-7.0.0.202310060912


[IMPEXP]

Import-Package
  java.lang                              
  java.lang.annotation                   

Export-Package
  org.jspecify.annotations               {version=0.0.0.SNAPSHOT}

[CAPABILITIES]

Require-Capability
  osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version=1.8))}

[COMPONENTS]



[METATYPE]



[API USES]


[USES]

org.jspecify.annotations                


[USEDBY]



[LIST]

META-INF
  MANIFEST.MF
META-INF/versions
META-INF/versions <no contents>
META-INF/versions/9
  module-info.class
META-INF/versions/9/OSGI-INF
  MANIFEST.MF
org
org <no contents>
org/jspecify
org/jspecify <no contents>
org/jspecify/annotations
  NonNull.class
  NullMarked.class
  NullUnmarked.class
  Nullable.class
```

The bundle-Version is automatically taken from the `version` inside `gradle.properties`.


chrisrueger:
Not entirely sure about `Bundle-SymbolicName`

Do we want:
a) `Bundle-SymbolicName: org.jspecify.jspecify`
or 
b) `Bundle-SymbolicName: org.jspecify.annotations`

a) is currently implemented. It is based on the maven artifact GAV (`org.jspecify:jspecify:jar:0.3.0` e.g. see [here](https://mvnrepository.com/artifact/org.jspecify/jspecify/0.3.0)

ben-manes:
As a ignorant developer supporting a library with OSGi users, I found adding a unit test helpful comfort, wdyt?
([example](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/test/java/com/github/benmanes/caffeine/OSGiTest.java#L37-L52) using pax-exam)

This project might want reproducible jars? I don't believe bnd does that by default (my [build snippet](https://github.com/ben-manes/caffeine/blob/bf7d640fba235fbfa2734c0de68e69e25c0405d4/gradle/plugins/src/main/kotlin/lifecycle/java-library-caffeine-conventions.gradle.kts#L49-L78) if helpful).

agentgt:
> As a ignorant developer supporting a library with OSGi users, I found adding a unit test helpful comfort, wdyt?
([example](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/test/java/com/github/benmanes/caffeine/OSGiTest.java#L37-L52) using pax-exam)

I'm going to steal that 😄 for my own project.

Also I think making this project reproducible is more important than OSGi so hopefully that will be the case. I didn't see it in the list: https://github.com/jvm-repo-rebuild/reproducible-central#readme

@ben-manes do you know if it is currently reproducible (jspecify jars)?

chrisrueger:
@ben-manes thanks a lot for the pointers. I tried adding your suggestions about reproducible builds - although I must admit I have no experience with it, so could you double check it?

Regarding unit test: generally good idea. What would you test for here?

Last point:

> As a ignorant developer
Sorry, german here, don't get it ;) What's is with this "ignorant"?

ben-manes:
>  could you double check it?

lgtm (but just an observer)

<details><summary>Inspection</summary>

```console
➜  Downloads gh repo clone jspecify/jspecify
fatal: destination path 'jspecify' already exists and is not an empty directory.
failed to run git: exit status 128
➜  Downloads cd jspecify
➜  jspecify git:(add-osgi-metadata) gh pr checkout 428
From https://github.com/jspecify/jspecify
 * branch            refs/pull/428/head -> FETCH_HEAD
Already up to date.
➜  jspecify git:(add-osgi-metadata) ./gradlew -q jar
➜  jspecify git:(add-osgi-metadata) cd build/libs
➜  libs git:(add-osgi-metadata) unzip -l jspecify-0.0.0-SNAPSHOT.jar
Archive:  jspecify-0.0.0-SNAPSHOT.jar
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  02-01-1980 00:00   META-INF/
      626  02-01-1980 00:00   META-INF/MANIFEST.MF
        0  02-01-1980 00:00   META-INF/versions/
        0  02-01-1980 00:00   META-INF/versions/9/
        0  02-01-1980 00:00   META-INF/versions/9/OSGI-INF/
      145  02-01-1980 00:00   META-INF/versions/9/OSGI-INF/MANIFEST.MF
      187  02-01-1980 00:00   META-INF/versions/9/module-info.class
        0  02-01-1980 00:00   org/
        0  02-01-1980 00:00   org/jspecify/
        0  02-01-1980 00:00   org/jspecify/annotations/
      434  02-01-1980 00:00   org/jspecify/annotations/NonNull.class
      498  02-01-1980 00:00   org/jspecify/annotations/NullMarked.class
      488  02-01-1980 00:00   org/jspecify/annotations/NullUnmarked.class
      436  02-01-1980 00:00   org/jspecify/annotations/Nullable.class
---------                     -------
     2814                     14 files
➜  libs git:(add-osgi-metadata) unzip jspecify-0.0.0-SNAPSHOT.jar
Archive:  jspecify-0.0.0-SNAPSHOT.jar
   creating: META-INF/
  inflating: META-INF/MANIFEST.MF
   creating: META-INF/versions/
   creating: META-INF/versions/9/
   creating: META-INF/versions/9/OSGI-INF/
  inflating: META-INF/versions/9/OSGI-INF/MANIFEST.MF
  inflating: META-INF/versions/9/module-info.class
   creating: org/
   creating: org/jspecify/
   creating: org/jspecify/annotations/
  inflating: org/jspecify/annotations/NonNull.class
  inflating: org/jspecify/annotations/NullMarked.class
  inflating: org/jspecify/annotations/NullUnmarked.class
  inflating: org/jspecify/annotations/Nullable.class
➜  libs git:(add-osgi-metadata) cat META-INF/MANIFEST.MF
Manifest-Version: 1.0
Bundle-Description: An artifact of well-specified annotations to power
  static analysis checks and JVM language interop.
Bundle-DocURL: https://jspecify.dev/docs/start-here
Bundle-License: https://www.apache.org/licenses/LICENSE-2.0
Bundle-ManifestVersion: 2
Bundle-Name: jspecify
Bundle-SymbolicName: org.jspecify.jspecify
Bundle-Vendor: jspecify.dev
Bundle-Version: 0.0.0.SNAPSHOT
Export-Package: org.jspecify.annotations;version="0.0.0.SNAPSHOT"
Import-Package: java.lang,java.lang.annotation
Multi-Release: true
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

chrisrueger:
> lgtm (but just an observer)

Thanks a lot.

bnd.bnd
```diff
@@ -0,0 +1,10 @@
+Bundle-SymbolicName: org.jspecify.jspecify
+Bundle-Name: jspecify
```
bjmi, line 2: How about to make it more human readable by `JSpecify annotations`  
Also pom.xml 0.3.0 states: `<name>JSpecify annotations</name>`

chrisrueger: Thanks for the review. Added your suggestion @bjmi

chrisrueger:
Regarding [failing JDK 11 build](https://github.com/jspecify/jspecify/actions/runs/9210685665/job/25338261471?pr=428)

I forgot that bnd 7.0.0 requires JDK 17 as a minimum. 
I'll later decrease bnd to 6.4.0 which should also work and requires JDK 8 asa minimum.

chrisrueger:
> I'll later decrease bnd to 6.4.0 which should also work and requires JDK 8 asa minimum.

@cpovirk could you trigger a rebuild to check if JDK11 build now runs?

cpovirk:
Done. Looks like it's unhappy with the module-info under the multi-release jar root. If that's a significant obstacle, feel free to move it back to the main root (https://github.com/jspecify/jspecify/issues/484).

chrisrueger:
> Done. Looks like it's unhappy with the module-info under the multi-release jar root. If that's a significant obstacle, feel free to move it back to the main root (#484).

Thanks. I have applied a fix. 
it is kind of a dilemma: unfortunatelly bnd in 6.4.0 does not support multi-release jars. It starts being support in [bnd 7.0.0](https://github.com/bndtools/bnd/wiki/Changes-in-7.0.0) but this requires min. JDK 17. 
So the only way is to suppress the warning via -fixupmessage (see https://github.com/bndtools/bnd/issues/3514 and https://bnd.bndtools.org/instructions/fixupmessages.html)

I try to investigate if there are other options like doing different things depending on JDK version.

ben-manes:
You can run gradle on 17 independently from the execution tasks. In the latest version, you can add the gradle-daemon-jvm.properties tool version and use bnd, but build on a lower jdk using the java.toolchain.languageVersion. Does that help?

chrisrueger:
> You can run gradle on 17 independently from the execution tasks. In the latest version, you can add the gradle-daemon-jvm.properties tool version and use bnd, but build on a lower jdk using the java.toolchain.languageVersion. Does that help?

Thanks. Not sure. 

My first thought was to try something like this: https://discuss.gradle.org/t/conditional-dependency-based-on-toolchains-version/46369

```
if (JavaVersion.current().isJava17Compatible()) {
    id "biz.aQute.bnd.builder" version "7.0.0"
  }
else {
    id "biz.aQute.bnd.builder" version "6.4.0"
  }

```

Is this what you mean?

ben-manes:
Nope. Bnd only needs to run in gradle so as long as the daemon is 17+ then everything else will work. You can now force that with a daemon toolchain.

The Java version for tasks like compile, test, exec are from the language toolchain. Instead of inferring it by the daemon jvm, it should be set explicitly. An external env variable can be used, which then lets you cross build. The jdk release flag is also useful as it will ensure that the output matches a desired version as well.

Caffeine’s build [uses](https://github.com/ben-manes/caffeine/blob/master/gradle/plugins/src/main/kotlin/lifecycle/java-library-caffeine-conventions.gradle.kts) this, but it’s a few parts to set up.

cpovirk:
(I'd be happy for someone to make that happen, especially after the fun in https://github.com/jspecify/jspecify/pull/517. I don't expect to get to it myself on any particular timeline, unfortunately.)

chrisrueger:
Thanks for the pointers @ben-manes 
I reverted my changes back to bnd 7.0.0 because this would be the right thing once JDK 17 is used to support Multi-Release jars too.

> In the latest version, you can add the gradle-daemon-jvm.properties tool version and use bnd, but build on a lower jdk using the java.toolchain.languageVersion.

It seems that it is currently in (unreleased) Gradle 8.8.-rc-1 and I first need to get familiar with the whole toolchain stuff. No gradle expert. 
So eventually somebody more familiar with the matter can take over or someday I'll find the time to wrap my head around it.

cpovirk (approved):
We're going to see if we can dig up any other people who know OSGi, but I've found that the community review has made me pretty comfortable with this.

I'm re-approving so that we can see where CI stands now.

cpovirk:
OK, so the failure when building under Java 11 is what's expected from the bnd version, as discussed above. (Sorry, I have half-watched this develop but not retained much information.)

I don't know that anyone will have time to sort out toolchains in the immediate future (though of course we've had opportunities dating back some time now...), so this might well miss the 1.0.0 release. But we can at least try to get back to it sometime afterward, and of course everyone is welcome to continue to rescue us by handing us a tweaked PR that passes while still including testing under the currently tested JDKs.

ben-manes:
Please try rebasing on https://github.com/jspecify/jspecify/pull/558

chrisrueger:
> Please try rebasing on #558

Thanks @ben-manes I rebased on your branch #558 and tried a successful build locally. I pushed now.
@cpovirk could you try rebuilding?

cpovirk (approved):
I'll try to find time to understand the toolchain stuff tomorrow, but it certainly looks good enough to kick off a CI run :)

bnd.bnd
```diff
@@ -0,0 +1,10 @@
+Bundle-SymbolicName: org.jspecify.jspecify
+Bundle-Name: JSpecify annotations
+Bundle-Vendor: jspecify.dev
```
wmdietl, line 3:
```suggestion
Bundle-Vendor: jspecify.org
```

Would it be nicer to use `org` consistently?

bnd.bnd
```diff
@@ -0,0 +1,10 @@
...
+Bundle-Description: An artifact of well-specified annotations to power static analysis checks and JVM language interop.
+Bundle-DocURL: https://jspecify.dev/docs/start-here
```
wmdietl, line 5:
```suggestion
Bundle-DocURL: https://jspecify.org/docs/start-here
```

bnd.bnd
```diff
@@ -0,0 +1,10 @@
+Bundle-SymbolicName: org.jspecify.jspecify
```
wmdietl, line 1: What is the usage of `SymbolicName`? Is it common to use some ID that isn't used for anything else?

bnd.bnd
cpovirk: https://jspecify.org doesn't work :( The redirect is active only at the plain-http http://jspecify.org. If we want https, we need to point straight to the .dev site.

wmdietl: Ah :-(
#267 would fix that.

cpovirk:
Ha, GitHub interpreted "This resolves https://github.com/jspecify/jspecify/pull/428#issuecomment-2225927711" as "This resolves #428" :)

bnd.bnd
cpovirk: https://cwiki.apache.org/confluence/display/SMX4/Creating+a+Bundle+Using+Maven seems to think that it makes sense to match this to the Maven artifact ID, but I think that's the unqualified artifact ID (which would be something like "guava" or "jspecify"), so that strikes me as iffy, like maybe they wanted to show a non-default behavior as a demonstration and they made a suboptimal choice? Guava [doesn't do that](https://github.com/google/guava/blob/4a2d3025250664ff248f5e82373a13f70ce98faa/guava/pom.xml#L73), so it ends up with `Bundle-SymbolicName: com.google.guava`.

At this particular moment, [this Google cache of a Red Hat doc](https://webcache.googleusercontent.com/search?q=cache%3Ahttps%3A%2F%2Fdocs.redhat.com%2Fen%2Fdocumentation%2Fred_hat_jboss_fuse%2F6.0%2Fhtml%2Fdeploying_into_the_container%2Fbestpractices-buildbundles&oq=cache%3Ahttps%3A%2F%2Fdocs.redhat.com%2Fen%2Fdocumentation%2Fred_hat_jboss_fuse%2F6.0%2Fhtml%2Fdeploying_into_the_container%2Fbestpractices-buildbundles&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg60gEIMTAwN2owajeoAgCwAgA&sourceid=chrome&ie=UTF-8) recommends either `${project.artifactId}` or `${project.groupId}.${project.artifactId}`. This PR implements the latter, and I guess I feel better about that, given what I said above?

ben-manes: bnd should auto fill in many of these. Caffeine doesn't specify it explicitly and its populated when I look at the jar's `META-INF/MANIFEST.MF`. I include a PAX-Exam test to validate that the jar works in under OSGi.

<details><summary>Caffeine's bundle</summary>

```ini
Manifest-Version: 1.0
Automatic-Module-Name: com.github.benmanes.caffeine
Build-Jdk-Spec: 11
Bundle-Description: A high performance caching library
Bundle-License: https://www.apache.org/licenses/LICENSE-2.0
Bundle-ManifestVersion: 2
Bundle-Name: com.github.ben-manes.caffeine
Bundle-SymbolicName: com.github.ben-manes.caffeine
Bundle-Version: 3.1.9.SNAPSHOT
Export-Package: com.github.benmanes.caffeine.cache;uses:="com.github.b
 enmanes.caffeine.cache.stats";version="3.1.9",com.github.benmanes.caf
 feine.cache.stats;uses:="com.github.benmanes.caffeine.cache";version=
 "3.1.9",com.github.benmanes.caffeine;version="3.1.9"
Implementation-Title: A high performance caching library
Implementation-Version: 3.1.9-SNAPSHOT
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=11))"
```
</details>

bnd.bnd
chrisrueger: > we should just use "The JSpecify Authors," as in our source files?

I will change it to "The JSpecify Authors"

bnd.bnd
chrisrueger: So I leave it as "Bundle-DocURL: https://jspecify.dev/docs/start-here" for now? (until the .org tld is working in the future)

bnd.bnd
chrisrueger: > bnd should auto fill in many of these

This is true ([see](https://bnd.bndtools.org/heads/bundle_symbolicname.html)). 

**BUT**, since it is a "unique" identifier which other OSGi bundles will use to specify a dependency to - so  it should live forever. I would not rely on this to be auto-generated. 

From [OSGi Docs](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.bsn):

> The Bundle-SymbolicName manifest header is a mandatory header. The bundle symbolic name and bundle version identify a unique bundle.

We had an issue in the past where we had  to rename a bundle folder (e.g. the project folder from `fooA.bar` to `fooB.bar` ... silly example). This changed the bsn too to `fooB.bar` too. This broke all other OSGi  bundles relying on `fooA.bar`. We could fix it by hardcoding the bsn in the `bnd.bnd` file, so it is independent of any folder naming. 

If this will not happen (who knows), then fine, but I personally would recommend to hard code it, so that you do not rely on auto-generation.

chrisrueger: > `${project.groupId}.${project.artifactId}`. This PR implements the latter,

Yes. But I admit I also searched around a bit and look how other bundles are doing it. 
For example: https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core/2.17.2

```
<!-- https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core -->
<dependency>
    <groupId>com.fasterxml.jackson.core</groupId>
    <artifactId>jackson-core</artifactId>
    <version>2.17.2</version>
</dependency>
```

**MANIFEST.MF (excerpt)**:

```
Bundle-License: https://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: com.fasterxml.jackson.core.jackson-core
Implementation-Vendor-Id: com.fasterxml.jackson.core
Specification-Title: Jackson-core
Bundle-DocURL: https://github.com/FasterXML/jackson-core
Bundle-Name: Jackson-core
Multi-Release: true
Build-Jdk-Spec: 1.8
Bundle-Description: Core Jackson processing abstractions (aka Streamin
 g API), implementation for JSON
Implementation-Title: Jackson-core
Implementation-Version: 2.16.1
Bundle-ManifestVersion: 2
Specification-Vendor: FasterXML
Bundle-Vendor: FasterXML
```

bnd.bnd
chrisrueger: Changed it to "The JSpecify Authors" in 87ce0eb

bnd.bnd
cpovirk: Ooo, ouch, yes, the docs appear to be correct that the default is the directory name. I verified that it doesn't even respect [the project name](https://github.com/jspecify/jspecify/blob/f0af7de6f72d01e3a8f3ffc9f528540de770f987/settings.gradle#L21), so it's 100% what you name the local directory on your system.

If there were a way to automatically derive the SymbolicName from the Maven artifact ID or something (as in the Maven plugin?), then I could see setting us up for that. But being explicit seems fine to me.

I wonder how Caffeine's works.... But I don't wonder enough to try to figure it out :)

bnd.bnd
chrisrueger: > https://jspecify.org doesn't work :( The redirect is active only at the plain-http http://jspecify.org. If we want https, we need to point straight to the .dev site.

I leave it as is. It can be changed at any time in the future without causing a problem.

bnd.bnd
```diff
@@ -0,0 +1,10 @@
...
+Bundle-Name: JSpecify annotations
+Bundle-Vendor: The JSpecify Authors
+Bundle-Description: An artifact of well-specified annotations to power static analysis checks and JVM language interop.
+Bundle-DocURL: https://jspecify.dev/docs/start-here
+Bundle-License: https://www.apache.org/licenses/LICENSE-2.0
+-exportcontents: org.jspecify.*
+-reproducible: true
+-noextraheaders: true
+-snapshot: SNAPSHOT
```
cpovirk, line 10: Is this line something that we'd need to change when we publish a proper 1.0.0 release? Or is it used only when the version number is already a snapshot?

(Even if we need to change it for the release, I am happy to have it set in the main branch.)

bnd.bnd
cpovirk: The rest of the `Bundle-` header look like they also override generic defaults or add otherwise absent values, so those look good. And I see that `-noextraheaders` saves us from `Created-By` and `Tool` headers with specific versions in them.

bnd.bnd
cpovirk: Also, extremely minor: I see that this file has to trailing newline. (Vim keeps re-adding one when I mess with it :)) I assume that we can include one without causing any problems?

cpovirk: It looks like, with or without this line, we get:

```
Bundle-Version: 0.0.0.SNAPSHOT
Export-Package: org.jspecify.annotations;version="0.0.0.SNAPSHOT"
```

And, with the version in `gradle.properties` bumped to `1.0.0`, we likewise get the same thing with or without this line:

```
Bundle-Version: 1.0.0
Export-Package: org.jspecify.annotations;version="1.0.0"
```

So maybe we don't need it, but maybe it's not causing any harm, either?

chrisrueger: > Is this line something that we'd need to change when we publish a proper 1.0.0 release? Or is it used only when the version number is already a snapshot?

I think it can stay as is ([see meaning of -snapshot instruction](https://bnd.bndtools.org/instructions/snapshot.html))

The `Bundle-Version` for the OSGi Manifest.mf is taken from `gradle.properties` `version` attribute, which is currently set to version = 0.0.0-SNAPSHOT:

https://github.com/jspecify/jspecify/blob/f0af7de6f72d01e3a8f3ffc9f528540de770f987/gradle.properties#L2

Given the current `bnd.bnd` this results in:

`jspecify-0.0.0-SNAPSHOT.jar`
and
`Bundle-Version: 0.0.0.SNAPSHOT`

Locally I simulated a 1.0.0 release for testing I changed gradle.properties version to:
`version = 1.0.0`

This resulted in:

`jspecify-1.0.0.jar`
and
`Bundle-Version: 1.0.0`

(so SNAPSHOT was removed and proper version was used)

In conclusion: I think this is what you would do for a release too, right?

cpovirk: Thanks, I should have been looking in the bnd docs instead of the OSGi docs... :)

Given that "The default is no substitution," it sounds like this line is a no-op. But that would also mean that it's harmless. Since it's easier to leave it how it is, I plan to do that.

cpovirk:
@wmdietl or anyone else, any other concerns?

bnd.bnd
ben-manes: oh, caffeine works because I completely missed that I hard coded the value like @chrisrueger suggested. It was just in the project specific build file,

```kts
tasks.jar {
  from(sourceSets["main"].output + sourceSets["codeGen"].output)
  dependsOn(compileCodeGenJava)
  bundle.bnd(mapOf(
    "Bundle-SymbolicName" to "com.github.ben-manes.caffeine",
    "Import-Package" to "",
    "Export-Package" to listOf(
      "com.github.benmanes.caffeine",
      "com.github.benmanes.caffeine.cache",
      "com.github.benmanes.caffeine.cache.stats").joinToString(","),
    "Automatic-Module-Name" to "com.github.benmanes.caffeine"))
}
```

bnd.bnd
chrisrueger: > Thanks, I should have been looking in the bnd docs instead of the OSGi docs... :)

No problem. It can be confusing. Thanks for all the pointers.
I will be around for another 2 hours some more changes are required.

build.gradle
```diff
@@ -56,6 +57,24 @@ java {
...
+    }
+    dirPermissions {
+        unix('rwxr-xr-x')
+    }
+}
+
+
+jar {
+    manifest {
+        attributes("Bundle-Version": version)
```
ben-manes, line 74: you might want to also add `Implementation-Version` (https://github.com/ben-manes/caffeine/issues/889)

chrisrueger: Done. Thanks.

wmdietl (approved):
Changes look good to me, but I understand none of the used tools :-)

cpovirk:
You and me both :) Thanks.

(Actually, the discussion here has given me plenty of reason to feel comfortable with merging this. I just don't think anyone will confuse me with someone who has knowledge of OSGi anytime soon....)

chrisrueger:
And thank you and kudos to @ben-manes for all the help 👍

cpovirk:
For the more OSGi-literate, here's what the 1.0.0 data is shaping up to look like:

#### META-INF/MANIFEST.MF
```
Manifest-Version: 1.0
Bundle-Description: An artifact of well-specified annotations to power
  static analysis checks and JVM language interop.
Bundle-DocURL: https://jspecify.dev/docs/start-here
Bundle-License: https://www.apache.org/licenses/LICENSE-2.0
Bundle-ManifestVersion: 2
Bundle-Name: JSpecify annotations
Bundle-SymbolicName: org.jspecify.jspecify
Bundle-Vendor: The JSpecify Authors
Bundle-Version: 1.0.0
Export-Package: org.jspecify.annotations;version="1.0.0"
Implementation-Version: 1.0.0
Import-Package: java.lang,java.lang.annotation
Multi-Release: true
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

#### META-INF/versions/9/OSGI-INF/MANIFEST.MF
```
Manifest-Version: 1.0
Import-Package: java.lang,java.lang.annotation
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

ben-manes:
haha, I think @chrisrueger is the only OSGi literate one here. If you want to validate then you could write a unit test ([example](https://github.com/ben-manes/caffeine/blob/56a705509d8024ab006b8689ab9086a3f19a3642/caffeine/src/test/java/com/github/benmanes/caffeine/OSGiTest.java#L37-L52)) to have some assurance that things were okay.

chrisrueger:
Thanks @cpovirk for the update.
I also re-tested locally (with the 0.0.0.SNAPSHOT) version how it looks, when I used the JSpecify OSGi bundle in our application. 
Summary: It works fine, from my OSGi user perspective. Also the Eclipse PDE / JDT people should be able to work with it I guess (@bjmi ?).

For documentation purposes here are some screenshots of how the [bndtools Eclipse Plugin](https://bndtools.org/) shows the bundle when I import it as a dependency. This time I added it to a local file repo, but 1.0.0 will e.g. come from Maven Central then. 

image

image

I also started the bundle for the sake of testing with our app and verified with the OSGi console that the bundle is successfully started and in an Active state:

```
lb jspec
START LEVEL 1
   ID|State      |Level|Name
  162|Active     |    1|JSpecify annotations (0.0.0.SNAPSHOT)|0.0.0.SNAPSHOT
```


For sake of completeness the pretty-printed output of the bndtools' JARViewer:

```
[MANIFEST]

Bundle-Description                      An artifact of well-specified annotations to power static analysis checks and JVM language interop.
Bundle-DocURL                           https://jspecify.dev/docs/start-here
Bundle-License                          https://www.apache.org/licenses/LICENSE-2.0
Bundle-ManifestVersion                  2
Bundle-Name                             JSpecify annotations
Bundle-SymbolicName                     org.jspecify.jspecify
Bundle-Vendor                           The JSpecify Authors
Bundle-Version                          0.0.0.SNAPSHOT
Export-Package                          org.jspecify.annotations;version="0.0.0.SNAPSHOT"
Implementation-Version                  0.0.0-SNAPSHOT
Import-Package                          java.lang
                                        java.lang.annotation
Manifest-Version                        1.0
Multi-Release                           true
Require-Capability                      osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"


[IMPEXP]

Import-Package
  java.lang                              
  java.lang.annotation                   

Export-Package
  org.jspecify.annotations               {version=0.0.0.SNAPSHOT}

[CAPABILITIES]

Require-Capability
  osgi.ee                                {filter:=(&(osgi.ee=JavaSE)(version=1.8))}

[COMPONENTS]



[METATYPE]



[API USES]


[USES]

org.jspecify.annotations                


[USEDBY]



[LIST]

META-INF
  MANIFEST.MF
META-INF/versions
META-INF/versions <no contents>
META-INF/versions/9
  module-info.class
META-INF/versions/9/OSGI-INF
  MANIFEST.MF
org
org <no contents>
org/jspecify
org/jspecify <no contents>
org/jspecify/annotations
  NonNull.class
  NullMarked.class
  NullUnmarked.class
  Nullable.class

```

cpovirk:
Thanks. Real-world usage is more reassuring than a test, but a test also adds reassurance. I cobbled together something based on Caffeine and ancient Google Groups postings in https://github.com/jspecify/jspecify/commit/3dd37003c734cb4dd6fcef0e8530439edeb1e280. It passes, and if I revert the commit from this PR, then it fails with `java.lang.ClassNotFoundException: org.jspecify.annotations.Nullable not found by PAXEXAM-PROBE-4d05fdef-9dab-45b4-a779-2808c46ee9f8 [16]`. I'm not sure if I'll ever clean it up for submission or not.

bjmi:
First of all, I would like to thank you for your efforts that have led to a solution.

Now for two irregularities that I noticed.

If both _MANIFEST.MF_ files don't differ in `Import-Package` and `Require-Capability` then why not omit _META-INF/versions/9/OSGI-INF/MANIFEST.MF_?
Or alternatively specify `Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9))"` for the latter.
**META-INF/MANIFEST.MF**
```
Import-Package: java.lang,java.lang.annotation
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

**META-INF/versions/9/OSGI-INF/MANIFEST.MF**
```
Import-Package: java.lang,java.lang.annotation
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

I'm not convince of BSN `org.jspecify.jspecify`. The second _jspecify_ is superfluous. Since Maven uses two parts to identify an artifact by _groupId_ and _artifactId_ OSGI uses a single identifier the BSN only.
E.g. _com.acme:acme-api_ -> _com.acme.api_ or _org.foo:foo_ -> _org.foo_ as decribed in [Apache Felix Maven Bundle Plugin (BND) doc](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_default_behavior)
What's next with jspecify? Further annotations in the current bundle then _org.jspecify.annotations_ is a good candidate. Or something like an uber jar for all kinds of things then `org.jspecify` is fine.

chrisrueger:
> I'm not convince of BSN org.jspecify.jspecify. The second jspecify is superfluous.

Feeling the same a bit after some distance now.

> What's next with jspecify? Further annotations in the current bundle then _org.jspecify.annotations_ is a good candidate. Or something like an uber jar for all kinds of things then `org.jspecify` is fine.

I like `org.jspecify.annotations`

It transports the intention of the bundle in its current state - which is about "annotations".  
I created PR #569 for this. @cpovirk 

---

Regarding the other thing, the redundancy of the `META-INF/versions/9/OSGI-INF/MANIFEST.MF`

This was auto-generated by bnd and I am not familiar with [Multi-Release-Jars (MRJ)](https://docs.oracle.com/en/java/javase/11/docs/specs/jar/jar.html#multi-release-jar-files) / JPMS. 

Only thing I can point out regarding the version range:
- `(version=1.8)` means `version >= 1.8` ([see](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3189032) ) which is sometimes confusing, since it "looks" like a fixed version, but it is actually a range.
- since MRJ were introduced with JDK 9 I would argue that `version >= 1.8` is not wrong. Only a JDK >= 9 would inspect the `META-INF/versions/` folder

PR #437, opened 2023-12-19
Change to install `java-library` and `distribution` plugins instead of `java-library-distribution` so that there's only one distribution.

### LINKED ISSUE #436: Conformance test zip is empy after `assemble`.
Reported by @wmdietl.

This is because there's a `distZip` task that also writes the same ZIP file, but from a different source.

netdpb:
Fixes #436.


wmdietl (approved):
Thanks, it worked for me locally.

PR #438, opened 2023-12-19
Put the samples files into the conformance tests zip.

netdpb:
This helps to decouple the builds. To be used in https://github.com/jspecify/jspecify-reference-checker/pull/123.

### DIFF
--- a/conformance-tests/build.gradle
+++ b/conformance-tests/build.gradle
@@ -40,6 +40,9 @@ distributions {
                 from depsJar
                 from configurations.assertionsRuntimeClasspath
             }
+            into('/samples') {
+                from '../samples'
+            }
         }
     }
 }


conformance-tests/build.gradle
```diff
@@ -40,6 +40,9 @@ distributions {
                 from depsJar
                 from configurations.assertionsRuntimeClasspath
             }
+            into('/samples') {
+                from '../samples'
```
wmdietl, line 44: Should this also be put into the `sourceSets` to have everything defined there?

wmdietl (approved):
LGTM

conformance-tests/build.gradle
netdpb: I don't want to move the samples at this time, since others might need them to be where they are.

And I don't want to compile them, since some of them may have compilation errors. For instance, many of them depend on `@NullUnspecified`, which doesn't actually exist.

Eventually, I'd like to see that directory deleted, once all the use cases are expressed in the real conformance tests.

PR #441, opened 2024-01-01
Update to gradle 8.5

wmdietl:


gradle/publish.gradle
```diff
@@ -41,9 +41,11 @@ publishing {
 
 nexusPublishing {
     repositories {
+        // TODO: the following line causes the `The Project.getConvention() method has been deprecated.` warning.
         sonatype {
-            // //only for users registered in Sonatype after 24 Feb 2021
+            // only for users registered in Sonatype after 24 Feb 2021
             nexusUrl = uri("https://s01.oss.sonatype.org/service/local/")
+            snapshotRepositoryUrl = uri("https://s01.oss.sonatype.org/content/repositories/snapshots/")
```
wmdietl, line 48: Not sure if we need/want this. It's from the corresponding block in the readme: https://github.com/gradle-nexus/publish-plugin#publishing-to-maven-central-via-sonatype-ossrh

gradle/publish.gradle
```diff
@@ -41,9 +41,11 @@ publishing {
 
 nexusPublishing {
     repositories {
+        // TODO: the following line causes the `The Project.getConvention() method has been deprecated.` warning.
```
wmdietl, line 44: The `gradle-nexus` plugin still is on [gradle 8.2.1](https://github.com/gradle-nexus/publish-plugin/blob/c2614e3a5fd61008c66633ca061760df8d4a5106/gradle/wrapper/gradle-wrapper.properties#L4).
I tracked down the deprecation warning to the `sonatype` block, but didn't find a way to avoid the warning.

gradle/publish.gradle
netdpb: We will if we want to publish snapshot releases (which I think we will want to, at least for the conformance tests). I'm doing the same thing in my work to decouple the builds.

gradle/publish.gradle
netdpb: Let's note that in the comment so we're not tempted to track it down again. It's not really a TODO until they fix the plugin.

netdpb (approved):

gradle/publish.gradle
wmdietl: Undid these changes to avoid conflicts with #446.

wmdietl (commented):
I've expanded the comment and reduced conflicts with what #446 will do.

netdpb (approved):

PR #449, opened 2024-01-05
Publish snapshots

netdpb:


netdpb:
I've created secrets based on my own Sonatype user token (not my real credentials).

cpovirk (approved):

PR #462, opened 2024-02-03
Configure dependabot behavior

wmdietl:
Dependabot by default only seems to update the npm packages: https://github.com/jspecify/jspecify/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+

Adding this configuration file hopefully updates also the gradle and github actions versions.


cpovirk (approved):
(Maybe it's been doing only the packages with security vulnerabilities? Regardless of exactly what it's doing now, +1 for doing more :))

PR #471, opened 2024-02-05
A JSpecify taglet for use in conformance tests

wmdietl:
This would allow us to replace [this](https://github.com/jspecify/jspecify/pull/461/files#diff-54afd55c9faa01261053afb345cada260320a3feb3fcf02e39fb35479e1a03edR25-R27):

````
 * @see <a
 *     href="https://jspecify.dev/docs/spec#recognized-locations-for-type-use-annotations">Recognized
 *     locations for type-use annotations</a> in the JSpecify specification.
````

with something like

````
 * @jspecify_spec_nullness #recognized-locations-for-type-use-annotations Recognized locations for type-use annotations
````

I just tried the taglet locally, as we currently don't produce javadoc for the conformance tests.
I've not developed a taglet before. Before investing more time, let me know whether this would be useful.


netdpb:
I'm not sure it's worth the time now. When we update the specification, its form and structure might change. (For instance, we might number the sections, or it might be split among pages, or its URL could change.) That would mean we'd have to adjust not only the Javadoc in the tests, but this taglet as well.

Second, people will be looking at the source of the tests, not just the generated Javadoc. It's nice to have the URL right there; many IDEs or other UIs (not GitHub apparently) make those clickable.

Third, we may also link to other documents, such as the JLS.

Fourth, we might want to link from inside symbols, where Javadoc isn't valid.

If and when we adopt a more formal structure for the spec, I could maybe see something like this being more valuable.

wmdietl:
I'm not going to argue about this too much, just some responses.

1. For some changes, e.g. finally fixing #267, we would only need to change the taglet and not change the conformance test cases at all. For more structural changes, we should make sure that the taglet can also handle them.
2. Fair.
3. OpenJDK actually uses a taglet for references to the JLS, both [inline ](https://github.com/openjdk/jdk/blob/50b17d9846f7727a5f7225e1b093b6bdff909478/src/java.compiler/share/classes/javax/annotation/processing/Processor.java#L287) and as a [block ](https://github.com/openjdk/jdk/blob/50b17d9846f7727a5f7225e1b093b6bdff909478/src/java.compiler/share/classes/javax/annotation/processing/Processor.java#L300) taglet. We could look into whether we can use that taglet for JLS references.
4. Sure, but hopefully the fast majority can be handled by a taglet. All of the references in #461 were in javadoc.

netdpb:
I'm still not convinced this is worth doing now. I'd rather not put anything else in the way of writing more tests.

netdpb:
Closing for now. We can always revisit.

PR #473, opened 2024-02-06
Add comment to sample files covered by conformance test assertions

netdpb:
Also remove comment from conformance test assertions that documents which sample files are covered.


wmdietl (approved):
I like this direction of comments.

PR #474, opened 2024-02-06
Use upload-artifact@v4 to match download-artifact@v4.

netdpb:
Fixes a docs build break ([example](https://github.com/jspecify/jspecify/actions/runs/7805130551/job/21288658577)).


cpovirk (approved):
Heh.

netdpb:
Dependabot is single-minded.

PR #478, opened 2024-02-07
Use the recommended gradle/actions/setup-gradle@v3

wmdietl:
#476 suggests doing this update instead of merging #476.


.github/workflows/build.yml
```diff
@@ -76,15 +76,15 @@ jobs:
       - name: Run Conformance Tests
         if: github.base_ref == 'main' # only PRs onto main
         continue-on-error: true
-        uses: gradle/gradle-build-action@v2
+        uses: gradle/actions/setup-gradle@v3
```
netdpb, line 79: I think the recommended pattern is to have a single, separate step to set up Gradle, and then to just run Gradle as a normal shell invocation.

See the `publish-snapshot` job above.

cpovirk:
> https://github.com/jspecify/jspecify/pull/476 suggests doing this update instead of merging https://github.com/jspecify/jspecify/pull/476.

Oops, sorry: I was merging old PRs (since those contain versions of files without licenses, which upsets our licensing tooling), and I merged that one before noticing this. Obviously we could still do this, but you'll now have a merge conflict :\

wmdietl:
Closing this as obsolete.

PR #485, opened 2024-02-26
Change from `? super Object` to `? super Number` to avoid errors

wmdietl:
In the samples errors I see 14 occurrences (at least in https://github.com/jspecify/jspecify-reference-checker/pull/165) of this kind of error:

````
      /home/runner/work/jspecify-reference-checker/jspecify-reference-checker/jspecify-reference-checker/build/conformanceTests/samples/AnnotatedWildcard.java:31: (type.invalid.super.wildcard) bounds must have the same annotations.
          Lib<@Nullable ? super Object> x3,
              ^
      super bound  : Object?
      extends bound: Object
````

That is a correct error, but not the point of these test cases.
This changes most occurrences to `? super Number` and adds expected errors otherwise.

### DIFF
--- a/samples/ContainmentSuperVsExtendsSameType.java
+++ b/samples/ContainmentSuperVsExtendsSameType.java
@@ -20,7 +20,7 @@ import org.jspecify.annotations.Nullable;
 class ContainmentSuperVsExtendsSameType {
   void x() {
     // :: error: jspecify_nullness_mismatch jspecify_but_expect_nothing
-    new Check<Lib<? extends Object>, Lib<? super Object>>();
+    new Check<Lib<? extends Number>, Lib<? super Number>>();
   }
 
   interface Lib<T extends @Nullable Object> {}
--- a/samples/SuperObject.java
+++ b/samples/SuperObject.java
@@ -20,6 +20,11 @@ import org.jspecify.annotations.NullnessUnspecified;
 @NullMarked
 class SuperObject {
   void foo(
+      // Wildcards with `? super Object` must have the same upper and lower bound. Here, the upper
+      // bound is @Nullable, whereas the lower bound is @NonNull.
+      // As javac doesn't allow us to keep track of the separate annotations, there is an error
+      // here.
+      // :: error: jspecify_conflicting_annotations
       Lib<? super Object> lib,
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: samples/SuperObject.java [hunk 1]
```diff
@@ -20,6 +20,11 @@
 @NullMarked
 class SuperObject {
   void foo(
+      // Wildcards with `? super Object` must have the same upper and lower bound. Here, the upper
+      // bound is @Nullable, whereas the lower bound is @NonNull.
+      // As javac doesn't allow us to keep track of the separate annotations, there is an error
+      // here.
+      // :: error: jspecify_conflicting_annotations
```
wmdietl, line 27: I need to add a mapping to `(type.invalid.super.wildcard) bounds must have the same annotations.` in a separate PR against the reference checker.

wmdietl: https://github.com/jspecify/jspecify-reference-checker/pull/166

samples/ContainmentSuperVsExtendsSameType.java
```diff
@@ -20,7 +20,7 @@
 class ContainmentSuperVsExtendsSameType {
   void x() {
     // :: error: jspecify_nullness_mismatch jspecify_but_expect_nothing
-    new Check<Lib<? extends Object>, Lib<? super Object>>();
+    new Check<Lib<? extends Number>, Lib<? super Number>>();
```
cpovirk, line 23: I had typed a comment here, but I don't know what happened to it....

My comment was that my guess was that https://github.com/jspecify/jspecify/commit/6f78b15de1bcf1dd7866a7663b88875cc756faac was describing a problem specific to `Object`. If so, then it would be nice to use `jspecify_conflicting_annotations` here instead of changing the type. (Or we could have that _plus_ this :))

It occurs to me now, though, that if the tests don't fail, then I'm wrong in claiming that the problem is specific to `Object` :)

wmdietl: There are still test cases for `? super Object` - the point here I think is something else.
There is an `jspecify_but_expect_nothing` here, so I will look later why no error is raised.

PR #486, opened 2024-02-26
Remove `jspecify_but_expect_nothing` b/c EISOP produces an error

wmdietl:
EISOP produces an expected error:

````
      /home/runner/work/jspecify-reference-checker/jspecify-reference-checker/jspecify-reference-checker/build/conformanceTests/samples/CaptureConvertedTypeVariableBounded.java:102: (return.type.incompatible) incompatible types in return.
          return p.get();
                      ^
      type of expression: capture#454 of ? extends T
      method return type: Object
````

### DIFF
--- a/samples/CaptureConvertedTypeVariableBounded.java
+++ b/samples/CaptureConvertedTypeVariableBounded.java
@@ -98,7 +98,7 @@ class CaptureConvertedTypeVariableBounded {
     }
 
     Object x1(Inner<? extends @NullnessUnspecified Object> p) {
-      // :: error: jspecify_nullness_not_enough_information jspecify_but_expect_nothing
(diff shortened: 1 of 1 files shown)


cpovirk (approved):

wmdietl:
There are currently 54 occurrences of `// :: error: jspecify_nullness_not_enough_information jspecify_but_expect_nothing` in the `samples-google-prototype-eisop` branch.
Should I remove the `jspecify_but_expect_nothing` from all of them?
The `jspecify_nullness_not_enough_information` should already express that having an error or not is okay, right?
Or what does the combination of the two express?

wmdietl:
Thanks, that makes sense! I'll remove the `but_expect_nothing` in batches when I verified them.

PR #487, opened 2024-02-26
Capture of `Object*` is not compatible with `Object*`.

wmdietl:
https://github.com/jspecify/jspecify-reference-checker/pull/165#discussion_r1503356503

Using a wildcard that captures `Object*` is not compatible with an `Object*` bound.


cpovirk:
As discussed in https://github.com/jspecify/jspecify-reference-checker/pull/165#discussion_r1504549615, I'm hoping that we can justify avoiding an error in this case.

If you want to get the tests to pass with the current behavior, then I suggest `jspecify_but_expect_warning` instead (to indicate that the checker currently produces an error in strict mode but (_if_ I'm right) that it shouldn't).

PR #489, opened 2024-03-04
A brief note on the non-hierarchicalness of packages

msridhar:
Just a first shot at this to get the fact somewhere in the user guide, happy to make edits


xenoterracide:
works for me.

cpovirk (approved):
Thanks.

I don't feel a deep sense of ownership over this particular doc, so if anyone else does, please do jump in with any post-merge comments. (@msridhar , I'll leave it to you to press the merge button, but go for it anytime.)

(I do see that https://jspecify.dev/docs/api/org/jspecify/annotations/NullMarked.html already mentions "This annotation has no effect on 'subpackages,'" so hooray for that!)

PR #498, opened 2024-03-22
Compile the `samples` files as tests

wmdietl:
This is into the `samples-google-prototype-eisop` branch. Let me know if I should do something similar for `main` and/or `samples-google-prototype`.

### DIFF
--- a/samples/ContainmentSuperVsExtendsSameType.java
+++ b/samples/ContainmentSuperVsExtendsSameType.java
@@ -19,8 +19,9 @@ import org.jspecify.annotations.Nullable;
 @NullMarked
 class ContainmentSuperVsExtendsSameType {
   void x() {
-    // :: error: jspecify_nullness_mismatch jspecify_but_expect_nothing
-    new Check<Lib<? extends Number>, Lib<? super Number>>();
+    // `conflicting_annotations` b/c upper and lower bound of `? super Object` differ.
+    // :: error: jspecify_nullness_mismatch jspecify_but_expect_nothing :: error: jspecify_conflicting_annotations
+    new Check<Lib<? extends Object>, Lib<? super Object>>();
   }
 
   interface Lib<T extends @Nullable Object> {}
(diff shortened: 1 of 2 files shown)


samples/ContainmentSuperVsExtendsSameType.java
```diff
@@ -19,8 +19,10 @@
 @NullMarked
 class ContainmentSuperVsExtendsSameType {
   void x() {
+    // `conflicting_annotations` b/c upper and lower bound of `? super Object` differ.
     // :: error: jspecify_nullness_mismatch jspecify_but_expect_nothing
```
wmdietl, line 23: Separately, as we go through `expect_nothing` and `but_expect_error` messages, it would be nice to have a comment why a mismatch is expected.
Here, I expect the problem is that `Lib<? super Object>` is not a subtype of `Lib<? extends Object>`, right?

samples/ContainmentSuperVsExtendsSameType.java
```diff
@@ -19,8 +19,10 @@
...
-    new Check<Lib<? extends Number>, Lib<? super Number>>();
+    // :: error: jspecify_conflicting_annotations
+    new Check<Lib<? extends Object>, Lib<? super Object>>();
```
wmdietl, line 25: #485 previously changed this, but that change introduced a java compilation failure that went unnoticed.

samples/ContainmentSuperVsExtendsSameType.java
cpovirk: Yes, I believe that's right.

And yes, much of the idea of the conformance-test framework is that it lets us replace "error of some kind on this line" with more of an "expected Foo but was Foo?" style.

cpovirk (approved):
> This is into the `samples-google-prototype-eisop` branch. Let me know if I should do something similar for `main` and/or `samples-google-prototype`.

I could see doing it for `main` (and then merging that into `samples-google-prototype`). Maybe it will help with the conformance-test scores there?

wmdietl:
> I could see doing it for `main` (and then merging that into `samples-google-prototype`). Maybe it will help with the conformance-test scores there?

Doing it on `main` fails, because `UnknownNullness` isn't declared in that branch. Is it worth going through the samples and removing the failing tests? As we're rewriting the samples as conformance tests, I think that's not worth it.

I've created #501 as additional PR, to also compile the samples in the sister-branch `samples-google-prototype`.

cpovirk:
Heh, oops. As soon as I saw that this thread had a new message on my phone, I suddenly realized my mistake about `main` :) I agree that it's not worth doing anything there.

PR #501, opened 2024-03-26
Compile the `samples` files as tests (samples-google-prototype)

wmdietl:
Version of #498 for the `samples-google-prototype` branch.


cpovirk (approved):
Thanks.

Issue #248, opened 2022-06-17
Substituting a specified-nullness type argument and (a) an unspecified-bound type variable or (b) an unspecified-nullness usage of a type variable


kevinb9n:
"Discussion" issue: we'll need to fork off independently resolvable questions as necessary.

Here's a basic unannotated generic class:

```
@NullUnmarked
static class UnmarkedClass<T> {
  T get() { ... }
  void set(T t) { ... }
}
```

First some general background -- principles I think we apply in how we think about this case:

* There is some correct way that `UnmarkedClass` could be annotated, and we want to plan for that future.
* To the extent feasible, we'd like callers to write the same code _now_ that they will have to write in that future. (It's better if correctly annotating the class doesn't break them!) This may involve some amount of guessing at *how* exactly `UnmarkedClass` will get annotated.
* Our guiding star has been the "many-worlds" model: consider each possible way `UnmarkedClass` might get annotated like it's a parallel universe; report a finding only if that finding exists in all universes. (If it's entirely valid in *some* universe, there is a good chance that's the universe we're in, so we don't have enough basis to complain.)
* A problem with this generic class is that due to type variable projection there are 13 of these universes. I suspect it's fair to say that we probably can't future-proof against all 13 of them at once very easily. So we have to have some tolerance for the idea that sometimes -- maybe, who knows, 20% of the time? -- it won't work great and you'll be compelled to either use ugly workarounds or just go provide the stub file already.
* We won't resort to the "just go provide the stub file already" argument very often, or it would undo the benefit of having the ternary design in the first place.

Now, here's a conformance test, written using an imagined form of @netdpb's [semi-magical proposed framework](https://docs.google.com/document/d/1U7DHHFz49nrByyUEhLSVUYrxblfo2dLtfO6XETcYBGs/edit?resourcekey=0-ByeD2fAlv3JqErgUissoFQ#heading=h.3eq51y2skfxb), with some pieces missing:

```
@NullUnmarked
static class UnmarkedClass<T> { // just repeating this from up above
  T get() { ... }
  void set(T t) { ... }
}

@NullMarked
void testUnmarkedClassOfNonNullFoo(UnmarkedClass<Foo> ofNonNullFoo) {
  assertThat(ofNonNullFoo).hasType("UnmarkedClass!<Foo!>");
  assertThat(ofNonNullFoo.get()).hasType("__?1?__");
  assertThat(ofNonNullFoo).set(accepts("__?2?__"));
}
```

Observations:

[...]

PR #505, opened 2024-04-11
Updates to better match `main-eisop` reference checker

wmdietl:


samples/ComplexParametric.java
```diff
@@ -43,7 +43,7 @@ interface SuperSuper<T extends @Nullable Object> {
 
     void checkNeverNull(Lib<? extends Object> lib);
 
-    <U> void checkUnspecNull(Lib<@NullnessUnspecified U> lib);
+    <U extends @Nullable Object> void checkUnspecNull(Lib<@NullnessUnspecified U> lib);
```
wmdietl, line 46: Without the more flexible bound, there are warnings about invalid inferred method type arguments.
As the bound of `U` doesn't really matter, I think this doesn't change anything substantially.
Let me know if this isn't correct.

samples/nullnessUnspecifiedTypeParameter/nullnessunspecifiedtypeparameter/NullnessUnspecifiedTypeParameter.java
```diff
@@ -31,8 +31,8 @@ class Test {}
 class Instances {
   static final NullnessUnspecifiedTypeParameter<Object> A1 =
       new NullnessUnspecifiedTypeParameter<>();
-  // :: error: jspecify_nullness_mismatch
   static final NullnessUnspecifiedTypeParameter<@Nullable Object> A2 =
+      // :: error: jspecify_nullness_mismatch
```
wmdietl, line 35: In the really old version of the CF `main` uses, the error was apparently on the `=`.
For a long while, it is now on the right-hand side of the assignment, which is here on a different line.
As these are tests for a particular tool, I think it's fine to move the expected error.
For the conformance tests, we should look into ensuring such tests are always on one line.

wmdietl (commented):
@cpovirk These are some changes I noticed when going through mismatches when using the `main-eisop` reference checker.

samples/ComplexParametric.java
cpovirk: I think that https://github.com/jspecify/jspecify/blob/8989d39851ae59b92db141b051ebdea37fbe4845/samples/ComplexParametric.java#L21-L25 is enough reason to worry too much about this file in general :)

samples/MultiBoundTypeVariableUnspecToObject.java
```diff
@@ -51,17 +51,17 @@ <T extends Object & Lib> Object x0(@NullnessUnspecified T x) {
...
 
   <T extends @Nullable Object & @NullnessUnspecified Lib> Object x7(@NullnessUnspecified T x) {
-    // :: error: jspecify_nullness_not_enough_information jspecify_but_expect_error
+    // :: error: jspecify_nullness_not_enough_information
     return x;
   }
 
   <T extends @Nullable Object & @Nullable Lib> Object x8(@NullnessUnspecified T x) {
-    // :: error: jspecify_nullness_mismatch
+    // :: error: jspecify_nullness_not_enough_information
```
cpovirk, line 64:
```suggestion
    // :: error: jspecify_nullness_mismatch jspecify_but_expect_warning
```

Under our current substitution rules, I think that `jspecify_nullness_mismatch` is correct: `@NullnessUnspecified T` can currently "add" nullness to the given type but not "remove" nullness from it. So, if `T` is a nullable type (as it can be, thanks to the bounds), `@NullnessUnspecified T` can be nullable, too.

This may change when we look at https://github.com/jspecify/jspecify/issues/248 again, but for now, it's what our rules say, and it's what other samples are set up for.

samples/nullnessUnspecifiedTypeParameter/nullnessunspecifiedtypeparameter/NullnessUnspecifiedTypeParameter.java
cpovirk: Moving it SGTM.

Putting everything on one line in the future sounds good if we can get away with it.

cpovirk (approved):
Nice to see a bunch of spurious errors melt away :)

samples/MultiBoundTypeVariableUnspecToObject.java
cpovirk: (Err, I was assuming above that the checker was currently producing a warning—hence "jspecify_but_expect_warning." But my main point is about why I think "jspecify_nullness_mismatch" is correct. Then we can fill in an "override" like "jspecify_but_expect_warning" or "jspecify_but_expect_nothing" if the checker currently has one of those behaviors and we want to encode it here.)

wmdietl: I've undone this change.

PR #517, opened 2024-05-24
Fix a couple problems with building under JDK 22:

cpovirk:
- Upgrade Spotless to upgrade google-java-format.
- Make nested test classes `static` so that we don't end up with
  [synthetic parameters that upset our JDK-8 integration
  test](https://bugs.openjdk.org/browse/JDK-8058322).
  (https://errorprone.info/bugpattern/ClassCanBeStatic!)

Thanks to cushon@ for the pointers.

While there, rename a test that I should have renamed as part of
https://github.com/jspecify/jspecify/pull/270.

And update our CI versions to include JDK 21. (I haven't added JDK 22,
as we've seen some potential signs of remaining flakiness there.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java_version: [11, 17, 18, 19]
+        java_version: [11, 17, 21]
         experimental: [false]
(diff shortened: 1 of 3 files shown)


cpovirk:
Thanks, that seems to be the consensus advice, which _someday_ I will get around to understanding and implementing in our various projects :)

netdpb (approved):

Issue #491, opened 2024-03-08
Take the Javadoc warning down a notch


kevinb9n:
Currently we have

> **Warning:** These annotations are under development, and **any** aspect of their naming, locations, or design is subject to change until the JSpecify 1.0 release. Moreover, supporting analysis tools will be tracking the changes on varying schedules. Releasing a library using these annotations in its API is **strongly discouraged** at this time.

This seems overstated at this point, yes?

kevinherron:
I was just about to ask about this. On the repo home page / README you have:
> [Version 0.3](https://github.com/jspecify/jspecify/releases/tag/v0.3.0) is relatively safe to depend on in your code. Or you can read [a more detailed answer](https://github.com/jspecify/jspecify/wiki/adoption).

But then I read the Javadoc and started to question whether it's time yet.

Is there somewhere I can read about the remaining obstacles to a 1.0 version?

kevinb9n:
Glad you're interested! You can filter issues by the 1.0 milestone. That should at least trend toward accurate.

However, there is an ambiguity between the _jar_ going 1.0 and the _spec_ going 1.0; there are several minor changes that can happen to the latter without any effect on the former. I'm not sure if we accounted for that in how we apply the milestone to issues.

kevinb9n:
For this issue I set the milestone to 1.0 but it's actually "the sooner the better", doesn't need to wait.

kevinherron:
> Honestly, what happens if we just remove this warning completely?

More users? 🙃

agentgt:
That warning btw I have seen cited on reddit and various places and the general outcome seems to be unfavorable inhibition even for library authors like myself.

Anyway I'm for more users 😄

kevinb9n:
(we need to regen the hosted javadoc)

PR #526, opened 2024-06-14
Remove now-empty "Important warning" heading.

cpovirk:
Its contents were removed in
https://github.com/jspecify/jspecify/pull/511

PR #527, opened 2024-06-14
Regenerate Javadoc to remove warnings.

cpovirk:
This picks up the changes of
https://github.com/jspecify/jspecify/pull/511 and the pending changes of
https://github.com/jspecify/jspecify/pull/526.


netdpb (approved):

cpovirk:
For reference, the command I ran was:

```shell
git rm -r docs/static/docs && rm -rf docs/static/docs/api && mkdir -p docs/static/docs && $HOME/jdk-23-ea+22/bin/javadoc -notimestamp -windowtitle 'jspecify 0.0.0-SNAPSHOT API' src/main/java/org/jspecify/annotations/*.java -d docs/static/docs/api && git add docs/static/docs
```

And the command that I probably _should_ have run was:

```shell
JAVA_HOME=$HOME/jdk-23-ea+22 ./gradlew clean javadoc && git rm -r docs/static/docs && rm -rf docs/static/docs/api && mkdir -p docs/static/docs && cp -a build/docs/javadoc docs/static/docs/api && git add docs/static/docs
```

But I have had sporadic trouble with JDK 23 and Gradle, as previously discussed somewhere or other.... (Compare https://github.com/jspecify/jspecify/pull/517 for JDK 22.) Hopefully we could use any version of the JDK to run _Gradle_ but still use JDK 23 to run _Javadoc_, presumably using "toolchains," as also discussed at various times (e.g., in https://github.com/jspecify/jspecify/issues/271). Or maybe we don't need to use the latest version of Javadoc.

Anyway, the two commands produce identical output except for the capitalization of "UTF-8" in the output's `meta` `Content-Type` value.

Presumably @netdpb will figure something out as part of https://github.com/jspecify/jspecify/pull/528.

netdpb:
Fixes #491.

PR #528, opened 2024-06-14
Stop checking in generated files for GitHub Pages

### LINKED ISSUE #519: Stop checking in generated files for the website
Instead of committing generated files (like HTML output from the Javadoc tool) into the [/docs](https://github.com/jspecify/jspecify/tree/main/docs) directory or into the [gh-pages](https://github.com/jspecify/jspecify/tree/gh-pages) branch, we should use a GitHub Actions workflow to generate those as part of the GitHub Pages build process, as described [here](https://docs.github.com/en/enterprise-cloud@latest/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site).

netdpb:
Fixes #519.


netdpb:
This is only a start. I'm basing this on [this documentation](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) and [this starter workflow](https://github.com/actions/starter-workflows/blob/main/pages/jekyll.yml).

I need to verify that this works and incorporate the generated Javadoc into this workflow before this is a real PR.

build.gradle
```diff
@@ -54,6 +55,25 @@ java {
 
 javadoc {
     options.encoding = 'UTF-8'
+    options.windowTitle = 'jspecify 0.0.0-SNAPSHOT API'
```
cpovirk, line 58: As best I can tell, this is unnecessary: Gradle fills it in. (I did have to pass it manually when I was running javac directly.)

Now, it sounds like today's Javadoc defaults might not require the `encoding` line above, either. But at least that one won't require a single-bit edit as part of the 1.0.0 release :)

cpovirk (approved):
Thanks!

> If you look at the Docs action, you can download the zipped artifact that would get pushed in the next job.

A zip with a tar inside it, apparently. Sure, GitHub, whatever :)

PR #540, opened 2024-07-02
Update the start-here page

### LINKED ISSUE #524: Document the package name in the User Guide and perhaps landing page
...perhaps alongside [the Maven coordinates](https://github.com/jspecify/jspecify/issues/503).

netdpb:
Note the actual package name (fixes #524).
Note that 1.0 is upcoming, not eventual.
Fix the link to the specification.

### DIFF
--- a/docs/docs/start-here.md
+++ b/docs/docs/start-here.md
@@ -105,6 +105,7 @@ to extend our support beyond nullness.
 [issues]: https://github.com/jspecify/jspecify/issues
 [release]: https://search.maven.org/artifact/org.jspecify/jspecify/0.3.0/jar
 [spec]: /docs/spec
+[specification]: /docs/spec
 [stackoverflow answer]: https://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use
 [user guide]: /docs/user-guide
 [wiki]: https://github.com/jspecify/jspecify/wiki
--- a/docs/docs/start-here.md
+++ b/docs/docs/start-here.md
@@ -43,13 +43,13 @@ questions, which you can look for in the [Nullness Design FAQ]. If you like,
 ### Begin with one of
 
 *   The [User Guide].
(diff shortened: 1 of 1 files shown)


docs/docs/start-here.md
```diff
@@ -38,53 +42,58 @@ questions, which you can look for in the [Nullness Design FAQ]. If you like,
 
 ### Begin with one of
 
-* The current [User Guide] draft
-* The [Javadoc], which is not a *tutorial* walkthrough, but is thorough,
-  specific, and current.
+*   The [User Guide].
+*   The [Javadoc], which is not a *tutorial* walkthrough, but is thorough,
+    specific, and current.
```
cpovirk, line 47: I think that the word "current" here was previous a contrast with the User Guide. Now that the User Guide is also current, I'd remove it.

(The spec is of course still a little out of date, but you cover that below, and it's of interest to fewer people.)

docs/docs/start-here.md
```diff
@@ -38,53 +42,58 @@ questions, which you can look for in the [Nullness Design FAQ]. If you like,
...
 
 ### Then if you're *really* interested
 
-* The [specification], written to be understood by owners of compilers and
-  static analysis tools. Note that it currently matches JSpecify 0.2, not 0.3.
-* Our [wiki] has about 20 informal, non-normative articles on various topics
-* Open [issues]
-* Or try it out...
+*   The [specification], written to be understood by owners of compilers and
+    static analysis tools. Note that it currently matches JSpecify 0.2, not 0.3.
```
cpovirk, line 52: Maybe change "0.3" to "1.0" here—or "and it will require some updates for 1.0" to make it sound less like we have already made every last decision? Or you could rephrase an even larger part of the sentence to just say more generally that it's out of date.

cpovirk (approved):
:+1:

PR #545, opened 2024-07-03
Update package name and remove issue references for "details common to all annotations".

netdpb:


cpovirk (approved):

netdpb:
I thought adding another commit would make the build-docs action run now. Let's see if it does on merge.

Issue #492, opened 2024-03-12
Not possible to annotation a return enum @Nullable from a method


alexandrenavarro:
Not possible to annotation a return enum with @Nullable from a method (compilation error).
Is it wanted?

kevinb9n:
Hi, can you share the sample code?

ascopes:
```java
enum Foo { BAR; }

static @Nullable Foo getBar(boolean isNull) {
  return isNull ? null : Foo.BAR;
}
```

does this not work?

netdpb:
That should work. Which nullness analyzer are you using? What error are you seeing?

agentgt:
@netdpb @kevinb9n @cpovirk  This is sort of orthogonal to this bug/issue but have you guys made it clear where to file bugs for the reference checker vs the spec itself?

I say this because I can see some future confusion of issues that are related with the spec (and doc) vs issues with the reference checker.

Right now in the https://jspecify.dev/docs/start-here/#how-can-i-get-involved

We have: 

> [File an issue](https://github.com/jspecify/jspecify/issues/new) to request a feature or if something is wrong.

Perhaps that should be clear that issues here should be about the spec and documentation? Then perhaps another link for the reference checker.

alexandrenavarro:
I have problem with a method like that

@Nullable
AEnum map(AnotherEnum anotherEnum)

agentgt:
@alexandrenavarro Can you paste the error output here. `javac` should not have a problem with you annotating the return of an enum method. Also can you paste the exact code in triple back ticks.

What I suspect might be your problem is if you do  have a true `javac` compiler error and not an analyzer error is that the enum is an inner class.

That is this will not work:

```java
@Nullable OuterClass.AEnum map(AnotherEnum anotherEnum) {}
```

It has to be:

```java
OuterClass.@Nullable AEnum map(AnotherEnum anotherEnum) {}
```

Or if you import the outer class:

```java
@Nullable AEnum map(AnotherEnum anotherEnum) {}
```

alexandrenavarro:
The compiler error was "Annotation not applicable to this kind of reference".

I understood the problem, it was linked to the package before the enum.

It was like that in reality :

@Nullable com.mypackage.AEnum map(AnotherEnum anotherEnum) {


By importing the enum (and removing the com.package), like :

@Nullable AEnum map(AnotherEnum anotherEnum) {

The problem is solved.

[...]

PR #552, opened 2024-07-10
Direct people to file an issue in the reference checker repo

netdpb:
Based on a comment from @agentgt in #492.


cpovirk (approved):
I was going to say that the reference checker also directs users there from its output, but it looks like it does not :)

https://github.com/jspecify/jspecify-reference-checker/blob/1a7b3edf6948f86b993a20f14272cc734fb6def1/src/main/java/com/google/jspecify/nullness/NullSpecChecker.java#L89

PR #555, opened 2024-07-11
Add usage instructions and whether-to-use advice

netdpb:


docs/docs/applying.md
```diff
@@ -0,0 +1,117 @@
...
+    complain.
+    *   In this case, it makes sense to suppress the nullness error from your
+        analyzer with a `@SuppressWarnings` annotation and document why your
+        logic is safe anyway.
+*   Maybe your code has an actual nullness bug? If so, annotating your code
+    helped you find it!
+    *   If you can't fix the bug right now, your nullness analysis tool should
+        allow you to suppress the error with a `@SuppressWarnings` annotation,
+        or some other mechanism that lets you keep track of these as TODOs.
+*   Maybe your nullness analyzer has a bug? (This is unlikely, but possible
```
cpovirk, line 93: I am trying to decide whether "unlikely" is too strong a word at this point... :)(

docs/docs/using.md
```diff
@@ -0,0 +1,151 @@
...
+
+1.  Check *all* annotations on array types. Any code that currently has a type
+    like `@Nullable Object[]` must change to `Object @Nullable []`. This change
+    is required by the syntax of [type-use annotations]. If you do *not* make
+    this change, your code will change from meaning "a nullable array of
+    objects" to "an array of nullable objects."
+
+1.  Rebuild your project, making changes as needed to correct build errors.
+    Build errors can arise from restrictions on placing type-use annotations.
+    For example, if you see an error about an annotation on a "scoping
```
cpovirk, line 81: In hindsight, I find the phrasing here a little weird. "Scoping construct" errors might be the _only_ such errors, not just an "example" of them. The part of this sentence that it is definitely an "example" is the part about `Map.Entry`.

> For example, if you see an error about an annotation on a "scoping construct" in `@Nullable Map.Entry<K, V>`, then you must change it to `Map.@Nullable Entry<K, V>`.

docs/docs/whether.md
```diff
@@ -0,0 +1,104 @@
...
+*   Do you have Kotlin users? Which compiler version do they use?
+*   Do you use whole-program annotation processors, such as [Dagger]?
+
+## If your Java code doesn’t use nullness annotations yet
+
+If your Java code doesn’t already use nullness annotations, we recommend that
+you [start using JSpecify annotations](using).
+
+Even if you are not currently using a nullness analyzer, applying JSpecify
+annotations can still provide benefits. * The annotations are useful
```
cpovirk, line 22: This list got turned into a paragraph with asterisks in it. If the paragraphing was intentional, then remove the asterisks; if not, re-listify it, and consider removing the transition words at the start of the later sentences/bullets.

netdpb:
Please review, bearing in mind that this is a PR to the `docs-1.0` branch. Merging it will not immediately update the public website.

docs/docs/whether.md
```diff
@@ -0,0 +1,104 @@
...
+to run *their* nullness analyzers on it. * Finally, annotating your code will
+make the eventual process of applying nullness analysis to your codebase easier.
+
+In particular, if your project includes Kotlin code, or Kotlin users depend on
+your code, JSpecify annotations will improve the null-safety of Kotlin code
+using your Java library. New versions of the Kotlin compiler
+[recognize and correctly interpret](#kotlin) the annotations, so JSpecify
+annotated Java code will have correct nullness typing in Kotlin.
+
+## Nullness checker support for JSpecify
```
cpovirk, line 34: It looks like the IntelliJ bullet hasn't been carried over yet.

docs/docs/whether.md
```diff
@@ -0,0 +1,104 @@
...
+[`-Xjspecify-annotations=strict` and
+`-Xtype-enhancement-improvements-strict-mode`
+flags](https://kotlinlang.org/docs/whatsnew1520.html#support-for-jspecify-nullness-annotations).
+The Kotlin team plans to emit errors by default in Kotlin 2.1.0.
+
+## Annotation processors
+
+If your project relies on annotation processors, like [Dagger], that interpret
+nullness annotations on symbols in the classpath, then you may need to wait to
+adopt JSpecify annotations until you can build with JDK 22. There was a bug in
```
cpovirk, line 84: I think we moved the "build with JDK 22" link (currently on line 96 below) up to here in the draft doc, so let's do it in the Markdown, too.

docs/docs/whether.md
```diff
@@ -0,0 +1,104 @@
...
+`javac` versions before JDK 22
+([JDK-8225377](https://bugs.openjdk.org/browse/JDK-8225377)) where
+[type-use](https://www.oracle.com/technical-resources/articles/java/ma14-architect-annotations.html)
+annotations (including JSpecify’s `@Nullable` and `@NonNull`) were not properly
+read from class files. JSR-305 and AndroidX nullness annotations are
```
cpovirk, line 89: Maybe rephrase this to include some transition like "This is in contrast to..."

cpovirk (approved):
Exciting!

I ended up with a few more comments, but this looks great.

docs/docs/applying.md
```diff
@@ -0,0 +1,117 @@
...
+If your code has no nullness annotations, we recommend starting with one class
+or package at a time, ideally one that depends on as few others as possible.
+Then move outward from called code to its callers.
+
+In the future, we'll have suggestions for
+[tools that can automate some of this process](https://github.com/jspecify/jspecify/issues/553).
+
+## 1. Look for nullable type usages
+
+Look for methods, parameters, and fields where some part of the type component
```
wmdietl, line 12:
```suggestion
Look for return types, parameters, and fields where some part of the type component
```

The `methods` doesn't make clear what part to look at. Alternatively, remove the `parameters`, as that would be subsumed by `methods`.

docs/docs/using.md
```diff
@@ -0,0 +1,151 @@
...
+</dependency>
+```
+
+Avoid using [provided] or [optional] scope.
+
+#### Gradle
+
+```groovy
+dependencies {
+  api("org.jspecify:jspecify:1.0.0")
```
wmdietl, line 34: This seems to only work for a `java-library` gradle plugin.
When I use this in a `java` gradle project I get `Could not find method api() for arguments [org.jspecify:jspecify:0.3.0] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.`

What is the recommendation for `java` projects? As they are not meant to be depended on, just `implementation`?

cpovirk: Oh, sorry, I may at one point in the past have known that Gradle had separate plugins for "Java" and "Java Library." Thanks for pointing that out.

We need, like, a flowchart here.... I guess that's probably true of adding Gradle dependencies in general. (Maybe things also differ for Android? Hopefully [not anymore](https://github.com/google/guava/commit/3d33cd45f5d15f58b9365b63478f4d1f936f4e85).)

I am wary of presenting `implementation` as one of the top-level options, given the concerns we discuss with that when used with _libraries_. But maybe we can rework the text below?

> - We recommend avoiding [implementation] in favor of [api] whenever api is available. When implementation is the only option (such as when you use [the Java plugin](https://docs.gradle.org/current/userguide/java_plugin.html) instead of [the Java Library plugin](https://docs.gradle.org/current/userguide/java_library_plugin.html)), then it is safe to use.
> - We recommend avoiding [compileOnlyApi], though you can use it if you're confident that no users will need the annotations at runtime.

(Aside: Possibly we should use code font for `provided`, `optional`, `implementation`, and `compileOnlyApi`?)

docs/docs/applying.md
netdpb: How's this?

docs/docs/applying.md
netdpb: How's this?

docs/docs/using.md
netdpb: How's this?

wmdietl (approved):
Changes look good, thanks!

docs/docs/using.md
cpovirk: I suspect that the java-library advice also applies to [com.android.library](https://developer.android.com/studio/projects/android-library). Hopefully there aren't many such other plugins, at least common ones, so we could just mention both? (I do like that your version is shorter than mine.)

(It also occurs to me that we could try to recommend against `compileJava`, which is sort of the java equivalent to java-library's `compileOnlyApi`. But that seems less important, since users would be hurting only themselves (not their users) and since they can hopefully generalize from the java-library advice. So I don't think we need to worry about that.)

docs/docs/using.md
```diff
@@ -25,18 +25,27 @@ guidance below.
...
+
 ```groovy
 dependencies {
   api("org.jspecify:jspecify:1.0.0")
 }
 ```
 
-Avoid using [implementation] or (if your users might need the annotations at
-runtime) [compileOnlyApi] configurations.
+Or, for projects that use the `java` plugin and don't publish libraries:
```
cpovirk, line 42: I don't know enough about the java plugin to know if "and don't publish libraries" is redundant or what.

My bigger fear here is still that, if someone in a hurry sees `implementation("org.jspecify:jspecify:1.0.0")` at the top level, that person will reach for that without considering `api`. My thinking would be that we could start this sentence with something like "Or, if you are using a plugin that does not support `api`, then...."

cpovirk (approved):

docs/docs/using.md
netdpb: Let's keep it simple for now. We can always add details.

docs/docs/using.md
```diff
@@ -0,0 +1,161 @@
...
+[`com.android.library`](https://developer.android.com/studio/projects/android-library)
+plugins, avoid using [`implementation`] or [`compileOnlyApi`] configurations:
+
+```groovy
+dependencies {
+  api("org.jspecify:jspecify:1.0.0")
+}
+```
+
+Or, if you're using a plugin that doesn't support the `api` configuration such
```
cpovirk, line 42:
```suggestion
Or, if you're using a plugin that doesn't support the `api` configuration, such
```

cpovirk (approved):

PR #560, opened 2024-07-15
Merge from main

netdpb:


cpovirk (approved):
Make sure to use "Create a merge commit" :)

netdpb:
I feel like this is an opportunity for rebasing, no?

cpovirk:
Oh, hmm, thanks. I had missed that we don't actually have any commits on `docs-1.0` at this point (beyond what's already on `main`). If we had any such commits, I'd want to think more about rebasing. But I agree that it should be fine here, assuming that GitHub's "Rebase and merge" button actually does what it sounds like it should. Now, I have my doubts about that.... But I guess it's worth a try. (My hope is that `main` and `docs-1.0` will end up pointing at the exact same ref/hash/whatever-it's-called, and my concern is that we _might_ instead get different copies of the 2 commits here. But sure, let's give it a try.)

PR #565, opened 2024-07-16
Update docs for 1.0.0

netdpb:


docs/docs/whether.md
```diff
@@ -0,0 +1,115 @@
+---
+sidebar_position: 3
+---
+
+# Should you use JSpecify annotations in your Java code?
+
+With the release of JSpecify 1.0, we guarantee backwards compatibility: we will
```
wmdietl, line 7:
```suggestion
With the release of JSpecify 1.0.0, we guarantee backwards compatibility: we will
```

I've seen that this was done in other places...

wmdietl (approved):
Looks great, just three small comments to consider.

cpovirk:
In hindsight, maybe I should have done all that as one commit. But I didn't see the button for it at first. Oh, well.

netdpb:
I will likely use `git` command line to rebase `docs-1.0` onto `main`. I can squash those individual commits.

netdpb:
This was done via #573 .

PR #569, opened 2024-07-17
Change OSGi BSN to org.jspecify.annotations or org.jspecify

chrisrueger:
This is a proposal based on discussion in https://github.com/jspecify/jspecify/pull/428#issuecomment-2231769425 

> I'm not convince of BSN org.jspecify.jspecify. The second jspecify is superfluous. Since Maven uses two parts to identify an artifact by groupId and artifactId OSGI uses a single identifier the BSN only.
> E.g. com.acme:acme-api -> com.acme.api or org.foo:foo -> org.foo as decribed in [Apache Felix Maven Bundle Plugin (BND) doc](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#_default_behavior)
> What's next with jspecify? Further annotations in the current bundle then org.jspecify.annotations is a good candidate. Or something like an uber jar for all kinds of things then org.jspecify is fine.

I liked the proposal by @bjmi `org.jspecify.annotations`

- It transports the intention of the bundle in its current state - which is about "annotations".

## But what about just org.jspecify ?

There is also the proposal to just use `org.jspecify`

- currently it seems that the module-info.java uses `org.jspecify` 

https://github.com/jspecify/jspecify/blob/5bf4ae46059cc4821879ecae10522b77dd16c59b/src/java9/java/module-info.java#L16

If this should be kept, then it also makes sense to use the same for the OSGi BSN / Bundle-SymbolicName

This PR should serve to decide which one to use.

### DIFF
--- a/bnd.bnd
+++ b/bnd.bnd
@@ -1,4 +1,4 @@
-Bundle-SymbolicName: org.jspecify.jspecify
+Bundle-SymbolicName: org.jspecify.annotations
 Bundle-Name: JSpecify annotations
 Bundle-Vendor: The JSpecify Authors
 Bundle-Description: An artifact of well-specified annotations to power static analysis checks and JVM language interop.


wmdietl:
[Earlier discussion](https://github.com/jspecify/jspecify/pull/428#discussion_r1678051990).

cpovirk:
Sorry, I see now that not only does the Felix doc suggest "org.jspecify" but that so too does [one doc I cited earlier](https://cwiki.apache.org/confluence/display/SMX4/Creating+a+Bundle+Using+Maven). (I had missed the section of that doc that gave actual recommendations in favor of the section that (as I theorized at the time) was demonstrating how to override those recommendations when necessary.) And sadly, that convention was not mentioned by [the recommendations I had dug up](https://webcache.googleusercontent.com/search?q=cache%3Ahttps%3A%2F%2Fdocs.redhat.com%2Fen%2Fdocumentation%2Fred_hat_jboss_fuse%2F6.0%2Fhtml%2Fdeploying_into_the_container%2Fbestpractices-buildbundles&oq=cache%3Ahttps%3A%2F%2Fdocs.redhat.com%2Fen%2Fdocumentation%2Fred_hat_jboss_fuse%2F6.0%2Fhtml%2Fdeploying_into_the_container%2Fbestpractices-buildbundles&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg60gEIMTAwN2owajeoAgCwAgA&sourceid=chrome&ie=UTF-8).

It is very possible that "org.jspecify" would be better, but I unfortunately didn't give this discussion time to develop after merging the original OSGi PR, and I pushed the release to Maven in the meantime. While we _could_ immediately release a 1.0.1 with the change, I think it will be simplest at this point to stick with what we have, partially (but not exclusively) because that it's still not 100% clear to me which approach (if either) is most conventional. (I do suspect that it would be "org.jspecify," but I think this is minor enough that it's not worth blocking on.) Sorry again for not giving time for this discussion. I do appreciate the link you dug up about the convention.

PR #564, opened 2024-07-16
1.0.0 release blog post

netdpb:
Fixes #456

PR #577, opened 2024-07-19
Fix half-broken links.

cpovirk:
These links _usually_ work. However, if you navigate from _an external
site_ to a URL like https://jspecify.dev/docs/using, then you'll end up
at https://jspecify.dev/docs/using/ (with a trailing slash). (Links
_within_ the site apparently get intercepted to load the content and
update the URL through the magic of JavaScript, thereby avoiding this
problem. Compare https://github.com/jspecify/jspecify/issues/411 and
friends.)

Anyway, once you're at https://jspecify.dev/docs/using/, links like
those in "see if the caveats about Kotlin or annotation processors apply
to your situation" are broken... sometimes. If you right-click and "Open
link in new tab," they work fine. (The browser sees a link to
https://jspecify.dev/docs/whether#kotlin, for example.) However, as soon
as you left-click one of them, you get a 404 because the JavaScript
kicks in and navigates you to
https://jspecify.dev/docs/using/whether#kotlin. Then (fun fact) if you
use your browser's Back button, then you'll find that not just one but
_both_ of the `whether` links in that sentence have changed to really
have their `href` set to the broken
https://jspecify.dev/docs/using/whether#... URL, so not even the
right-click approach will work anymore.

The fix for all this is to write our links in the style `/docs/whether`
instead of just `whether`. netdpb@ had already discovered this pattern,
but [the
lesson](https://github.com/jspecify/jspecify/pull/564#pullrequestreview-2180877440)
didn't stick for me, even after I initially published the release notes
with a broken link :)

### DIFF
--- a/docs/docs/start-here.md
+++ b/docs/docs/start-here.md
@@ -53,7 +53,7 @@ questions, which you can look for in the [Nullness Design FAQ]. If you like,
     JSpecify 0.2.0, not 1.0.0. We're working on updating it.
 *   Our [wiki] has about 20 informal, non-normative articles on various topics
 *   Open [issues]
-*   [Try it out](using)
+*   [Try it out](/docs/using)
 
 ### Reference implementation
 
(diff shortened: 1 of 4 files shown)


msridhar (approved):
👍

PR #578, opened 2024-07-23
Add trailing slashes to URLs.

cpovirk:
The resulting URLs look less clean, but they will be more consistent
and, more importantly, more reliable.

To recap:

- GitHub Pages adds a trailing slash if you navigate directly to a URL
  from an external source. See my discussion in
  https://github.com/jspecify/jspecify/pull/577, and see the table at
  https://github.com/slorber/trailing-slash-guide?tab=readme-ov-file#summary.
- Docusaurus currently does _not_ add a trailing slash during intra-site
  navigation. This leads to inconsistent results that depend on the
  source of the link. And that makes it easy for links to be broken in
  some cases but us not to notice—nor Docusaurus to report an error.
- This PR makes Docusaurus add a trailing slash. See
  https://docusaurus.io/docs/api/docusaurus-config#trailingSlash.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -25,6 +25,7 @@ const config = {
   tagline: 'Standard Annotations for Java Static Analysis',
   url: 'http://jspecify.org/',
   baseUrl: '/',
+  trailingSlash: true,
   onBrokenLinks: 'throw',
   onBrokenMarkdownLinks: 'throw',
   favicon: 'img/jspecify-favicon.ico',


cpovirk:
I downloaded the output from the "Upload artifact" section of https://github.com/jspecify/jspecify/actions/runs/10061111540/job/27810348394?pr=578, and I played around with it with a local `python3 -m http.server`. It appeared to be working, including adding the trailing slashes as expected.

I also tried creating a broken link like the ones we had before, and I confirmed that Docusaurus will now catch the problem. See https://github.com/jspecify/jspecify/pull/579.

(I suppose we could now try to get in the habit of linking to, e.g., /docs/whether/ instead of /docs/whether, since the former is where users will ultimately land. But this would be just about what we as developer see: Docusaurus already rewrites the links to include the trailing slash even in the rendered HTML, not only in the JavaScript navigation magic.)

One final thing that I should have noted is that another possibility would be (naturally enough) `trailingSlash: false`. My main fear with that is that it would break any existing links with trailing slashes in the wild (e.g., https://jspecify.dev/docs/spec/). (Again, see the table I linked above, which shows an X for the GitHub Pages's handling of /file/.) The risk of that is probably relatively low in the early days, so if we want to do it, now would be the time. I think that `trailingSlash: true` is superior in some ways (including that it shows fairly consistent behavior across hosting providers), but the trailing slash itself is obviously a bit of an eyesore from some people's perspective, and it's hard to know how to weigh actual fuzzy human factors against a solution that is theoretically "technically correct" :)

msridhar (approved):
LGTM

PR #579, opened 2024-07-23
Test breaking a link.

cpovirk:
Let's see if Docusaurus now catches this problem.

Before, it did not, so I needed to fix some links in
https://github.com/jspecify/jspecify/pull/577.

My hope is that now, after https://github.com/jspecify/jspecify/pull/578
(which is included in this PR), it will.

DO NOT MERGE

### DIFF
--- a/docs/docs/applying.md
+++ b/docs/docs/applying.md
@@ -61,7 +61,7 @@ be nullable, so you would declare `bar` to return `Optional<@NonNull T>` to
 indicate that it returns `Optional<String>` even for `Foo<@Nullable String>`.
 
 (There are other edge cases, such as wildcards. Consult the
-[User Guide](/docs/user-guide) for more information.)
+[User Guide](user-guide) for more information.)
 
 ## 2. Add `@NullMarked`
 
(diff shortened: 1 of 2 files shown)


cpovirk:
Excellent:

```
  Exhaustive list of all broken links found:
  - Broken link on source page path = /docs/applying/:
     -> linking to user-guide/ (resolved as: /docs/applying/user-guide/)
```

cpovirk:
I'm going to delete the PR branch, but for the record, the test change was:

```patch
diff --git a/docs/docs/applying.md b/docs/docs/applying.md
index 8009574..dbf33b7 100644
--- a/docs/docs/applying.md
+++ b/docs/docs/applying.md
@@ -61,7 +61,7 @@ be nullable, so you would declare `bar` to return `Optional<@NonNull T>` to
 indicate that it returns `Optional<String>` even for `Foo<@Nullable String>`.
 
 (There are other edge cases, such as wildcards. Consult the
-[User Guide](/docs/user-guide) for more information.)
+[User Guide](user-guide) for more information.)
 
 ## 2. Add `@NullMarked`
 
```

PR #583, opened 2024-08-07
Samples google prototype changes to work with main-eisop

wmdietl:
Goes together with https://github.com/jspecify/jspecify-reference-checker/pull/155


cpovirk (approved):
Somewhat separate from this PR, I kicked off a fresh run with main-eisop and samples-google-prototype-eisop to see where we stand. The results have more errors than I'd thought.

Here's what I ran:

```
$ cd $(mktemp -d)
$ git clone git@github.com:jspecify/jspecify-reference-checker.git
$ git clone git@github.com:jspecify/jspecify.git
$ git clone git@github.com:eisop/checker-framework.git
$ ( cd jspecify && git checkout samples-google-prototype-eisop )
$ cd jspecify-reference-checker/
$ git checkout main-eisop

# It would have been simpler to use some `-b` flags during the `clone` commands....

# And, at the time:
$ JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 ./gradlew jspecifySamplesTest --include-build ../jspecify --include-build ../checker-framework

# In the future, I may want to switch to:
$ JAVA_HOME=$HOME/jdk-21.0.2 ./gradlew jspecifySamplesTest --include-build ../jspecify --include-build ../checker-framework
```

If I cram all the results together, I see this:

```
    1198 out of 1217 expected diagnostics were found.
    18 unexpected diagnostics were found:
    19 expected diagnostics were not found:
    7 out of 7 expected diagnostics were found.
    1 unexpected diagnostic was found:
    1208 out of 1217 expected diagnostics were found.
    180 unexpected diagnostics were found:
    9 expected diagnostics were not found:
    21 out of 21 expected diagnostics were found.
    7 unexpected diagnostics were found:
    7 out of 7 expected diagnostics were found.
    1 unexpected diagnostic was found:
    2 out of 2 expected diagnostics were found.
    2 unexpected diagnostics were found:
```

The big one is of course the "180 unexpected diagnostics were found." My impression is that `Foo<?>` is now considered to be out of bounds if we have `@NullMarked class Foo<T extends Object> {}`, at least sometimes. I don't know if that's a recent regression or something that's been around a while. Here's [an example](https://github.com/jspecify/jspecify/blob/98adea63133a1aa0bf629bc041728376a0438ece/samples/wildcardsWithDefault/wildcardswithdefault/WildcardsWithDefault.java#L24) that now produces an error:

```
    2 unexpected diagnostics were found:
      WildcardsWithDefault.java:24: error: (type.argument.type.incompatible)
      WildcardsWithDefault.java:27: error: (type.argument.type.incompatible)
```

I somehow _don't_ see the error in some Google-internal code that refers to `Cache<?, ?>` (when both type parameters of `Cache` are non-nullable), so maybe it has to do with the explicit `extends Object`, or maybe it happens only when everything is all in one compilation unit, or maybe something else is up. Do you have any theories?

Beyond that, I notice:

- `CastToPrimitive` no longer has errors. I think we were promoting a _warning_ to an error there, and we were doing it in some hacky way, so probably we lost it. I think maybe I'd been considering ways to address this better a while back, I think involving making `int?` a real type instead of treating every `int` as non-nullable? I think this one is well-understood enough that we could suppress it and file a bug.
- `OverrideParameters` shows that we no longer produce errors for widening parameters. That's probably fine for users, though we'll eventually want a way to identify those cases for conformance-test purposes.
- `DereferenceIntersection` now has errors were they were desired but previously missing. I'll see if I see any other opportunities to remove `jspecify_but...`.
- \[edit: We also have the errors that will be fixed by https://github.com/jspecify/jspecify-reference-checker/pull/197. With that fix, the "18" goes down to 11, and the "180" goes down to 146.\]

cpovirk:
> * `DereferenceIntersection` now has errors were they were desired but previously missing. I'll see if I see any other opportunities to remove `jspecify_but...`.

If my calculations are correct, we can remove `jspecify_but...` from all the following:

```
CaptureAsInferredTypeArgument.java:41
CaptureAsInferredTypeArgument.java:43
CaptureAsInferredTypeArgument.java:45
CaptureConvertedToObject.java:52
CaptureConvertedToObject.java:57
CaptureConvertedToObject.java:66
CaptureConvertedToObjectUnspec.java:52
CaptureConvertedToObjectUnspec.java:57
CaptureConvertedToObjectUnspec.java:66
CaptureConvertedToOther.java:52
CaptureConvertedToOther.java:57
CaptureConvertedToOther.java:66
CaptureConvertedToOtherUnspec.java:52
CaptureConvertedToOtherUnspec.java:57
CaptureConvertedToOtherUnspec.java:66
CaptureConvertedTypeVariableBounded.java:77
ContainmentExtends.java:26
ContainmentExtends.java:34
DereferenceIntersection.java:59
DereferenceIntersection.java:65
DereferenceIntersection.java:76
NotNullMarkedContainmentExtends.java:22
NotNullMarkedContainmentExtends.java:25
NotNullMarkedContainmentExtends.java:31
NotNullMarkedContainmentExtends.java:34
NotNullMarkedTypeVariableBound.java:53
NotNullMarkedTypeVariableBound.java:58
NotNullMarkedTypeVariableBound.java:63
NotNullMarkedUseOfWildcardAsTypeArgument.java:32
NotNullMarkedUseOfWildcardAsTypeArgument.java:37
UseOfTypeVariableAsTypeArgument.java:69
```

cpovirk:
(https://github.com/jspecify/jspecify/pull/584 for those lines)

cpovirk:
I had the thought that maybe the wildcard errors are related to undoing the capture conversion "wonkyness." I haven't followed up, but I'll try to do so eventually.

cpovirk:
Oh, but also I had some other memory of hacking up the fork to disable some check on the grounds that it should be unnecessary. I think it may have been related to capture conversion, too? I'll dig that up, maybe tomorrow.

cpovirk:
I'm taking the specific issues I've mentioned above to separate issues:
- https://github.com/jspecify/jspecify-reference-checker/issues/198
- https://github.com/jspecify/jspecify-reference-checker/issues/199
- https://github.com/jspecify/jspecify-reference-checker/issues/200

I'm not expecting to fix them all before merging `main-eisop` into `main`, but:
- I want to at least look at https://github.com/jspecify/jspecify-reference-checker/issues/200 while this is all fresh in my mind.
- I want to be able to reference those issues when we suppress any remaining errors.

PR #584, opened 2024-08-08
Remove `jspecify_but_...` where incorrect behavior has been fixed.

cpovirk:
This makes `NullSpecTest$Strict` go from having 180 unexpected
diagnostics to having 146.


wmdietl (approved):
Thanks!

PR #585, opened 2024-08-08
Rebase specification-1.0 onto main

netdpb:


netdpb:
Does this seem like the right git thing to do, @cpovirk?

netdpb:
Hmm. The text for "Rebase and merge" looks wrong. I don't want to rebase those commits onto this branch; I want to rebase this branch's commits onto main's.

image

I can do this in the CLI with

```sh
$ git checkout specification-1.0
$ git rebase main
```

Maybe I should just do that and push?

wmdietl (approved):
Pulling in `main` lgtm.

netdpb:
> Pulling in `main` lgtm.

Do you mean using `git`'s CLI as I suggested in https://github.com/jspecify/jspecify/pull/585#issuecomment-2276624121?

wmdietl:
> > Pulling in `main` lgtm.
> 
> Do you mean using `git`'s CLI as I suggested in [#585 (comment)](https://github.com/jspecify/jspecify/pull/585#issuecomment-2276624121)?

If you want to keep only the commits in `specification-1.0` that are new to it, you can do

````
$ git checkout specification-1.0
$ git pull origin main --rebase
````

That should rebase the `specification-1.0` commits against what is in `main`.
However, you will then need to force-push your local `specification-1.0` to origin, as there are now two different histories.

I did this for a while for `main-eisop` branches, but it is rather annoying, as you need to make sure to re-clone all your copies of `specification-1.0`.
In the end, `specification-1.0` will be squash-merged into `main`, so for me it would be okay to just pull in `main` into `specification-1.0` and have the extra merge commit.

netdpb:
> In the end, `specification-1.0` will be squash-merged into `main`, so for me it would be okay to just pull in `main` into `specification-1.0` and have the extra merge commit.

I'm not sure. Maybe we'll want to do a fast-forward merge.

I'm going to try that.

cpovirk (approved):
I'm fine with a force-push for the rebased branch.

If that doesn't work, I'm also fine with a merge or probably with whatever :)

netdpb:
Done. And GitHub automatically closed this PR because all its proposed commits are now on the `specification-1.0` branch.

PR #598, opened 2024-08-15
Update Docusaurus to 3.5.2.

netdpb:
Dependabot doesn't know that all the Docusaurus dependencies have to be in sync, so #595 and #596 fail.


cpovirk (approved):
Yay, glad that it's not something more complex. Thanks for figuring it out.

netdpb:
I guess I could have just updated one of the Dependabot branches instead.

PR #599, opened 2024-08-19
Perform dependency updates in weekly batches.

cpovirk:
Compare https://github.com/google/truth/pull/1304

This would presumably have avoided the problem discussed in
https://github.com/jspecify/jspecify/pull/598.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,28 +4,28 @@ updates:
   directory: "/"
   schedule:
     interval: "weekly"
-	groups:
-		dependencies:
-			applies-to: version-updates
-			patterns:
-				- "*"
+  groups:
+    dependencies:
+      applies-to: version-updates
+      patterns:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -3,14 +3,29 @@ updates:
 - package-ecosystem: "gradle"
   directory: "/"
   schedule:
-    interval: "daily"
+    interval: "weekly"
+	groups:
```
netdpb, line 7: Is this too much indentation? Or did tabs sneak in here?

netdpb (approved):

.github/dependabot.yml
cpovirk: Sigh, thanks. Better now?

(I figured out a while back that I end up with tabs when I `set paste` in Vim....)

netdpb (approved):

PR #601, opened 2024-08-19
Phrasing and non-normative changes to Substitution.

cpovirk:
- Use "output" consistently instead of a mix of that and "result." This
  is also more consistent with "input."
- Mention another case in which we're likely to require that checkers
  implement the all-worlds version of a rule.


netdpb (approved):

cpovirk:
> Mention another case in which we're likely to require that checkers implement the all-worlds version of a rule.

I just wanted to write down here that I expect that the all-worlds version of a rule might well also end up required during type inference—not that we have done any meaningful work toward understanding how type inference will work, let alone specifying it :)

PR #603, opened 2024-08-19
Fix bug in null-exclusivity rules.

cpovirk:
The bug:

As discussed under "My implementation suggestion is this" in [this
post](https://github.com/jspecify/jspecify-reference-checker/pull/175#discussion_r1567591777)
and especially my _following_ post, the type `null` is just as
null-exclusive as `String` or `Object[]`.

The non-normative bug:

I think I just wrote this backward? It's the "lenient" / "some-world"
rules under which a type like `Foo*` can be both null-inclusive ("You
can put `null` into it") and null-exclusive ("You can dereference it").


docs/docs/spec.md
```diff
@@ -726,10 +726,10 @@ A type is null-exclusive under every parameterization if it has a
 [nullness-subtype-establishing path] to either of the following:
 
 -   any type whose [nullness operator] is `MINUS_NULL`
--   any augmented class or array type
+-   any augmented class or array type or the augmented null type
```
wmdietl, line 729:
```suggestion
-   any augmented class, array, or null type
```

The `augmented` is supposed to apply to all three, right?
Alternatively, we could add `augmented` to `array type` as well.

wmdietl: I have a hard time interpreting this. At first it seems like the "augmented" means that as soon as any class type is reached, including e.g. `Object?`, which seems wrong - how would that establish non-exclusiveness?

However, it is reachable through a `Nullness-subtype-establishing path`... which actually establishes that it is non-null? Would `Non-nullness-subtype-establishing path` be a better way to describe these paths?

Or am I missing something else?

docs/docs/spec.md
```diff
@@ -631,7 +631,7 @@ The same-type relation is *not* defined to be reflexive or transitive.
     > having to cover that case.)
 
 > A further level of complexity in all this is `UNSPECIFIED`. For example, in
-> the [all-worlds] version of the following rules, a type with nullness operator
+> the [some-world] version of the following rules, a type with nullness operator
```
wmdietl, line 634: `some-world` does make sense to me. It would be nice to add something like the two examples from the PR description `null-inclusive ("You can put null into it") and null-exclusive ("You can dereference it").`

docs/docs/spec.md
```diff
@@ -726,10 +726,10 @@ A type is null-exclusive under every parameterization if it has a
 [nullness-subtype-establishing path] to either of the following:
 
 -   any type whose [nullness operator] is `MINUS_NULL`
--   any augmented class or array type
+-   any augmented class or array type or the augmented null type
 
-    > This rule refers specifically to a "class or array type," as distinct from
-    > other types like type variables and [intersection types].
+    > This rule refers to particular kinds of types as distinct from other types
```
wmdietl, line 731: Maybe I'm confused by this distinction as well: doesn't `Nullness-subtype-establishing path` already make the required distinction between different types, through the `Nullness-subtype-establishing direct-supertype edges`?

wmdietl (commented):
A few questions, but I need to read through these definitions again in more detail.

docs/docs/spec.md
cpovirk: Yes, `augmented` applies to all three. I think I was just feeling squeamish about saying "any... null type" as if there are a whole bunch of different null types. But that's not worth making the text more confusing for.

(Also, oops: There _are_ in fact multiple different null types :) And nullness-subtype-establishing path could lead to `null NO_CHANGE`, `null UNSPECIFIED` (in the "some-world" variant of the rule), or `null MINUS_NULL` (if that comes up in practice. That leads into your next question....)

---

Indeed, we don't want a path to `Object?` to establish null-exclusiveness. We avoid that by defining ["Nullness-subtype-establishing direct-supertype edges"](https://jspecify.dev/docs/spec/#nullness-subtype-establishing-direct-supertype-edges) to exist only when the destination node has an appropriate nullness operator. (Similarly, it's important that `Object?` doesn't have a path to itself (an empty path, which we otherwise permit) so that we don't treat _`Object?`_ as null-exclusive. That's handled by the nullness-operator requirement in the definition of ["Nullness-subtype-establishing path."](https://jspecify.dev/docs/spec/#nullness-subtype-establishing-path) It's possible that there's a way to deduplicate the nullness-operator checking between those two sections, even beyond what https://github.com/jspecify/jspecify/pull/602 could achieve.)

---

As for the phrasing: In this section, a "nullness-subtype-establishing path" is used only to establish non-nullness. However, it's also used in the definition of ["Nullness subtyping"](https://jspecify.dev/docs/spec/#nullness-subtyping), where it can establish _subtyping_ between types even when both those types can include null, as in:

```java
@NullMarked
class Foo<T extends @Nullable Object, U extends T> {}
```

In that example, we can determine that `U NO_CHANGE` is a nullness subtype of `T NO_CHANGE` by following the `U`->`T` nullness-subtype-establishing path.

docs/docs/spec.md
cpovirk: Sounds good. How do you feel about the existing non-normative comment at the end of the [null-exclusive](https://jspecify.dev/docs/spec/#null-exclusive-under-every-parameterization) section?
- Is it a clear description of the dereferencing use case?
- Does it work well for it to be at the _end_ of the section instead of the beginning?
- Would the comment fit better in some other section entirely?

docs/docs/spec.md
cpovirk: "Nullness-subtype-establishing direct-supertype edges" can include type variables, such as to support the `U`->`T` case from my previous message. To establish null-exclusivity, then, we need to reach an "actual type" (or a non-null-projected type variable). Otherwise, we'd consider `U` to be null-exclusive once we reach `T`. We instead want to look through `T` to the bound. Since the bound is `@Nullable Object`, `U` (like `T`) is _not_ null-exclusive. If it bound were `@NonNull Object`, then both type variables would be null-exclusive. (And if the bound were yet another type variable, then we'd keep looking :))

cpovirk (commented):
(GitHub-UI complaint: Somehow only _now_ do I see a "Start a review" button for my comment? Oh, I guess that I receive that button only if I'm looking at the `files`/diff view? In the "main" discussion page, my only option appears to be to post individual responses as separate emails. Maybe I'll remember that next time.)

cpovirk:
@wmdietl , I can't remember when your schedule gets back to normal, so when you have a chance, let me know if you think you will have time for another look at this this week. I'd like to make the case at the meeting tomorrow that the spec is basically ready for 1.0 :) I can probably get netdpb to take a look if that would help.

cpovirk:
netdpb also encountered the non-normative bug and ended up fixing it in another commit. So now all that this one does is fix the null-exclusivity bug.

wmdietl (approved):
The clarification LGTM. The PR title should maybe be changed after the non-normative change moved to a different PR.

I will re-read the rules, but that doesn't block rc1.

Issue #604, opened 2024-08-21
Reenable Error Prone


cpovirk:
Doing so naively...

```diff
--- a/build.gradle
+++ b/build.gradle
@@ -19,13 +19,13 @@ plugins {
     id 'com.diffplug.spotless' version '6.25.0'
     id 'com.github.node-gradle.node' version '7.0.2'
     id 'io.github.gradle-nexus.publish-plugin' version '1.3.0' apply false
-    id 'net.ltgt.errorprone' version '3.0.1'
+    id 'net.ltgt.errorprone' version '4.0.1'
     id "biz.aQute.bnd.builder" version "7.0.0"
 }
 
 ext {
     junitVersion = '5.9.3'
-    errorproneVersion = '2.20.0'
+    errorproneVersion = '2.30.0'
 }
 
 repositories { mavenCentral() }
```

...leads to an error...

```
$ JAVA_HOME=$HOME/jdk-21.0.2 ./gradlew clean assemble

FAILURE: Build failed with an exception.

* What went wrong:
Circular dependency between the following tasks:
:classes
\--- :compileJava
     \--- :jar
          +--- :classes (*)
          +--- :compileJava (*)
          +--- :compileJava9Java
          |    \--- :jar (*)
          \--- :java9Classes
               \--- :compileJava9Java (*)

(*) - details omitted (listed previously)


* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 777ms
```

...presumably because Error Prone now depends on JSpecify.

It's possible that we should just disable Error Prone, at least for our prod code, given how trivial that prod code (i.e., just the annotations) is :)

tbroyer:
Did you try without the `includeBuild('.')`?
(adapting the conformance-tests project dependencies to use `project(':')` instead of the coordinates)

cpovirk:
Nope, because I remain totally unqualified to maintain a Gradle build :)

Removing `includeBuild('.')` does indeed help. Assuming that I interpreted `project(':')` correctly, I've also added that, and it at least is enough for `./gradlew assemble` to work. I've created #605 to run the rest of the CI tasks.

Thank you!

wmdietl:
#605 needed to be reverted with #609, because it caused a circular dependency: https://github.com/jspecify/jspecify-reference-checker/issues/202.

[...]

PR #605, opened 2024-08-21
Upgrade Error Prone to 2.30.0.

cpovirk:
And break the resulting circular dependency as advised by @tbroyer in
https://github.com/jspecify/jspecify/issues/604#issuecomment-2302220782.

### DIFF
--- a/conformance-tests/build.gradle
+++ b/conformance-tests/build.gradle
@@ -68,8 +68,8 @@ distributions {
 }
 
 dependencies {
-    assertionsImplementation "org.jspecify:jspecify:0.0.0-SNAPSHOT"
-    depsImplementation "org.jspecify:jspecify:0.0.0-SNAPSHOT"
+    assertionsImplementation project(":")
+    depsImplementation project(":")
 }
 
 // Make sure assertions compile.
--- a/conformance-tests/build.gradle
+++ b/conformance-tests/build.gradle
@@ -68,6 +68,9 @@ distributions {
 }
 
 dependencies {
+    // `project(":")` refers to the root project.
+    // To make this clearer, we should move the actual annotations to a subproject.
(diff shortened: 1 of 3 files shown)


tbroyer (approved):

conformance-tests/build.gradle
```diff
@@ -68,8 +68,8 @@ distributions {
 }
 
 dependencies {
-    assertionsImplementation "org.jspecify:jspecify:0.0.0-SNAPSHOT"
-    depsImplementation "org.jspecify:jspecify:0.0.0-SNAPSHOT"
+    assertionsImplementation project(":")
```
wmdietl, line 71: I would find a comment helpful here. I did a quick search and didn't find what `project(":")` means.

wmdietl (approved):
One comment to consider...

conformance-tests/build.gradle
tbroyer: The `project(…)` notation takes the "project path" to the depended-on project, and because the JSpecify "prod code" is in the root project, its path is `:`: https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:project_dependencies / https://docs.gradle.org/current/userguide/intro_multi_project_builds.html#sec:project_path

Alternatively, one can also directly pass a project instance, so this should work too but is less idiomatic:
```groovy
    assertionsImplementation rootProject
```

wmdietl: I would find `rootProject` much more self-explanatory, but if `project(":")` is the more idiomatic way, that's also fine with me.

tbroyer: Even more _idiomatic_ would be to move the code to a subproject (leaving the root project empty) so you'd use something like `project(":lib")` here :wink:

cpovirk: I've added a comment, and I've included the recommendation that we redo the project structure.

PR #609, opened 2024-08-26
Stop running Error Prone.

cpovirk:
Running it is good practice, but running it has also gotten more complex
now that Error Prone depends on JSpecify.

Luckily, there is only so much that we can mess up in the declarations
of the annotations (and only a little more than that once you include
the integration tests).

(Please don't think too hard about the fact that Error Prone is
currently producing legitimate InvalidBlockTag and
UnrecognisedJavadocTag warnings because we're using a JDK11 toolchain
for Javadoc at the moment....)

Fixes https://github.com/jspecify/jspecify-reference-checker/issues/202

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,13 +19,11 @@ plugins {
     id 'com.diffplug.spotless' version '6.25.0'
     id 'com.github.node-gradle.node' version '7.0.2'
     id 'io.github.gradle-nexus.publish-plugin' version '1.3.0' apply false
-    id 'net.ltgt.errorprone' version '4.0.1'
     id "biz.aQute.bnd.builder" version "7.0.0"
 }
 
 ext {
     junitVersion = '5.9.3'
-    errorproneVersion = '2.30.0'
 }
 
 repositories { mavenCentral() }
@@ -37,10 +35,6 @@ apply from: "gradle/mrjar.gradle"
 apply from: "gradle/integration-test.gradle"
 apply from: "gradle/format.gradle"
 apply from: "gradle/publish.gradle"
-
-dependencies {
-    errorprone "com.google.errorprone:error_prone_core:${errorproneVersion}"
-}
 
 java {
     if (javaVersion.canCompileOrRun(15)) {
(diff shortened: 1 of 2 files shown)


wmdietl (approved):
I tried a few `excludes`, similar to [this code](https://github.com/eisop/checker-framework/blob/dff6bd53633d8cb00ecb873a2f46de12a1573982/build.gradle#L182), but didn't figure out the problem.
I don't see how Error Prone can cause the circular dependency and am sad to see it go, but removing it for now seems like the best option.
If somebody figures out the dependency issue we can add it again.

cpovirk:
Thanks for investigating more targeted approaches.

cpovirk:
> Error Prone is
> currently producing legitimate InvalidBlockTag and
> UnrecognisedJavadocTag warnings because we're using a JDK11 toolchain
> for Javadoc at the moment....

Good news: I'm wrong. While the toolchain for javac (which of course hosts Error Prone) and javadoc [defaults to JDK 11](https://github.com/jspecify/jspecify/blob/0ab79c51e0b876fe2bc3bba6dbed5b0b80ac6596/build.gradle#L31), we skip Javadoc generation [unless we're running under JDK 15+](https://github.com/jspecify/jspecify/blob/0ab79c51e0b876fe2bc3bba6dbed5b0b80ac6596/build.gradle#L40). And we generate the docs at https://jspecify.dev/docs/api/ [with JDK 22](https://github.com/jspecify/jspecify/blob/0ab79c51e0b876fe2bc3bba6dbed5b0b80ac6596/.github/workflows/docs.yml#L31).

We should probably have Gradle automatically download and use JDK 22 as the toolchain for everything (except of course any tests that we want to run under older versions, such as those we run under JDK 8). My understanding is that that [would not be hard](https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning), and I might have even gone through that process in some other context recently?

ben-manes:
This sounds like the problem I faced when EP started depending on caffeine. I had to add the jar to the classpath [manually](https://github.com/ben-manes/caffeine/blob/2e31df3eefab01d0faca40571a37a81327576d97/gradle/plugins/src/main/kotlin/quality/errorprone.caffeine.gradle.kts#L25).

PR #617, opened 2024-09-03
Delete "under every parameterization". It's simpler.

netdpb:


cpovirk:
Mainly, I think it's potentially confusing to say: "This value is not null-inclusive, but you can't dereference it because it might be `null`." That's the case for a case like:

```java
@NullMarked
class Foo<T extends @Nullable Object> {
  T t;
}
```

`t` has type `T NO_CHANGE`, which can "hold" `null` if the type argument is nullable. So it would seem strange to answer "Is `t` null-inclusive?" with the answer "no." I think that "null-inclusive under every parameterization" makes that clearer.

It's still possible that we're better off spelling that out explicitly somewhere and then using shorthand. Even then, I might prefer inventing weirder shorthand, like "omni-null-inclusive" or something? That may be less clear that it's about type parameters, though (as opposed to being about "multiple worlds").

netdpb:
Interesting. I still wish we had a simpler term.

Maybe the confusing thing is that we have two levels of qualifier: type parameters and "worlds". So you can't easily use "always" or "never" because it's unclear over which qualifier.

Does "null-inclusive under every parameterization" (NIUEP) correspond to "nullable" in the nullable/non-null/parametric/unspecified taxonomy? And does NEUEP correspond to "non-null"? And does being neither (`T NO_CHANGE` with a nullable bound) correspond to "parametric"?

cpovirk:
I agree that the two levels of "always" are what make this tricky.

---

NIUEP corresponds to a combination of "nullable" and "unspecified (if we're requiring the check to hold only in some worlds)."

---

NEUEP is trickier because the nullable/non-null/parametric/unspecified taxonomy doesn't have a clear term for cases like this one:

```java
@NullUnmarked
class Foo<T> {
  @NullMarked
  T get();
}
```

`get()` returns `T NO_CHANGE`, where `T` has a bound of `Object UNSPECIFIED`. The nullable/non-null/parametric/unspecified taxonomy would probably call that "non-nullable," but that's an awkward fit: A `Foo<@Nullable String>` would produce a `@Nullable String` if you call `get()` on it. Accordingly, a strict checker would produce an error if `Foo` declared a `getAsString()` method that returned `get().toString()` (since that's a possible null dereference).

That example, like `String UNSPECIFIED`, would be considered NEUEP _if_ we're requiring the check to hold only in some world (since there's a world in which the type-parameter bound is `@NonNull Object`).

---

Parametric is indeed the main case in which a type is neither NIUEP nor NEUEP. The only(?) other way for a type to be neither NIUEP nor NEUEP is for it to have the operator `UNSPECIFIED` and for us to be requiring that the check holds in all worlds.

netdpb:
What about "invariably non-nullable (in all/some worlds)" and "invariably nullable (in all/some worlds)"?

netdpb:
Let's not worry about this now.

PR #592, opened 2024-08-13
Update recognized locations for the annotation(s)

netdpb:

PR #619, opened 2024-09-05
Update the spec for `@NonNull`.

cpovirk:


docs/docs/spec.md
```diff
@@ -135,9 +135,9 @@ A nullness operator is one of 4 values:
 >         etc., plus `null`.
 >     -   The type-variable usage `T UNION_NULL` includes all members of `T`,
 >         plus `null` if it wasn't already included.
-> -   `NO_CHANGE`: This is the operator produced by *not* putting `@Nullable` on
->     a type usage (aside from the exception discussed under `UNSPECIFIED`
->     below).
+> -   `NO_CHANGE`: This is the operator produced by putting neither `@Nullable`
```
netdpb, line 138: It might be nicer to define the operators' meanings without referring to how to determine which operator applies based on annotations, which is in a later section.

* `UNION_NULL` means that `null` is a valid value.
* `NO_CHANGE` means the same as `MINUS_NULL` except for type variables, where it means that `null` is a valid value only if it is for the substituted type argument.
* `UNSPECIFIED` means that we don't know whether `null` is a valid value.
* `MINUS_NULL` means that `null` is not a valid value.

docs/docs/spec.md
```diff
@@ -209,13 +212,13 @@ For all named annotations referred to by this spec:
 All annotations have runtime retention. None of the annotations are marked
 [repeatable].
 
-## The type-use annotation
+## The type-use annotations
 
-We provide a parameterless type-use annotation called `@Nullable`.
+We provide two parameterless type-use annotations, `@Nullable` and `@NonNull`.
```
netdpb, line 217:
```suggestion
We provide two parameterless type-use annotations: `@Nullable` and `@NonNull`.
```

docs/docs/spec.md
```diff
@@ -209,13 +212,13 @@ For all named annotations referred to by this spec:
...
 
 ### Recognized locations for type-use annotations
 
-A location is a *recognized* location for our type-use annotation in the
+A location is a *recognized* location for our type-use annotations in the
```
netdpb, line 221: Should you update the non-normative comment below as well?

docs/docs/spec.md
```diff
@@ -366,8 +369,10 @@ usage, this section covers only how to determine its [nullness operator].
 To determine the nullness operator, apply the following rules in order. Once one
 condition is met, skip the remaining conditions.
 
--   If the type usage is annotated with `@Nullable`, its
-    nullness operator is `UNION_NULL`.
+-   If the type usage is annotated with `@Nullable` and *not* annotated with
+    `@NonNull`, its nullness operator is `UNION_NULL`.
+-   If the type usage is annotated with `@NonNull` and *not* annotated with
+    `@Nullable`, its nullness operator is `MINUS_NULL`.
```
netdpb, line 375:
```suggestion
-   If the type usage is annotated with `@Nullable` and *not* with
    `@NonNull`, its nullness operator is `UNION_NULL`.
-   If the type usage is annotated with `@NonNull` and *not* with
    `@Nullable`, its nullness operator is `MINUS_NULL`.
```

netdpb: Also, do you want to non-normatively point out that if both annotations appear on the same type use, then it is as if neither does?

docs/docs/spec.md
```diff
@@ -379,20 +384,9 @@ condition is met, skip the remaining conditions.
...
-> the section on [substitution]. However, its main purpose is to provide tools
-> with a way to represent the nullness of certain expressions in implementation
-> code: Consider `ArrayList<E>`. `ArrayList` supports null elements, so the
-> class has to handle the possibility that any expression of type `E` may be
-> null. However, if implementation code contains the statement `if (e != null) {
-> ... }`, then tools can assume that `e` is non-null inside. The purpose of
-> `MINUS_NULL` is to represent that such an expression is known not to be null,
-> even though its base type `E` suggests otherwise.
+> If tool authors prefer, they can safely produce `MINUS_NULL` in any case in
+> which it is equivalent to `NO_CHANGE`. For example, there is no difference
```
netdpb, line 388: Is that just non-type-variable usages?

docs/docs/spec.md
cpovirk: Here, I'm kind of throwing a bunch of explanations at the wall to see what will stick for any given person. It's true that this explanation is a sort of forward reference, but I think it would be overly coy to _not_ yell "`UNION_NULL` is `@Nullable`!" (The mention is `@Nullable` is probably especially welcome after the lecture I give above about how I'm not going to talk about a type "is nullable" :)) I do think that the sort of explanation that you give is also valuable, and I have an ancient attempt at that in the sub-bullets.

Do you think that reshuffling and tweaking what we currently have could improve matters? Is there just _too much_ here?

docs/docs/spec.md
cpovirk: Sorry, in what way?

(Maybe this thread will end up being obsolete now that your #592 removes most of that comment.)

docs/docs/spec.md
cpovirk: Sure, that sounds fine.

docs/docs/spec.md
cpovirk: Doing so for intersection types would also be a problem (and I guess maybe union types if we got into those, which maybe we should now that we're getting into implementation code more: #621).

It's probably just those two (or three). Worth mentioning?

docs/docs/spec.md
netdpb: I think `@Nullable` and `@NonNull` represent things that people have seen before, so using that to signify what those operators mean makes sense. I think saying "completely unannotated code" is helpful for `UNSPECIFIED`.

But I don't think that this statement for `NO_CHANGE` will be helpful to someone who hasn't read the rest of the spec.

Maybe part of the problem is that we're using these operators in two ways: as a component of an augmented type (base type + operator) and as an input to the substitution rule (where we apply a nullness operator to an augmented type to produce a new augmented type). `NO_CHANGE` is not obvious in the first case, but makes a lot of sense in the second. That's part of why I'd like to change the rules so that `NO_CHANGE` only applies to type variable usages (and, I guess, intersection types). But we don't have to do that for 1.0.

docs/docs/spec.md
netdpb: I mean this part:

> For now, we've chosen to restrict ourselves to API locations for which tools
> mostly agree on what it means for a type in that location to be `@Nullable`.

What about the other one?

Happy for this to go away if it does.

docs/docs/spec.md
netdpb: If you have a specific idea of which cases you mean, please list them. But list the smaller set. Is it shorter to list type variables and intersection/union types?

docs/docs/spec.md
cpovirk: Ah, right. Since this is going in after your PR, we should be good to go.

docs/docs/spec.md
cpovirk: They might just be the same length :) Have a look at the two lists I've included, and see what you think about which one(s) to keep.

(See https://github.com/jspecify/jspecify/issues/621 for a relevant note about union types.)

docs/docs/spec.md
cpovirk: Thanks, that's very helpful! Now you get to be the judge of whether I put that helpful comment to good use :)

docs/docs/spec.md
```diff
@@ -133,40 +133,34 @@ A nullness operator is one of 4 values:
...
->     `UNSPECIFIED` below).
+>     -   The type-variable usage `T UNION_NULL` includes all members of the
+>         type argument substituted in for `T`, plus `null` if it wasn't already
+>         included.
+> -   `MINUS_NULL`: This is the operator produced by putting `@NonNull` on a
+>     type usage.
+>     -   The type usage `String MINUS_NULL` includes `"a"`, `"b"`, `"ab"`,
+>         etc., without including `null`.
+>     -   The type-variable usage `T MINUS_NULL` includes all members of the
+>         type substituted in for `T` except that it does not include `null`
```
netdpb, line 144:
```suggestion
>         type argument substituted in for `T` except that it does not include `null`
```

docs/docs/spec.md
```diff
@@ -392,7 +386,9 @@ condition is met, skip the remaining conditions.
 
 > If tool authors prefer, they can safely produce `MINUS_NULL` in any case in
 > which it is equivalent to `NO_CHANGE`. For example, there is no difference
-> between `String NO_CHANGE` and `String MINUS_NULL`.
+> between `Foo NO_CHANGE` and `Foo MINUS_NULL` for any class type `Foo` (nor for
+> any array type or the null type). The difference *is* significant for
+> intersection types, type variables, and probably union types.
```
netdpb, line 391: "Probably"? 😆

netdpb (approved):

docs/docs/spec.md
cpovirk: Done.

docs/docs/spec.md
cpovirk: I mean, it's not like I'd thought about union types at all until Friday :) But hey, it's a non-normative comment, so let's be bold!

cpovirk (commented):
Oops, I tried to do the thing where I avoid generating lots of emails, and I succeeded only in how I pushed my changes, not in how I responded to comments.

Anyway, I'll deal with conflicts and then merge this.

PR #620, opened 2024-09-05
Update the spec for `@NullUnmarked`.

cpovirk:
Sneak in some discussion of how we define "enclosing" perhaps somewhat
differently from how the Java compiler API does.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -329,7 +329,7 @@ locations listed below:
 
 -   A *named* class.
 -   A package. \[[#34]\]
--   A module. \[[#34]\]
+-   A module (for `@NullMarked` only, not `@NullUnmarked`). \[[#34]\]
 
 > *Not* a method \[[#43]\], constructor \[[#43]\], or field \[[#50]\].
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -344,9 +344,9 @@ innermost.
 
 -   Each class member is enclosed by a class.
 -   Each non-top-level class is enclosed by a class or class member.
--   Each top-level class is enclosed by a package.
+-   Each top-level class is enclosed by its package.
 -   Each package may be enclosed by a module.
--   Each module is enclosed by nothing.
+-   Modules are not enclosed by anything.
 
 > Packages are *not* enclosed by "parent" packages.
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -319,7 +319,7 @@ All locations that are not explicitly listed as recognized are unrecognized.
 
 ## The declaration annotations
 
-We provide two parameterless declaration annotations, `@NullMarked` and
+We provide two parameterless declaration annotations: `@NullMarked` and
 `@NullUnmarked`.
 
 ### Recognized locations for declaration annotations
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -342,10 +342,11 @@ innermost.
 
 "Enclosing" is defined as follows:
 
--   Class members are enclosed by classes.
--   Classes may be enclosed by other class members or classes.
--   Top-level classes are enclosed by packages.
--   Packages may be enclosed by modules.
+-   Each class member is enclosed by a class.
+-   Each non-top-level class is enclosed by a class or class member.
+-   Each top-level class is enclosed by a package.
+-   Each package may be enclosed by a module.
+-   Each module is enclosed by nothing.
 
 > Packages are *not* enclosed by "parent" packages.
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -340,9 +340,12 @@ To determine whether a type usage appears in a null-marked scope:
 Iterate over all the declarations that enclose the type usage, starting from the
 innermost.
 
-"Enclosing" is defined as follows: Class members are enclosed by classes, which
-may be enclosed by other class members or classes. Top-level classes are
-enclosed by packages, which may be enclosed by modules.
+"Enclosing" is defined as follows:
+
+-   Class members are enclosed by classes.
+-   Classes may be enclosed by other class members or classes.
+-   Top-level classes are enclosed by packages.
+-   Packages may be enclosed by modules.
 
 > Packages are *not* enclosed by "parent" packages.
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -360,8 +360,8 @@ At each declaration, check two rules:
 -   If the declaration is annotated with `@NullUnmarked` and *not* with
     `@NullMarked`, the type usage is *not* in a null-marked scope.
 
-> If a given declaration is annotated with both `@NullMarked` and `@NullUnmarked`, these                                                                                                                                   
-> rules behave as if neither annotation is present.                                                                                                                                                            
+> If a given declaration is annotated with both `@NullMarked` and
+> `@NullUnmarked`, these rules behave as if neither annotation is present.
 
 If none of the enclosing declarations meet either rule, then the type usage is
 *not* in a null-marked scope.
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -324,6 +324,9 @@ locations listed below:
 
 > *Not* a field.
 
+If our declaration annotations appear in any other location, they have no
+meaning.
+
 ## Null-marked scope
 
 To determine whether a type usage appears in a null-marked scope:
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -317,14 +317,14 @@ All locations that are not explicitly listed as recognized are unrecognized.
 > But note that types "inside" some of these locations can still be recognized,
 > such as a *type argument* of a supertype.
 
-## The declaration annotation
+## The declaration annotations
 
-We provide a single parameterless declaration annotation called `@NullMarked`.
-\[[#5], [#87]\]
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -317,14 +317,14 @@ All locations that are not explicitly listed as recognized are unrecognized.
 > But note that types "inside" some of these locations can still be recognized,
 > such as a *type argument* of a supertype.
 
-## The declaration annotation
+## The declaration annotations
 
-We provide a single parameterless declaration annotation called `@NullMarked`.
-\[[#5], [#87]\]
+We provide two parameterless declaration annotations, `@NullMarked` and
```
netdpb, line 322:
```suggestion
We provide two parameterless declaration annotations: `@NullMarked` and
```

docs/docs/spec.md
```diff
@@ -317,14 +317,14 @@ All locations that are not explicitly listed as recognized are unrecognized.
...
+`@NullUnmarked`.
 
 ### Recognized locations for declaration annotations
 
-Our declaration annotation is specified to be *recognized* when applied to the
+Our declaration annotations are specified to be *recognized* when applied to the
```
netdpb, line 327: `@NullUnmarked` isn't recognized at the module level, is it?

docs/docs/spec.md
```diff
@@ -337,21 +337,31 @@ locations listed below:
...
 
-Look for a `@NullMarked` annotation on any of the scopes
-enclosing the type usage.
+Iterate over all the declarations that enclose the type usage, starting from the
+innermost.
 
-Class members are enclosed by classes, which may be enclosed by other class
-members or classes. and top-level classes are enclosed by packages, which may be
-enclosed by modules.
+"Enclosing" is defined as follows: Class members are enclosed by classes, which
```
netdpb, line 343: Perhaps bullets would be better here?

docs/docs/spec.md
```diff
@@ -337,21 +337,31 @@ locations listed below:
...
 
 > Packages are *not* enclosed by "parent" packages.
 
-> This definition of "enclosing" likely matches
-> [the definition in the Java compiler API](https://docs.oracle.com/en/java/javase/14/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement\(\)).
+> This definition of "enclosing" largely matches
+> [the definition in the Java compiler API](https://docs.oracle.com/en/java/javase/22/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement\(\)).
+> The JSpecify definition differs slightly by skipping type-parameter
+> declarations (which cannot be annotated with declaration annotations) and by
+> defining that there exists a series of enclosing declarations for any type
```
netdpb, line 353: Really? The definition talks about class members, classes, packages, and modules. It doesn't actually say anything about parameters or local variables or cast expressions. Should it?

docs/docs/spec.md
```diff
@@ -337,21 +337,31 @@ locations listed below:
...
-`@NullMarked`, then the type usage is in a null-marked
-scope. Otherwise, it is not.
+At each declaration, check two rules:
+
+-   If the declaration is annotated with `@NullMarked` and *not* with
+    `@NullUnmarked`, the type usage is in a null-marked scope.
+-   If the declaration is annotated with `@NullUnmarked` and *not* with
+    `@NullMarked`, the type usage is *not* in a null-marked scope.
+
+If none of the enclosing declarations meet either rule, then the type usage is
```
netdpb, line 363: Should we point out, non-normatively, that a declaration has both annotations then it is as if it had neither?

docs/docs/spec.md
cpovirk: ...or record components or supertypes or `instanceof` or class-creation expressions or array-creation expressions or really even that, say, a return type is enclosing by its method or a type argument _inside_ the return type is enclosed by the method...

So I can't even write this one off to our attempt to ignore implementation code.

I think that you can get the right result by walking up the javac syntax tree, but that's not a great answer, even assuming that it is true today.

I've filed https://github.com/jspecify/jspecify/issues/622.

docs/docs/spec.md
```diff
@@ -329,7 +329,7 @@ locations listed below:
 
 -   A *named* class.
 -   A package. \[[#34]\]
--   A module. \[[#34]\]
+-   A module (for `@NullMarked` only, not `@NullUnmarked`). \[[#34]\]
```
netdpb, line 332: Should we remove some of these issue links?

docs/docs/spec.md
```diff
@@ -340,9 +340,12 @@ To determine whether a type usage appears in a null-marked scope:
 Iterate over all the declarations that enclose the type usage, starting from the
 innermost.
 
-"Enclosing" is defined as follows: Class members are enclosed by classes, which
-may be enclosed by other class members or classes. Top-level classes are
-enclosed by packages, which may be enclosed by modules.
+"Enclosing" is defined as follows:
+
+-   Class members are enclosed by classes.
+-   Classes may be enclosed by other class members or classes.
```
netdpb, line 346: "other class members" refers to local and anonymous classes, right?

docs/docs/spec.md
```diff
@@ -340,9 +340,12 @@ To determine whether a type usage appears in a null-marked scope:
 Iterate over all the declarations that enclose the type usage, starting from the
 innermost.
 
-"Enclosing" is defined as follows: Class members are enclosed by classes, which
-may be enclosed by other class members or classes. Top-level classes are
-enclosed by packages, which may be enclosed by modules.
+"Enclosing" is defined as follows:
+
+-   Class members are enclosed by classes.
```
netdpb, line 345: Should we make these singular to avoid ambiguity about how many (directly) enclosing things there are?

"Each class member is enclosed by a class", etc.

docs/docs/spec.md
```diff
@@ -340,9 +340,12 @@ To determine whether a type usage appears in a null-marked scope:
...
+-   Classes may be enclosed by other class members or classes.
+-   Top-level classes are enclosed by packages.
+-   Packages may be enclosed by modules.
 
 > Packages are *not* enclosed by "parent" packages.
```
netdpb, line 350: Nor are modules enclosed by parent modules, right?

netdpb (approved):

docs/docs/spec.md
cpovirk: Yes, that sounds good to me.

docs/docs/spec.md
cpovirk: I think this text predates my involvement, but I was figuring that it meant "methods or constructors," with the cases of local and anonymous classes covered by "or classes."

Does that makes sense? I think this definition could be tightened up in general, but I think tools can largely just lean on the AST representation and get the right behavior as a result. For our immediate purposes, I want to make sure I'm not missing anything big.

docs/docs/spec.md
cpovirk: True. I don't know that I've heard people who want it to be otherwise (partially because module usage is still somewhat low and "parent" module naming is probably rare), so I'm less worried about that. But we should probably mention in general that modules aren't enclosed by anything, so I've added that.

cpovirk (commented):
Thanks. I'd appreciate one more look in light of the changes to "enclosing."

docs/docs/spec.md
netdpb: Sorry. I meant to ask whether the only kinds of classes that are enclosed by class members are local and anonymous ones.

(Of course even that is ambiguous because a nested class is also a class member.)

cpovirk: Ah. Yes, I believe so, with the caveat you give about a class nested in an already nested class.

docs/docs/spec.md
```diff
@@ -342,10 +342,11 @@ innermost.
 
 "Enclosing" is defined as follows:
 
--   Class members are enclosed by classes.
--   Classes may be enclosed by other class members or classes.
--   Top-level classes are enclosed by packages.
--   Packages may be enclosed by modules.
+-   Each class member is enclosed by a class.
+-   Each non-top-level class is enclosed by a class or class member.
```
netdpb, line 346: What about this:

* Each local or anonymous class is enclosed by the class member in which it is declared.
* Each class member (including member classes) is enclosed by the class in which it is declared.
* Each top-level class is enclosed by its package.
* Each package may be enclosed by a module.
* Modules are not enclosed by anything.

[By the way, how are unexported and unopened packages enclosed by a module? I can't figure it out from the JLS.]

netdpb (approved):

docs/docs/spec.md
cpovirk: Anonymous (and perhaps local) classes aren't necessarily enclosed by a class member:

```
class Foo {
  Object o = new Object() {};
}
```

(And a local class inside a `static {}` block probably isn't enclosed by a "member" in normal JLS terms, though I'd have to check that.)

We might be able to eliminate some overlap between the "class member" and "class" cases by changing "Each class member is enclosed by a class" to "Each method or constructor is enclosed by a class." But it's probably just as well to wait until we can [rework this section more thoroughly](https://github.com/jspecify/jspecify/issues/622).

As for unexported and unopened packages: I would expect that to work like exported packages (though there might be restrictions on what annotation processors can see; I don't know). The JLS probably doesn't need to get into it, since it probably doesn't define "enclosing," so our best reference is probably still [`Element.getEnclosingElement()`](https://docs.oracle.com/en/java/javase/22/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement()).

PR #632, opened 2024-09-10
Improve discussion of subtyping and same-type rules.

cpovirk:
This PR addresses two comments from @netdpb:

- We now cite a JLS concept of "same type."

- For `A` to a JSpecify subtype of `F`, we now require that the
  corresponding *base* types pass the Java subtyping test.

Also, it sneaks in some real em dashes, which I'm guessing got converted
to `---` at some point under the assumption that that would be rewritten
by doc tooling to an em dash... after which I guess that assumption got
violated?

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -622,21 +622,23 @@ The same-type relation is *not* defined to be reflexive or transitive.
 
 `A` is a subtype of `F` if all of the following conditions are met:
 
+-   The base type of `A` is a subtype of the base type of `F` according to
+    Java's subtyping rules.
 -   `A` is a [nullness subtype] of `F`.
 -   `A` is a subtype of `F` according to the
     [nullness-delegating subtyping rules for Java].
--   The base type of `A` is a subtype of the base type of `F` according to
-    Java's subtyping rules.
 
-> The first condition suffices for most cases.
->
-> The second condition is necessary only for types that have
-> subcomponents—namely, parameterized types and arrays. And it essentially says
-> "Check the first condition on subcomponents as appropriate."
->
-> The third condition is rarely of direct interest to tools' JSpecify support:
+> The first condition is rarely of direct interest to tools' JSpecify support:
 > If code does not pass Java's subtyping rules, then most JSpecify tools will
 > never see it.
+>
+> As for the remaining conditions:
+>
+> The nullness-subtype condition suffices for most cases.
+>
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -615,18 +615,28 @@ The same-type relation is *not* defined to be reflexive or transitive.
...
+
 ## Subtyping
 
-`A` is a subtype of `F` if both of the following conditions are met:
+`A` is a subtype of `F` if all of the following conditions are met:
 
 -   `A` is a [nullness subtype] of `F`.
 -   `A` is a subtype of `F` according to the
     [nullness-delegating subtyping rules for Java].
+-   The base type of `A` is a subtype of the base type of `F` according to
```
netdpb, line 628: I feel like this one should come first. I guess the order doesn't technically matter, but this one will usually get applied first, as you imply.

Is there any opportunity to simplify the nullness-delegating subtyping rules for Java now that it's explicit that the regular subtyping rules for Java have to apply as well?

netdpb (approved):

docs/docs/spec.md
cpovirk: Sure, I've moved it first.

As for the nullness-delegating subtyping rules for Java: I think I already wrote those under the implicit assumption that we could assume that base-type subtyping held, so I don't see opportunity for simplification.

I do hope to simplify those rules someday, but it is likely to be a big project. I've filed https://github.com/jspecify/jspecify/issues/638 to at least track that beyond the "we should fix it" note in the spec.

docs/docs/spec.md
```diff
@@ -622,21 +622,23 @@ The same-type relation is *not* defined to be reflexive or transitive.
...
-    Java's subtyping rules.
 
-> The first condition suffices for most cases.
->
-> The second condition is necessary only for types that have
-> subcomponents—namely, parameterized types and arrays. And it essentially says
-> "Check the first condition on subcomponents as appropriate."
->
-> The third condition is rarely of direct interest to tools' JSpecify support:
+> The first condition is rarely of direct interest to tools' JSpecify support:
```
netdpb, line 631: Should these be inline (after each condition)?

netdpb (approved):

docs/docs/spec.md
cpovirk: Yeah, that sounds better. Done.

netdpb (approved):

Issue #522, opened 2024-06-10
Fix substitution rules for `NO_CHANGE`+`MINUS_NULL`


cpovirk:
As part of discussing spec changes for https://github.com/jspecify/jspecify/issues/248, @netdpb found that we have a bug in how we currently handle `NO_CHANGE`+`MINUS_NULL`. We'll want to update https://jspecify.dev/docs/spec/#applying-operator to fix that.

I notice that we already have some handling for that case in https://jspecify.dev/docs/spec/#substitution (and I think in the reference checker), but the spec (though perhaps _not_ the checker?) is written in a way that I'm not sure it would apply to type-variable usages in intersection types, maybe? I'd have to look.

cpovirk:
If #248 goes the way that I've been hoping, then I'll fix this as part of the general simplification of the substitution rules. You can see [a preview of that in a WIP commit in my fork](https://github.com/cpovirk/jspecify/commit/be78dac44fec068b572bf960cd46e664ba639041#diff-e4d0d9d06bacf40ef44c340edf37d7859839b5a409a67867028e2bacc37721c1R945).

cpovirk:
(closed by https://github.com/jspecify/jspecify/commit/d9ee436c9d2ddb79002f820da9f0f1496e80f608)

Issue #623, opened 2024-09-06
Update spec for any to-be-finalized issues: #248, #618, #158


cpovirk:
Those might be the only ones with a decent chance to change at this point? Anyway, I'm filing this meta-issue to add it to our 1.0 tracking. We may replace it with individual issues if we find ourselves with a lot of changes to make.

- #248 
- #618 
- #158

cpovirk:
I [just realized](https://github.com/jspecify/jspecify/pull/619/files#r1752351381) that I accidentally put the specification-1.0 branch into a state of being inconsistent about the answer to #248. So no matter what we decide about #248, we'll still need to make some spec change.

cpovirk:
Update:
- [#248](https://github.com/jspecify/jspecify/issues/248): We changed the rule as expected, and I updated the spec.
- [#618](https://github.com/jspecify/jspecify/issues/618): It looks like we may keep the current rule, in which case there would be nothing to do. I'll check on this in the meeting today.
- [#158](https://github.com/jspecify/jspecify/issues/158): I claimed that [there is nothing _in the spec_ specifically about this](https://github.com/jspecify/jspecify/issues/158#issuecomment-2349319331), so it can wait until we are ready to get back into the conformance tests more deeply.

So: If no one objects to the current proposal for [#618](https://github.com/jspecify/jspecify/issues/618) today, I'll be able to close this issue (and that one) as completed.

cpovirk:
Tentatively closing, since https://github.com/jspecify/jspecify/issues/618 is probably going to stick. (I'll keep that one open, though, for a little longer, _just in case_. We can reopen this one then if needed.)

PR #633, opened 2024-09-10
Update rules for a change in `T*` substitution.

cpovirk:
This PR implements the "`*/*/*/*`" (aka "`String*`") option from
https://github.com/jspecify/jspecify/issues/248.

As such, we should merge it only if we decide to go that way. If we
decide to go another way, then we probably need to [make a different
change to the the
spec](https://github.com/jspecify/jspecify/issues/623#issuecomment-2341517812):
We need to do _something_ for #248 as part of
https://github.com/jspecify/jspecify/issues/623.

Assuming that we do go that way:

I think this PR changes the rules such that we never say either "If `T`
has nullness operator `UNION_NULL`..." or "If `T` has nullness operator
`MINUS_NULL`...." Instead, checks for those nullness operators always
use the "worried" and "comfortable" rules. This is part of what I
conceptually like about both these new rules and the worried/comfortable
model, though it didn't fully sink in until I was writing this commit
message.

As part of simplifying the rules of substitution, this PR also...

Fixes https://github.com/jspecify/jspecify/issues/522

If we do not merge this PR, then we'll need to merge an alternative fix.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -672,10 +672,24 @@ The same-type relation is *not* defined to be reflexive or transitive.
     > `ArrayList` may be instantiated as either an `ArrayList<@Nullable String>`
     > or an `ArrayList<String>`.)
     >
-    > Subtyping questions for type-variable usages are more complex: `E` is a
-    > nullness subtype of `E`; `@Nullable E` is not. Similarly, if `<F extends
-    > E>`, then `F` is a nullness subtype of `E`. But if `<F extends @Nullable
-    > E>`, it is not.
+    > Subtyping questions for type-variable usages are more complex. For
+    > example:
+    >
+    > -   `E` is a nullness subtype of `E`; `@Nullable E` is not.
+    > -   Similarly, if `<F extends E>` (in null-marked code), then `F` is a
+    >     nullness subtype of `E`. But if `<F extends @Nullable E>`, it is not.
+    > -   `E` is a nullness subtype of `E` but not of `@NonNull E`.
+    >
+    > When some types have unspecified nullness, the rules become more complex
+    > still:
+    >
+    > -   A declaration like `<F extends E>` might or might not be "intended" to
+    >     be `<F extends @Nullable E>`. Depending on what was indended, `E`
+    >     *might* be intended to be a nullness subtype of `F`.
+    > -   Or that declaration might be "intended" to be `<F extends @NonNull
+    >     E>`. In that [world], `F` would be not only a nullness subtype of `F`
+    >     but a nullness subtype of *all* types, since it would be
+    >     null-exclusive under every parameterization.
 
 -   `F` is a type-variable usage that meets *both* of the following conditions:
 
@@ -1092,5 +1106,6 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 [subtyping]: #subtyping
 [type component]: #type-components
 [type components]: #type-components
+[world]: #multiple-worlds
 [worlds]: #multiple-worlds
 [worried]: #worried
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -649,8 +649,8 @@ The same-type relation is *not* defined to be reflexive or transitive.
     > This is the second easy case: `A` never includes `null`.
 
 -   `A` has a [nullness-subtype-establishing path] to any type whose base type
-    is the same as the base type of `F`, and `F` does *not* have
-    [nullness operator] `MINUS_NULL`.
+    is the same as the base type of `F`, and there is *not* reason to be
+    [worried] that `F` has [nullness operator] `MINUS_NULL`.
```
netdpb, line 653: Should we add an unspecified type example to the non-normative example below?

docs/docs/spec.md
```diff
@@ -968,33 +971,19 @@ The process of applying a [nullness operator] requires two inputs:
...
+-   If the nullness operator to apply is `NO_CHANGE`, then the output augmented
+    type is the input augmented type.
+
+-   If the input augmented type is an [intersection type], then the output is
+    also an intersection type. For every element `Tᵢ` of the input type, the
+    output type has an element that is the output of applying the desired
+    nullness operator to `Tᵢ`.
+
+-   The output is a type that is the same as the input augmented type except
+    with its nullness operator set to the nullness operator to apply.
```
netdpb, line 986:
```suggestion
To determine the output, apply the following rules in order.

-   If the nullness operator to apply is `NO_CHANGE`, then the output augmented
    type is the input augmented type.

-   Otherwise, if the input augmented type is an [intersection type], then the output is
    also an intersection type. For every element `Tᵢ` of the input type, the
    output type has an element that is the output of applying the desired
    nullness operator to `Tᵢ`.

-   Otherwise, the output is a type that is the same as the input augmented type except
    with its nullness operator set to the nullness operator to apply.
```

Is this more direct?

netdpb (approved):

docs/docs/spec.md
cpovirk: Probably. We have a couple other "skip the remaining conditions" sections that you're welcome to touch up if you'd like.

docs/docs/spec.md
cpovirk: I see that we don't even have a `MINUS_NULL` example right now. I've added one and then said some things about unspecified nullness. I'm not sure if I picked good things to say or not.

docs/docs/spec.md
```diff
@@ -658,10 +658,24 @@ The same-type relation is *not* defined to be reflexive or transitive.
...
+    > -   Similarly, if `<F extends E>` (in null-marked code), then `F` is a
+    >     nullness subtype of `E`. But if `<F extends @Nullable E>`, it is not.
+    > -   `E` is a nullness subtype of `E` but not of `@NonNull E`.
+    >
+    > When some types have unspecified nullness, the rules become more complex
+    > still:
+    >
+    > -   A declaration like `<F extends E>` might or might not be "intended" to
+    >     be `<F extends @Nullable E>`. Depending on what was indended, `E`
+    >     *might* be intended to be a nullness subtype of `F`.
```
netdpb, line 674: I _think_ you meant this?

```suggestion
    >     be `<F extends @Nullable E>`. Depending on what was indended, `F`
    >     *might* be intended to be a nullness subtype of `E`.
```

docs/docs/spec.md
```diff
@@ -658,10 +658,24 @@ The same-type relation is *not* defined to be reflexive or transitive.
...
+    > -   Or that declaration might be "intended" to be `<F extends @NonNull
+    >     E>`. In that [world], `F` would be not only a nullness subtype of `F`
```
netdpb, line 676:
```suggestion
    >     E>`. In that [world], `F` would be not only a nullness subtype of `E`
```

netdpb (approved):

cpovirk:
Thanks. I will now try my best to remember not to merge this until we've actually made a decision :)

cpovirk:
Gah, I didn't think I was clicking the "close" button....

PR #634, opened 2024-09-10
Clarify all-worlds interpretation a bit

netdpb:
Clarify that when a type is both null-inclusive under every parameterization and null-exclusive under every parameterization in an all-worlds interpretation, those are each true *in different worlds*.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -683,7 +683,7 @@ The same-type relation is *not* defined to be reflexive or transitive.
 > A further level of complexity in all this is `UNSPECIFIED`. For example, in
 > the [all-worlds] version of the following rules, a type with nullness operator
 > `UNSPECIFIED` can be both null-_inclusive_ under every parameterization and
-> null-_exclusive_ under every parameterization.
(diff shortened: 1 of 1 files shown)


cpovirk (approved):
Thanks.

I suspect you are already aware of https://github.com/jspecify/jspecify/pull/603, which will fix line 684. If not, FYI, since that will also help with this section. (I think it's best to make both changes.)

netdpb:
Wow, I totally missed that the paragraph is talking about _all worlds_. Is that a simple mistake in the non-normative text or does it represent a problem with the rule?

netdpb:
And I can contribute this change to that PR instead of making a new one.

cpovirk:
Just a mistake in the non-normative text.

I'm fine with either separate PRs or a combined PR.

netdpb:
Ugh. I tried to create a PR on your PR but mine is based on specification-1.0 and yours on main, so ugh. I'll just fix the non-normative mistake here.

cpovirk (approved):

PR #636, opened 2024-09-10
Apply a Markdown formatter.

cpovirk:
(The formatter I have handy is a full-file formatter, so it's nice to be
able to run it on "real" PRs without accidentally including changes like
this in the resulting commit.)

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -221,8 +221,8 @@ exceptions in the subsequent sections:
 
 -   the return type of a method
 
--   a formal parameter type of a method or constructor, as defined in [JLS
-    8.4.1]
+-   a formal parameter type of a method or constructor, as defined in
+    [JLS 8.4.1]
 
     > This excludes the receiver parameter but includes variadic parameters.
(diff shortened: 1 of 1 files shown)


netdpb (approved):
This is fine, but it's not what `mdformat --compatibility` does (which is what I've been using).

What are you using? Can we standardize?

cpovirk:
Oh, I hadn't realized. I have been running `mdformat` but _without_ `--compatibility`. But: It looks like this change is made whether I pass the flag or not?

(You got my hopes up with `--compatibility` would avoid mangling the header :) I have been meaning to write a quick script to format everything _except_ the header....)

cpovirk:
For reference, I am currently going with:

```
perl -ne 'print if (/----/ .. 0)' docs/docs/spec.md > /tmp/toformat.md && mdformat --in_place --compatibility /tmp/toformat.md && ( perl -ne 'print if (1 .. /----/) && !/----/' docs/docs/spec.md && cat /tmp/toformat.md ) > /tmp/formatted.md && cat /tmp/formatted.md > docs/docs/spec.md
```

Issue #611, opened 2024-08-27
Clarify in the spec and docs how `@NullMarked` at a package or module level applies to Kotlin-declared symbols within that container


netdpb:
As of now, Kotlin emits classes that don't contain any JSpecify annotations but whose types are null-augmented. At some future date, Kotlin may emit JSpecify nullness annotations that represent the null-augmented types.

If a package or module is annotated with `@NullMarked`, how should tools interpret the types of Kotlin-declared symbols in that package or module? (Of course, tools may and probably should interpret Kotlin's native meta-annotation that describes the Kotlin-defined nullness of those symbols' types; this is just about what JSpecify's specification says.)

One idea is to consider Kotlin-declared symbols as null-unmarked regardless of any package or module annotation. So Kotlin classes would be null-unmarked unless annotated with `@NullMarked` or contained within a `@NullMarked` (Kotlin) class. That would allow the documentation to remain unchanged if and when Kotlin starts emitting JSpecify annotations.

netdpb:
This was fixed by #637.

PR #637, opened 2024-09-10
Treat Kotlin code as null-unmarked.

cpovirk:
This rule is written such that a Kotlin class _would_ still be treated
as null-marked _if_ it (or an enclosing class) were annotated with
`@NullMarked`. That enables a user to annotate it if desired, but more
importantly, it allows a future version of Kotlin to produce that
annotation automatically.

Fixes https://github.com/jspecify/jspecify/issues/611

~This rule is written such that a Kotlin class _would_ still be treated
as null-marked _if_ it were annotated with `@NullMarked`. That enables a
user to annotate it if desired, but more importantly, it allows a future
version of Kotlin to produce that annotation automatically.~

~Still, we are likely to want to revisit this rule, especially if Kotlin
begins generating `@NullMarked`. In particular, as [noted by
netdpb@](https://github.com/jspecify/jspecify/issues/611#issue-2489956831),
if a Kotlin class is nested inside a `@NullMarked` Kotlin class, then
the nested class should probably also be treated as null-marked. Under
this current rule, it probably isn't, since we would (IIUC) see a
`@kotlin.Metadata` annotation on the nested class before seeing the
`@NullMarked` annotation on the enclosing class.~

~We could try to do better, but this PR's approach has the virtue of
being very easy to implement.~

~(So this PR might not really "fix"
https://github.com/jspecify/jspecify/issues/611, but I propose that it's
good enough for spec-1.0.0 purposes. We could probably even release spec
1.0.0 without any PR like this if we really wanted.)~

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -352,7 +352,7 @@ At each declaration, check the following rules in order:
     `@NullUnmarked`, the type usage is in a null-marked scope.
 -   If the declaration is annotated with `@NullUnmarked` and *not* with
     `@NullMarked`, the type usage is *not* in a null-marked scope.
--   If the declaration is annotated with `@kotlin.Metadata`, then the type
+-   If the declaration is a top-level class annotated with `@kotlin.Metadata`, then the type
     usage is *not* in a null-marked scope.
 
 > If a given declaration is annotated with both `@NullMarked` and
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -352,8 +352,8 @@ At each declaration, check the following rules in order:
     `@NullUnmarked`, the type usage is in a null-marked scope.
 -   If the declaration is annotated with `@NullUnmarked` and *not* with
     `@NullMarked`, the type usage is *not* in a null-marked scope.
--   If the declaration is a top-level class annotated with `@kotlin.Metadata`, then the type
-    usage is *not* in a null-marked scope.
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -346,12 +346,14 @@ innermost.
...
 
-At each declaration, check two rules:
+At each declaration, check the following rules in order:
 
 -   If the declaration is annotated with `@NullMarked` and *not* with
     `@NullUnmarked`, the type usage is in a null-marked scope.
 -   If the declaration is annotated with `@NullUnmarked` and *not* with
     `@NullMarked`, the type usage is *not* in a null-marked scope.
+-   If the declaration is annotated with `@kotlin.Metadata`, then the type
+    usage is *not* in a null-marked scope.
```
netdpb, line 356: Would "doing better" be as simple as this?

```suggestion
-   If the declaration is a top-level class annotated with `@kotlin.Metadata`, then the type
    usage is *not* in a null-marked scope.
```

netdpb: I'm presuming that the generated `FilenameKt` class that encloses Kotlin top-level functions and properties has a `kotlin.Metadata` annotation.

netdpb (approved):

docs/docs/spec.md
cpovirk: Good point. Yes, it looks like it does in the test I just ran.

Issue #626, opened 2024-09-09
Spec should forbid annotations anywhere within `instanceof` types


cpovirk:
(assigning to @netdpb in case he wants to include this in https://github.com/jspecify/jspecify/pull/592)

javac accepts code like this:

```java
import org.jspecify.annotations.Nullable;

class InstanceOfAnnotated {
  void go(Object o) {
    if (o instanceof Foo<? extends @Nullable Number>) {}
  }

  interface Foo<T extends Number> {}
}
```

We already ban annotations on the _root_ type of the `instanceof` types. We should also ban them on any subcomponent types:

> [INSTANCEOF -- we should not permit anything that a user could think was being checked but wasn't.](https://github.com/jspecify/jspecify/issues/17#issuecomment-484997275)

This will be a way in which casts differ from `instanceof`.

agentgt:
Another oddity is that you can do this:

```java
String a = switch(o) {
  case @Nullable String l -> "stuff";
  case null -> "";
};
```

Which is fine for `javac` (as it has no idea what `@Nullable` is) but does not make sense with JSpecify.

I'm not sure if that is covered anywhere else.

cpovirk:
Thanks, I think we had kind of resigned ourselves to sticking with the language features of Java 11ish, but we should at minimum track that for later. My hope would be that `case` patterns are similar to `instanceof` patterns. Does that seem believable?

Hmm, and that gets me to think about them in more detail. I guess that it's a little sad that we would be forbidding:

```java
Collection<? extends @Nullable String> collection = ...;
if (collection instanceof List<? extends @Nullable String> list) { ... }
```

It's especially sad that that's actually _worse_ than omitting the annotation, since omitting the annotation makes the type more specific, and the specificity is exactly what we're trying to communicate is not checked:

```java
Collection<? extends @Nullable String> collection = ...;
if (collection instanceof List<? extends String> list) { ... } // worse!
```

Maybe we "need" to allow the annotations (and then make checkers responsible for making sure that they match?)? Otherwise, users have to use plain `instanceof List`, which can't be used with a pattern-style variable declaration (unless you want the raw type, yuck), IIRC.

So this needs more thought, and it is not a good candidate for sneaking into #592.

(Well, we could still consider forbidding them in spec 1.0 and then being more permissive later.)

agentgt:
For sealed classes I think we forbid doing something like this right?

```java
sealed interface Parent<T extends @Nullable Object> {
[...]

PR #644, opened 2024-09-18
Deem annotations in `instanceof` to be unrecognized, even in type arguments.

cpovirk:
This addresses https://github.com/jspecify/jspecify/issues/626 enough
for 1.0 purposes, but as I note in a non-normative comment in the spec,
we may want to revisit it in the future. I'll keep the bug open but
remove it from the spec-1.0 milestone.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -247,7 +247,7 @@ However, the type-use annotation is unrecognized in any of the following cases:
 
 -   type arguments of a receiver parameter's type
 
--   any component of an `instanceof` expression
+-   any component of the type after the `instanceof` operator
 
     > We may revisit this rule in the future in light of
     > [Pattern Matching for instanceof][JEP 394].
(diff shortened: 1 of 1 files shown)


cpovirk (commented):
I should probably remove "root type in a cast of `instanceof` expression" from the list of "other" locations below.

cpovirk:
I notice that [the Javadoc](https://jspecify.dev/docs/api/org/jspecify/annotations/Nullable.html) currently mentions "the argument to `instanceof`." That could arguably be meant to refer to the whole thing—except that it comes in a bullet about "intrinsically non-null type usage" that goes on to say: "In such locations, a nullness annotation could only be contradictory (`@Nullable`) or redundant (`@NonNull`)." So it's pretty clear that we're referring only to the root type.

(My fresh look at the Javadoc just made me realize that I should say something in the spec about "a method return type in an annotation interface" :) The rule for that would apply to the whole type (to cover array and class elements). I don't think we need to get into that for the Javadoc, but I'll try to cover it in the spec.)

Issue #301, opened 2022-10-07
Some parts of bytecode may need to be considered always null-unmarked: synthetic methods, local variables (because of `var`), anything else?


cpovirk:
The nullness annotations for those either aren't known to javac at all or aren't propagated by it into the places we'd want in the bytecode.

This isn't a catastrophe: The typical use for nullness information is to perform checking during compilation. However, it does mean that any tools that rewrite bytecode (such as to insert runtime null checks) will have to be careful about what they assume. Presumably we'd want to document that somewhere, though only in developer-facing docs—definitely not in, e.g., the Javadoc.

### `synthetic` methods

```java
import org.jspecify.nullness.NullMarked;
import org.jspecify.nullness.Nullable;

@NullMarked
class SyntheticDemo {
  java.util.function.Consumer<@Nullable Object> c = o -> System.out.println(o);
}
```

javac outputs the annotation that is part of the `c` field's type, as expected. But when it generates a method to "back" the lambda, it doesn't include an annotation on the method parameter. It couldn't: Even though a nullness analyzer will infer that `o` has the type `@Nullable Object`, we obviously can't rely on stock javac to do that. (And even when an analyzer is running, we don't generally expect it to encode annotations into the bytecode beyond what stock javac does.)

I suppose it's possible that _some_ synthetic methods have correct JSpecify nullness information. For example, maybe it will happen in bytecode generated by kotlinc once it writes exhaustive JSpecify annotations. We could theoretically provide a way for those methods to express that they _are_ annotated, but I doubt that's worth the trouble. (For example, I wouldn't want that code to put a redundant `@NullMarked` annotation on its `synthetic` methods as a way of saying "This method, which was already enclosed by a class-level `@NullMarked` annotation, is _really_ null-marked." Not to mention that I wouldn't be shocked if some part of javac somewhere copied a `@NullMarked` annotation from one hand-written method to a derived `synthetic` method automatically, ruining our fun :) Not that I've checked.)

Another case of `synthetic` methods is bridge methods.

(IIRC, default constructors are _not_ flagged as `synthetic`, but because they have no parameters, that doesn't hurt us. OK, fine, they can have an implicit outer-instance parameter, but I'm not losing sleep over that one.)

### `var`

```java
import java.util.HashSet;
import org.jspecify.nullness.NullMarked;
import org.jspecify.nullness.Nullable;

class VarDemo {
  void explicit() {
    HashSet<@Nullable String> s = new HashSet<@Nullable String>();
  }

[...]

Issue #618, opened 2024-09-04
Revisit substitution for a type parameter `<T extends @NonNull Object>`


cpovirk:
https://jspecify.dev/docs/spec/#substitution discusses our current handling of this situation, handling that which we sometimes refer to as "the `ImmutableList.Builder` rule." Discussions of that go [way back](https://github.com/jspecify/jspecify/issues/91#issuecomment-581580007), and even today we have have [an issue open for a small detail of the current handling](https://github.com/jspecify/jspecify/issues/521).

@wmdietl was concerned that (if I'm summarizing correctly) a user who writes `NullHostileList<@Nullable String>` might want to receive an error not only when writing the type "`NullHostileList<@Nullable String>`" and when inserting nulls into the list but also when dereferencing an element retrieved from the list: If `NullHostileList` is written without runtime checks, then a null value might make it the whole way through to that point.

We should write up a doc on this before spec 1.0.

cpovirk:
(If we keep the current rule, then the spec should provide better justification (maybe inline but probably by linking to whatever we write). The justification should touch on not just "Why do we want a special case here?" but also "Why do we implement the special case as a special case _in the substitution rules_, rather than through a rule that says that `ImmutableList<@Nullable String>` gets silently reinterpreted to `ImmutableList<@NonNull String>`. That question has become a FAQ among spec enthusiasts.)

cpovirk:
(Some related discussion of this happened in https://github.com/jspecify/jspecify/issues/640.)

netdpb:
Is the task for this issue simply to enhance the non-normative text in the spec, or is it to change the normative text or its meaning?

If it's just to make the non-normative text better, then is it really necessary to do so before we release 1.0?

cpovirk:
This is to discuss whether to change the normative text. (There is also https://github.com/jspecify/jspecify/issues/521, which may end up being only non-normative text, too.)

I think that @wmdietl is no longer has strong concerns about the behavior we currently specify, so I'm hoping that we can dispatch this this _relatively) quickly.

cpovirk:
While people continue to find the current rule a little surprising when they first encounter it, they've so far also felt OK about it after discussion. I'm going to propose in today's meeting that we stick with the current rule. So I'll say here, as well, that now would be the time to let you know if you object or just would like more time to consider this.

cpovirk:
[...]

Issue #158, opened 2020-12-03
Is the declaration "Foo<@NullnessUnspecified Object>" a possible mismatch when Foo forbids nullable type arguments?


cpovirk:
The premise:

```java
@NullExplicit
interface Foo<T> {}

...
Foo<@NullnessUnspecified Object> foo; // the declaration in question
```

We already have a special rule related to this case:

> [If U is `[@NonNull]`, then U is the type variable usage’s effective nullability specifier](https://docs.google.com/document/d/15NND5nBxMkZ-Us6wz3Pfbt4ODIaWaJ6JDs4w6h9kUaY/edit#heading=h.u8phmghg96wr)

More concretely, this is "the `ImmutableList.Builder` rule": If you have an `ImmutableList.Builder<@NullnessUnspecified Object>`, we _never_ let you put nullable objects into it, even in "lenient mode." In [sample-syntax terms](https://github.com/jspecify/jspecify/blob/9da93c51a4d3c8bb87a10f42eb45dbce091c8a7e/samples/README.md#syntax), we'd call it an outright "mismatch." This is in contrast to a case of "not enough information," as you would have when putting nullable objects into a `List<@NullnessUnspecified Object>`. The difference is that `ImmutableList.Builder` is known to forbid nullable type arguments, while `List` is known to permit them.

What that rule does _not_ speak to is the _declaration_ of the variable/parameter/whatever with type `ImmutableList.Builder<@NullnessUnspecified Object>` in the first place. (Of course it's very unlikely ([perhaps impossible!](https://github.com/jspecify/jspecify/issues/137)) for a user to literally write the declaration in that form. But that type would be the _effective_ type of the object if it's declared in non-nullness-explicit code. And that _will_ happen a lot.)

And the question about the declaration is: Is it a case of "not enough information," or is it completely in the clear?

My starting point is "not enough information." I get there from naive model that that means "mismatch if you replace `@NullnessUnspecified` with `@Nullable`." (And in fact that's what the prototype checker I've been playing with currently does.)

But I see that that's not useful: Because of the special rule discussed above, `@NullnessUnspecified U` doesn't end up letting you do anything that plain `U` wouldn't already. So we already have enough information to ensure that there is no mismatch.

[...]

Issue #70, opened 2019-09-16
Special handling for equals(Object)


cpovirk:
A sentiment that I've heard in past discussions of nullness is: "`Object.equals(Object)` is known to be required to handle a `null` input, so implementations of `equals` shouldn't need to annotate the parameter explicitly."

I personally believe we should require people to annotate it anyway. (This is part of [my general dislike for special cases](https://github.com/google/codeanalysis-annotations/issues/51#issuecomment-514830030) \[edit: link goes to a discussion of `java.lang.Void`\]. FWIW, I believe [the Checker Framework requires it to be annotated](https://checkerframework.org/manual/#nullness-getting-started).) But I'm filing this bug in anticipation that we'll get such a request someday -- maybe from someone who is reading this right now and will tell me that I have it all wrong :) Even if not, we will eventually want to write up a justification of why we're going this direction.

\[note to self / other Googlers: internal issue 12964927\]

cpovirk:
In #301, I just mentioned that https://github.com/uber/NullAway/issues/619 points out that record classes end up with an `equals` method whose parameter is not annotated `@Nullable`.

I don't think that's an argument for special cases for `equals` in general: Record classes are unusual in that you don't write the source code for them. So I'd apply any special case only for record classes.

kevinb9n:
Perhaps it's like this:

Since the `MyRecord.equals` parameter comes from somewhere that is not a null-marked context, it has unspecified nullness. As such tools can treat is as they see fit, and smart ones know to treat it as nullable.

PR #645, opened 2024-09-18
Deem "`equals(Object o)`" in a record class to always have a nullable parameter type.

cpovirk:
I chose to write a rule directly about the _type usage_ that is the
parameter, rather than to write a rule that makes the whole method a
null-unmarked scope as I'd been discussing in the issue thread. I did
this so that at least the _implementation_ can remain null-marked if the
class is.

~One arguable theoretical downside to that choice is that even the
parameter's nullness-would be unspecified even in the case of:~

```
@NullMarked record Foo() {
  @NullMarked public boolean equals(Object o) { ... }
}
```

~But it would _also_ be weird if the method-level `@NullMarked` _had_ an
effect, since it's otherwise redundant with the class-level annotation.
So I'm not worried about that.~

~It would be nice if there were an obvious way to have our cake and eat
it, too, when analyzing the _source code_ for a null-marked record that
declares `equals` explicitly and omits `@Nullable` from its parameter:~

- ~We'd like that to be an error, since the rules otherwise say that the
  type is non-nullable. (This exception for record `equals` methods
  exists only because we can't distinguish compiler-generated `equals`
  methods from handwritten ones when looking at _bytecode_.)~
- ~But we don't want _callers_ to treat the `equals` method differently
  based on whether they happen to be part of the same javac invocation
  as the record class (especially in the presence of incremental
  recompilation and sharded compilation).~

~This PR's solution seems good enough for now, especially since tools can
(and [do](https://errorprone.info/bugpattern/EqualsMissingNullable))
have special cases for suggesting `@Nullable` on `equals` parameters
independent of whether the parameter is currently considered to be
non-null or to have unspecified nullness.~ We can hope that someday Java
will [provide a way to distinguish compiler-generated `equals` methods
from handwritten ones when looking at
bytecode](https://github.com/jspecify/jspecify/issues/301#issuecomment-2012625990).

This PR fixes the easily fixable part of
https://github.com/jspecify/jspecify/issues/301. I am going to see if I
can subsequently easily add at least a non-normative comment about the
other considerations that we're aware of related to bytecode analysis.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1041,7 +1041,7 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 ## Expected annotations on record classes' `equals` methods
 
 > Because of the special case [above](#augmented-type-of-usage) that makes
-> record classes' `equals` method parameter always nullable, we include this rule so that tools
+> parameters of record classes' `equals` methods always nullable, we include this rule so that tools
 > can produce expected errors in some cases when the parameter is not annotated
 > with `@Nullable`.
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -383,6 +383,14 @@ condition is met, skip the remaining conditions.
     > If the type usage is annotated with both `@Nullable` and `@NonNull`, these
     > rules behave as if neither annotation is present.
 
+-   If the type usage is the parameter of `equals(Object)` in a subclass of
+    `java.lang.Record`, then its nullness operator is `UNSPECIFIED`.
+
+    > This special case handles the fact that the Java compiler automatically
+    > generates an implementation of `equals` in `Record` but does not include a
+    > `@Nullable` annotation on its parameter, even when the class is
+    > `@NullMarked`.
+
 -   If the type usage appears in a [null-marked scope], its nullness operator is
     `NO_CHANGE`.
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1038,7 +1038,7 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
     > See ["Augmented null types."](#null-types)
 
-## Harmonizing annotations on special type usages {#harmoize-annotations}
+## Expected annotations on record `equals` methods
 
 > Because of the special case [above](#augmented-type-of-usage) that makes
 > record `equals` parameters always nullable, we include this rule so that tools
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1041,9 +1041,9 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 ## Expected annotations on record classes' `equals` methods
 
 > Because of the special case [above](#augmented-type-of-usage) that makes
-> parameters of record classes' `equals` methods always nullable, we include this rule so that tools
-> can produce expected errors in some cases when the parameter is not annotated
-> with `@Nullable`.
+> parameters of record classes' `equals` methods always nullable, we include
+> this rule so that tools can produce expected errors in some cases when the
+> parameter is not annotated with `@Nullable`.
 
 If a type usage is the parameter of `equals(Object)` in a subclass of
 `java.lang.Record`, then:
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1040,7 +1040,7 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
 ## Harmonizing annotations on special type usages {#harmoize-annotations}
 
-> Because of a special case [above][#augmented-type-of-usage] that makes record
+> Because of a special case [above](#augmented-type-of-usage) that makes record
 > `equals` parameters nullable, there is a mismatch between what `public boolean
 > equals(Object o)` means in a record class and what it would otherwise mean. To
 > harmonize this, we anticipate that tools will sometimes still produce errors
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1038,10 +1038,10 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
     > See ["Augmented null types."](#null-types)
 
-## Expected annotations on record `equals` methods
+## Expected annotations on record classes' `equals` methods
 
 > Because of the special case [above](#augmented-type-of-usage) that makes
-> record `equals` parameters always nullable, we include this rule so that tools
+> record classes' `equals` method parameter always nullable, we include this rule so that tools
 > can produce expected errors in some cases when the parameter is not annotated
 > with `@Nullable`.
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1040,9 +1040,10 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
 ## Harmonizing annotations on special type usages {#harmoize-annotations}
 
-> Because of the special case [above](#augmented-type-of-usage) that makes record
-> `equals` parameters always nullable, we include this rule so that tools can produce
-> expected errors in some cases when the parameter is not annotated with `@Nullable`.
+> Because of the special case [above](#augmented-type-of-usage) that makes
+> record `equals` parameters always nullable, we include this rule so that tools
+> can produce expected errors in some cases when the parameter is not annotated
+> with `@Nullable`.
 
 If a type usage is the parameter of `equals(Object)` in a subclass of
 `java.lang.Record`, then:
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1040,11 +1040,9 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
 ## Harmonizing annotations on special type usages {#harmoize-annotations}
 
-> Because of a special case [above](#augmented-type-of-usage) that makes record
-> `equals` parameters nullable, there is a mismatch between what `public boolean
-> equals(Object o)` means in a record class and what it would otherwise mean. To
-> harmonize this, we anticipate that tools will sometimes still produce errors
-> unless authors include the usual `@Nullable` annotation on the parameter.
+> Because of the special case [above](#augmented-type-of-usage) that makes record
+> `equals` parameters always nullable, we include this rule so that tools can produce
+> expected errors in some cases when the parameter is not annotated with `@Nullable`.
 
 If a type usage is the parameter of `equals(Object)` in a subclass of
 `java.lang.Record`, then:
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -384,7 +384,7 @@ condition is met, skip the remaining conditions.
     > rules behave as if neither annotation is present.
 
 -   If the type usage is the parameter of `equals(Object)` in a subclass of
-    `java.lang.Record`, then its nullness operator is `UNSPECIFIED`.
+    `java.lang.Record`, then its nullness operator is `UNION_NULL`.
 
     > This special case handles the fact that the Java compiler automatically
     > generates an implementation of `equals` in `Record` but does not include a
@@ -1037,6 +1037,21 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
     its nullness operator is `NO_CHANGE`.
 
     > See ["Augmented null types."](#null-types)
+
+## Harmonizing annotations on special type usages {#harmoize-annotations}
+
+> Because of a special case [above][#augmented-type-of-usage] that makes record
+> `equals` parameters nullable, there is a mismatch between what `public boolean
+> equals(Object o)` means in a record class and what it would otherwise mean. To
+> harmonize this, we anticipate that tools will sometimes still produce errors
+> unless authors include the usual `@Nullable` annotation on the parameter.
+
+If a type usage is the parameter of `equals(Object)` in a subclass of
+`java.lang.Record`, then:
+
+-   It is not expected to be annotated with `@NonNull`.
+-   If it appears in null-marked code, or if this rule is required to hold in
+    [all worlds], then it is expected to be annotated with `@Nullable`.
 
 [#100]: https://github.com/jspecify/jspecify/issues/100
 [#157]: https://github.com/jspecify/jspecify/issues/157


cpovirk:
Just to record here some points from an offline discussion with @netdpb:

- The most important thing is for the nullness not _not_ be `NO_CHANGE`/`MINUS_NULL`; we don't care _much_ whether it's `UNSPECIFIED` vs. `UNION_NULL`: Lenient tools will let you pass nullable arguments either way.
- I felt that `UNSPECIFIED` offered a few small advantages:
   - It allows a strict tool to soundly produce an error in the unlikely event that someone provides a custom implementation of `Record.equals` that actually rejects null values.
   - I feel that it's more confusing to users for the type of the `equals` parameter to become automatically nullable than for it to become automatically unspecified: A nullable type would invite users to invent other reasons for the type like ["Is the type being inherited?"](https://jspecify.dev/docs/spec/#:~:text=corresponding%20location%20in%20a%20supertype) and ["Is `equals` (always) special?"](https://github.com/jspecify/jspecify/issues/70) Not that the unspecified type makes a lot of sense, either... but at least it's a mysterious _absence_ of information instead of a suggestion that we're pulling additional information from another source.
   - I feel that it's more justified for a tool to suggest adding `@Nullable` to a manual record-`equals` declaration if the parameter is not automatically nullable.

To address the "have our cake and eat it, too" concerns noted above, @netdpb is toying with an idea that the type "of the method" would be what is written in the source code but that the type _seen by callers_ ("of the value sink?") would be different. I don't think this is a distinction that we currently make, but there may be a clean way to make it. If so, we could produce an error for manual `equals(Object)` overrides in null-marked code that omit `@Nullable` (because the type would be non-null from that perspective), but we could still let callers treat compiler-generated `equals` methods (and still the hand-written ones, sadly) as having a nullable parameter type.

As always, we hope that Java gets `?` and `!` markers, at which point it will presumably automatically generate `equals(Object? o)` for records, at which point we'll have the information we need—at least for users who target a new enough version to get that :) (And maybe Java will give us a way to identify compiler-generated `equals` methods somewhere along the way, too.)

cpovirk:
I just thought I had a cool idea, but then I realized that I didn't :(

Some [discussion of intrinsically non-null types](https://github.com/jspecify/jspecify/pull/644#issuecomment-2359409320) suggested to me that we could say that the parameter of an `equals` method in a record class is intrinsically _nullable_. That would presumably give checkers license to report an error if it were annotated as non-nullable. It would also have the magic effects that I said bad things about above, but users largely wouldn't notice.

However, our policy for intrinsically _non-null_ types suggests that tools should _also_ produce errors for the redundant use of `@NonNull`. That would suggest that tools should also produce errors for the redundant use of `@Nullable` in the case here. That would be bad, as it would essentially force users to rely on the magic.

So, upon reflection, I've realized that my new idea is _not_ a good idea.

netdpb:
I expect nobody to declare `equals(Object)` explicitly on record classes. Part of the purpose of record classes is to make that unnecessary. In fact I would expect Error Prone and other static analysis tools to flag any such declarations as presumptive errors (for `hashCode()` too, for the same reason).

The only reason we're considering a special case here is that there's no way for tools to tell the difference between the generated `equals` and an explicit one, right? If there were, then we'd say that the generated method's parameter is nullable, and we'd be happy, right?

So let's narrow the special case way down by declaring the following:

1. Record classes' `equals(Object)` method parameter is nullable _if the method is generated, not explicit_. This covers us if/when Java makes it possible to tell the difference.

2. Because tools currently can't distinguish generated `equals(Object)` methods in bytecode and explicit ones are rare, we'd suggest, non-normatively, that tools consider all record classes' `equals(Object)` methods to be generated (so see 1 above) when analyzing references to them, but not when analyzing their declarations and implementations.

kevinb9n:
Unfortunately "nobody" is quite too strong; I don't think it will be hard to find records that do it.

Have we ever discussed any variations of the following general idea?:
That when considering the "many worlds" for an unspecified type usage, we deprioritize (for some meaning thereof) the worlds that actually contravene the supertype contract... because those worlds just aren't as likely.

This has uncomfy implications, as it suggests that `boolean equals(/*unspec*/ Object o)` is sort of the same as `boolean equals(@Nullable Object o)`, and in general we don't want to inherit nullness information.

But it's only a coping strategy for unspecified nulllness; it would still have to be annotated when you null-mark....

cpovirk:
I agree with kevinb9n that people will at least occasionally override a record `equals` method, perhaps because they can get better performance by comparing the components in a different order than the order that those components appear in the API. (I said some more in your internal bug 368593933.)

> The only reason we're considering a special case here is that there's no way for tools to tell the difference between the generated `equals` and an explicit one, right? If there were, then we'd say that the generated method's parameter is nullable, and we'd be happy, right?

Exactly.

To your proposal:

- I feel a little weird about specifying something that we know is impossible at the class-file level to detect.
- I am a little hesitant to go beyond "What is the type of _x_?" to cover "What is the type of _x_ when _y_ is asking / for purpose _z_?" I do feel better about it if it's non-normative, but I'm still reluctant.
- Overall, I think we can give an imperfect but simple rule ("always unspecified if no annotation"), or we can give a more complex rule and then some non-normative advice for translating it into practice. I'd feel better about the simple rule, especially since it seems possible that we'll get language support for `Object?` before we get language support for detecting whether a record `equals` method is generated or not.

All that said, I think this is a case in which I could also say:

- The stakes here are low.
- _Anything_ that we do here is going to be imperfect and to call attention to some weirdness that we'd rather tool authors not need to know about.

So I could try to whip up something along the lines you've suggested if none of this post moves you :)

cpovirk:
To recap the stakes here:

I think that record `equals` methods in null-_unmarked_ code are actually "fine," whether they're hand-written or compiler-generated: By "fine," I mean that they're like a typical other `equals` method in null-unmarked code: imperfect but not "wrong."

The trouble here is in null-marked code, in which you've carefully annotated all the types in your code, only for the compiler to define `public boolean equals(Object o)` behind your back with no annotations. The type that arises from that is `Object!`, which actually "wrong" in a way that `Object*` is not.

The @kevinb9n proposal above could help in the unmarked case, but I think the more interesting implications lie elsewhere: It matters little _here_ because lenient tools will let callers pass null values to `equals` whether the parameter type is unspecified or nullable. A couple past discussions that have touched on more user-facing aspects of that approach are https://github.com/jspecify/jspecify/issues/158#issuecomment-2344786796 and https://github.com/jspecify/jspecify/issues/618.

netdpb:
> * I feel a little weird about specifying something that we know is impossible at the class-file level to detect.

But that seems like an oversight bug in the Java spec, not an intentional indistinguishability. I'd feel a little weird about specifying something that we expect to change as soon as the bug is fixed.

>  it seems possible that we'll get language support for Object? before we get language support for detecting whether a record equals method is generated or not.

Possible. Likely? I haven't seen any arguments against [JDK-8251375](https://bugs.openjdk.org/browse/JDK-8251375), although to be fair there hasn't been any movement at all.

I'm not familiar enough with the process. Could we push for that bug to be resolved?

netdpb:
I have a slight modification of @cpovirk's idea above.

What if we make it that the parameter to `equals(Object)` on null-marked record classes[^1] *must be nullable* (different from *intrinsically nullable*)? That's almost as simple, and we can have consistency: tools will interpret calls that way regardless of annotation, and they will also flag declarations that omit `@Nullable` (or that add `@NonNull`).

[^1]: Or maybe even on all record classes?

Both of these would be errors at compile time:

```java
@NullMarked record class Foo(...) {
  boolean equals(Object obj) { ... }
}
@NullMarked record class Bar(...) {
  boolean equals(@NonNull Object obj) { ... }
}

```

And if it was compiled anyway, any callers would correctly see a nullable parameter type, the `@NonNull` annotation notwithstanding.

cpovirk:
I am a little reluctant to use the term "bug" when the effect is limited to unusual (extralinguistic?) cases like ours. But I guess I'm willing to call our response a "workaround," so I'm apparently willing to _imply_ that term? :) And it _is_ a little weird to encode a subpar rule that could change in the future.

My impression is mostly than [JDK-8251375](https://bugs.openjdk.org/browse/JDK-8251375) might never be worth the effort of a spec change—and possibly(??) using up one of the bits in the class file, given that they [recently](https://bugs.openjdk.org/browse/JDK-8266524) [repurposed](https://openjdk.org/jeps/8303099#:~:text=acc_strict) `ACC_STRICT`. I am not holding my breath for `?`, either.

Given the low stakes here (and the lag before most users are on a new enough version of Java to absorb any change, and the fact that they might still skip to a release that _also_ has `?`), I'm not sure the Java change is worth pushing on. Maybe I'd feel differently if I were to get to the point that I actually had a JDK bug-tracker account and a few PRs under my belt so that I could feel more like a contributor and less like a source of endless demands :)

---

I started to update the PR to go with your proposal, but that brought back to me part of what had worried me about that style of proposal.

Suppose that I have this code:

```java
@NullMarked
@SuppressWarnings("nullness") // TODO: resolve nullness errors
record class Foo(int i) {
  @Override public boolean equals(Object o) {
    return o instanceof Foo other && i == other.i;
  }
}

class User {
  boolean b = new Foo(0).equals(null);
}
```

Now suppose that I compile `Foo.java` and `User.java` as part of the same javac invocation. In that case, we know that `equals` is explicit, so we treat its parameter as non-null, and we produce an error. Now suppose that we compile `Foo.java` and `User.java` as part of separate javac invocations. In that case, we don't know that `equals` is explicit, so we treat its parameter as nullable, and we don't produce an error.

Is that a practical problem? Maybe, maybe not. I had mentioned incremental recompilation and sharded compilation above, and I think there could be a problem herewith sharded compilation, especially if the contents of the shards were shuffled when a new file is added: The compile succeeds (because the two files end up in different shards), but the you add another file, and it files (because the two files end up in the same shard).

That's probably still not going to come up in practice, since the code above is terrible for multiple reasons. But it brings me back to my position that we'd be writing a more complex rule that has minimal practical user impact—only now some of that minimal impact is negative.

And I see that you posted again while I was typing. I'll have a look shortly.

cpovirk:
For [the "must be nullable" proposal](https://github.com/jspecify/jspecify/pull/645#issuecomment-2364581014): That sounds like the effect that we want, but my worry is still complexity: We're introducing a new concept of "must be _x_" that supports only this case (and we _hope_ only temporarily, but really, I'm still not holding my breath).

I guess I would formulate it as something like:

- The nullness of a record `equals` parameter _is_ `UNION_NULL`. This is the new first rule of computing the nullness of a type usage, applied even before looking for annotations.
- Tools must be able to identify cases in which they're analyzing code that declares a record `equals` parameter that has a `@NonNull` annotation _or_ has no nullness annotation and is located in null-marked code (_or_ has no nullness annotation and is located in null-_unmarked_ code if the property is being checked for whether it holds in all worlds?).

I think I could write that down if it would make this issue go away. I also feel like it would go straight to the bottom of the pile of features/changes yet to be implemented in the reference checker, and I would be fairly tempted to implement the first half and leave the other half for the indefinite future. But that's probably still OK, given the relatively low stakes beyond making sure that `someRecord.equals(someNullableType)` isn't an error. Does that sound right?

netdpb:
I mean, whatever we do we're introducing a sad special case to deal with this one Java wart. If we go with what you suggested at first this will be the only case of a null-marked method containing an unspecified type component. 😢 all around.

I wouldn't bother specifying that tools should do anything about explicit null-unmarked unannotated parameters. A strict environment won't have any of those, and a lenient environment certainly won't care.

I'm perfectly fine with the reference checker or any other tool deprioritizing handling of expliicit `equals` in records since I'm on record (ha) saying that I expect there to be zero of them.

cpovirk:
> I wouldn't bother specifying that tools should do anything about explicit null-unmarked unannotated parameters. A strict environment won't have any of those

My concern would be specifically that the rule we're writing _would_ allow a strict environment to have those: `public boolean equals(Object o)` would be treated as having a nullable parameter, so it would be correct even to a strict checker.

But that, too, comes up only in the unlikely event that a user overrides a record `equals` method in the first place.

And you've convinced me that "automatically nullable and must not be annotated as non-null" is better than "automatically unspecified." To go into extreme detail, I think the breakdown of the tradeoff between those two is:
- (the very, very, very common case) No one overrides a record `equals` method, and people use lenient tools: In that case, unspecified and nullable are equivalent, but nullable might lead to better... tooltips, I guess, since that's the main case in which the type would ever become visible to users?
- Someone overrides a record `equals` method, and people use lenient tools: If the `equals` method handles `null` correctly, then nullable would be better. And if the `equals` method _doesn't_ handle `null` correctly, then nullable is also better because a checker run over the `equals` method can produce an error.
- No one overrides a record `equals` method, and people use strict tools: Nullable is better because it lets callers pass nullable expressions without an error.
- Someone overrides a record `equals` method, and people use strict tools: If the checker is run over the `equals` declaration, then any null dereferences will be caught either way, and the checker will require an annotation either way (assuming that I include the rule discussed at the beginning of this post). If the checker is run over _callers_, then nullable is more convenient, and unspecified is safer. But either is safe if we assume that the checker is run over the `equals` declaration. (I'm assuming here that no one is _intentionally_ writing an `equals` method that rejects nulls and then trying to document that fact by leaving the parameter unannotated, since adding `@NonNull` would be "redundant" (kind of like our discussion of how it would be redundant in https://github.com/jspecify/jspecify/issues/618) :))

The only case that's even _arguably_ better for "unspecified" is that last one, and it's only better in that it's stricter about unannotated code that might not have been checked.

Hmm, ugh, I'm not sure I had [the difference between null-marked and null-unmarked code](https://github.com/jspecify/jspecify/pull/645#issuecomment-2364181893) clear in my head when I wrote all that. Well, I still don't think it's going to change things much. I'm comfortable going with the "nullable" approach. I'll update this PR so that we have something concrete to look at.

cpovirk:
OK, back at you.

docs/docs/spec.md
```diff
@@ -1038,6 +1038,21 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
     > See ["Augmented null types."](#null-types)
 
+## Harmonizing annotations on special type usages {#harmoize-annotations}
```
netdpb, line 1041: I'm expecting The Beach Boys to jump in here. (_Ba-ba-ba-ba-ba-Barbara Annotation!_)

If this bit is really speciifc to `equals(Object)`'s parameter in record classes, then lets name this section more specifically: no harmonization, just "Record class `equals` parameter" or something.

On the other hand, we could make at least this bit a little more general, saying that in any case where an explicit type usage in code is nullable regardless of annotation, then it can't have `@NonNull` and it must have `@Nullable` in null-marked code or in all-worlds checkers. Then, non-normatively, we could point out that the only such case is the special one for record class `equals` parameters.

docs/docs/spec.md
```diff
@@ -1038,6 +1038,21 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
...
+
+> Because of a special case [above](#augmented-type-of-usage) that makes record
+> `equals` parameters nullable, there is a mismatch between what `public boolean
+> equals(Object o)` means in a record class and what it would otherwise mean. To
+> harmonize this, we anticipate that tools will sometimes still produce errors
+> unless authors include the usual `@Nullable` annotation on the parameter.
```
netdpb, line 1047: Maybe there's a clearer thing to say than "there is a mismatch between what X means in a record class and what it would otherwise mean."

What about something like this:

```suggestion
> Because of the special case [above](#augmented-type-of-usage) that makes record
> `equals` parameters always nullable, we include this rule so that tools can produce
> expected errors in some cases when the parameter is not annotated with `@Nullable`.
```

docs/docs/spec.md
cpovirk: I had hoped to find a title to distinguish this section from the _other_ discussion of record `equals` methods above. But then I had gotten worried that I'd end up using terminology that suggests that we're mandating user-facing tool behavior :) I ended up here, with a title that attempts to suggest that it's a piece of a larger puzzle.

But in the end, I also had to settle on some phrasing for the bullet points under this section, and I decided I could live with "expected." That's a little more generic than I'd aim for in a section title, since people could consider various annotations to be "unexpected" or "unexpected" in various cases. But I think it's fine, especially with an explicit mention of record `equals` methods (and a link back to the _other_ section about record `equals` methods to emphasize that it's not the whole story).

So:

#### Expected annotations on record `equals` methods

docs/docs/spec.md
```diff
@@ -1038,13 +1038,12 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
 
     > See ["Augmented null types."](#null-types)
 
-## Harmonizing annotations on special type usages {#harmoize-annotations}
+## Expected annotations on record `equals` methods
```
netdpb, line 1041:
```suggestion
## Expected annotations on record classes' `equals` methods
```

docs/docs/spec.md
```diff
@@ -1038,13 +1038,12 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
...
 
-> Because of a special case [above](#augmented-type-of-usage) that makes record
-> `equals` parameters nullable, there is a mismatch between what `public boolean
-> equals(Object o)` means in a record class and what it would otherwise mean. To
-> harmonize this, we anticipate that tools will sometimes still produce errors
-> unless authors include the usual `@Nullable` annotation on the parameter.
+> Because of the special case [above](#augmented-type-of-usage) that makes
+> record `equals` parameters always nullable, we include this rule so that tools
```
netdpb, line 1044:
```suggestion
> record classes' `equals` method parameter always nullable, we include this rule so that tools
```

netdpb (approved):

PR #647, opened 2024-09-18
Deem type-use annotations to be recognized on record-component types.

cpovirk:
They are [automatically copied everywhere that we'd hope
for](https://openjdk.org/jeps/395#:~:text=Type%20annotations%20on,automatically%20derived%20members.).

I can't seem to find any issue discussions of this, but we cover it in
[the
Javadoc](https://jspecify.dev/docs/api/org/jspecify/annotations/Nullable.html).

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -231,6 +231,8 @@ exceptions in the subsequent sections:
 
 -   a field type
 
+-   a record component type
+
 -   a type parameter upper bound
 
 -   a non-wildcard type argument


netdpb (approved):
This is fine, but how do we square mentioning records with our statement that we're only covering language features up to Java 11?

cpovirk:
Do you know what the most official place we've said that we're covering only up to Java 11 is? I know it's come up, but if we're just talking about a conflict between the spec and, say, some comments on issues, then that's probably a lesser evil than not mentioning records at all.

I'm also not sure how much else we would need to cover, especially to get only to [Java 16 for records](https://openjdk.org/jeps/395) ([14 if you count previews](https://openjdk.org/jeps/359)).

And I guess we've sort of crossed that line already by mentioning annotations inside the `instanceof` type, which [weren't permitted in Java 11](https://github.com/jspecify/jspecify/issues/626#issuecomment-2344326037).

I think we're both OK with merging this, but if you have thoughts on any of that, we may be able to come up with a better story. Or the story might just remain "We're trying to be pragmatic" :)

PR #648, opened 2024-09-18
Expand Javadoc to cover that all parts of the `instanceof` type are unrecognized locations.

cpovirk:
See
https://github.com/jspecify/jspecify/pull/644#issuecomment-2359409320


cpovirk:
I am not super-optimistic that I found good phrasing for this, but I figured you'd appreciate an attempt in a PR more than a bug assigned to you :)

kevinb9n (approved):

cpovirk:
Given what I settled on in https://github.com/jspecify/jspecify/pull/668, I should redo this to cover all patterns (to match what that other PR did in the spec). Or we could just live with the Javadoc text as it is for now, given that we aren't sure what we'll ultimately do with patterns.

Whatever we ultimately do for https://github.com/jspecify/jspecify/issues/626, we'll obviously want agreement between the spec and the Javadoc :)

PR #650, opened 2024-09-18
Make annotation element types intrinsicially non-nullable.

cpovirk:
This is already covered in [the
Javadoc](https://jspecify.dev/docs/api/org/jspecify/annotations/Nullable.html),
but the spec didn't cover it until now.


cpovirk:
(as noticed in https://github.com/jspecify/jspecify/pull/644#issuecomment-2359409320)

netdpb (approved):

PR #651, opened 2024-09-19
Specify that type variables from capture conversion are `NO_CHANGE`.

cpovirk:
(@kevinb9n had prompted me to think about this almost two years ago, and
I'm only now getting around to it as I go through notes to prepare for
spec 1.0.)


docs/docs/spec.md
```diff
@@ -1025,6 +1025,9 @@ The Java rules are defined in [JLS 5.1.10]. We add to them as follows:
     > This is just a part of our universal rule to treat a bare `?` like `?
     > extends Object`.
 
+-   Every fresh type variable generated by the rule has a nullness operator of
```
netdpb, line 1028: JLS doesn't define "fresh type variable"; is it _funky fresh_, or is it impertinent?

But I guess they mean "a type usage of a brand new type parameter". Is it important to clarify that you mean the type variable has `NO_CHANGE` and not, somehow, the type parametr?

That is, `void method(Bar<? extends @Nullable Foo>)` becomes `<FRESH extends Foo UNION_NULL> void method(Bar NO_CHANGE<F NO_CHANGE>);`, and _not_ `<FRESH NO_CHANGE extends Foo UNION_NULL> void method(Bar NO_CHANGE<F NO_CHANGE>);`

Or is that clear to everyone?

netdpb (approved):

docs/docs/spec.md
cpovirk: Yeah, I was wondering if it was worth saying "Every fresh _usage_ of a type variable," but then I worried that that would be deviating enough from the language in the JLS that it might lead to confusion.

But also, as I re-read this 11 minutes after sending the PR, I do notice that it looks reads as crystal-clear that I'd remembered.

I've rephrased to keep "fresh type variable" as a standalone, intact phrase but to still speak to the _usage_ of it.

And I noticed that "rule" should be plural here (or at least "a rule" instead of "the rule"), so I've touched that up, too.

PR #652, opened 2024-09-19
Clarify that we reserve _single_ capital letters for types.

cpovirk:
(I think this is in response to another two-year-old comment from
@kevinb9n.)

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -187,9 +187,9 @@ and interface types, array types, and type variables but also
 > for ["Bound of an unbounded wildcard,"](#unbounded-wildcard) for which it
 > specifies a bound of `Object` that the JLS does not specify.
 
-When this spec uses single capital letters, they refer to augmented types
-(unless otherwise noted). This is in contrast to the JLS, which typically uses
-them to refer to base types.
+When this spec uses single capital letters in `code font`, they refer to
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -187,9 +187,9 @@ and interface types, array types, and type variables but also
 > for ["Bound of an unbounded wildcard,"](#unbounded-wildcard) for which it
 > specifies a bound of `Object` that the JLS does not specify.
 
-When this spec uses capital letters, they refer to augmented types (unless
-otherwise noted). This is in contrast to the JLS, which typically uses them to
-refer to base types.
+When this spec uses single capital letters, they refer to augmented types
```
netdpb, line 190: Do we care about saying single capital letters in `code font`?

Separately (and not necessary for 1.0) should we in fact use a different notation? `T`<sub>`operator`</sub>?

netdpb (approved):

docs/docs/spec.md
cpovirk: Sure, I'll mention code font.

And thanks, I've filed https://github.com/jspecify/jspecify/issues/658 about notation.

I take part of your point to be that we might then follow the JLS in using `T` for base types because we could easily notate unknown/arbitrary _augmented_ types as `T`<sub>`x`</sub>. It would be interesting to see how that works out.

PR #657, opened 2024-09-20
Improve explanation of why "the `ImmutableList.Builder` rule" produces `MINUS_NULL`.

cpovirk:
I'm actually wondering if we should just cut this discussion entirely.
It's strictly less important than the _main_ explanation of the
`ImmutableList.Builder` rule, and we may remove _that_ explanation in
favor of a link to a doc.

PLus, I think it would be reasonable to argue that "obviously" you use
`MINUS_NULL` because you're substituting for a non-nullable type
parameter. Thus, the explanation I'm giving is a more complicated
explanation of something that users would naturally already expect.

(As for the actual content of the PR: I noticed that I _think_ the
`Sequence` class has to be null-marked (and not _merely_ to have
`extends @Nullable Object` in its type-parameter declaration) in order
for the example to hold up after our decision in
https://github.com/jspecify/jspecify/issues/248. If not for
`@NullMarked`, then the type (sans `IL.B` rule) would be `Optional<E*>`,
and dereferencing `E*` would not be an error to a lenient checker. We
need for the type to be `Optional<E>` (again, sans `IL.B` rule) so that
a dereference would be an error. That way, we demonstrate that the
`IL.B` rule saves us from that false-positive error.)

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -927,7 +927,7 @@ the output of the following operation:
     > demonstrated in
     > https://github.com/jspecify/jspecify-reference-checker/pull/197.
 
-    > The purpose of this special case is to improve behavior in "the
+    > The primary purpose of this special case is to improve behavior in "the
     > `ImmutableList.Builder` case": Because `ImmutableList.Builder.add` always
     > throws `NullPointerException` for a null argument, we would like for
     > `add(null)` to be a compile error, even under lenient tools.
@@ -948,15 +948,14 @@ the output of the following operation:
     > `MINUS_NULL`.
     >
     > The choice between `NO_CHANGE` and `MINUS_NULL` makes little difference
-    > for the parameter types of `ImmutableList.Builder`, but it can matter more
-    > for other APIs' *return types*. For example, consider `@NullMarked class
-    > Foo<E extends @Nullable Object>`, which somewhere uses the type
-    > [`FluentIterable<E>`]. `FluentIterable` has a method `Optional<E>
-    > first()`. Even when `E` is a type like `String UNION_NULL` (or `String
-    > UNSPECIFIED`), we know that `first().get()` will never return `null`. To
-    > surface that information to tools, we need to define our substitution rule
-    > to return `E MINUS_NULL`: If we instead used `E NO_CHANGE`, then the
-    > return type would look like it might include `null`.
+    > for the `ImmutableList.Builder` case and for most other cases. It may
+    > matter only in unusual cases that involve misannotated code. For example,
+    > consider a `@NullMarked class Sequence<E extends @Nullable Object>` that
+    > declares a method `Optional<E> firstNonNull()`. That method *should* be
+    > declared with a return type of `Optional<@NonNull E>`. Still, this spec
+    > rule ensures that `Sequence` can call `firstNonNull().get().toString()`.
+    > That's possible because the rule makes the type of `firstNonNull().get()`
+    > be `E MINUS_NULL`, rather than `E NO_CHANGE`.
 
 -   Otherwise, replace `V` with the output of applying the nullness operator of
     `V` to `Aᵢ`.


cpovirk:
Hmm, wait, `MINUS_NULL` is _even more_ obviously correct than I'd thought!

I realized that we check whether to apply the `ImmutableList.Builder` rule at the very beginning of our substitution rules—and that it acts as a _substitute_ (sorry, confusing word choice!) for the normal substitution rule.

The rule "triggers" for any null-exclusive type. It's designed for types like the `E` in `@NullMarked class Optional<E>`, but of course it would also trigger for a type like `@NonNull T` in a class where `<T extends @Nullable Object>`. In that latter case, it's _very clearly_ important that substitution produces a non-nullable type, even for a (legal!) nullable type argument. We'd naturally get that from the normal substitution rule simply because `@NonNull T` has nullness operator `MINUS_NULL`. _However_, because we first check whether to apply the `ImmutableList.Builder` rule, what _actually_ matters is what the `ImmutableList.Builder` rule says to do. Thus, it's very important that it says to apply `MINUS_NULL` rather than `NO_CHANGE`.

I am probably going to follow through on my threat above to remove this part of the discussion entirely....

docs/docs/spec.md
```diff
@@ -942,21 +942,10 @@ the output of the following operation:
...
-    > for other APIs' *return types*. For example, consider `@NullMarked class
-    > Foo<E extends @Nullable Object>`, which somewhere uses the type
-    > [`FluentIterable<E>`]. `FluentIterable` has a method `Optional<E>
-    > first()`. Even when `E` is a type like `String UNION_NULL` (or `String
-    > UNSPECIFIED`), we know that `first().get()` will never return `null`. To
-    > surface that information to tools, we need to define our substitution rule
-    > to return `E MINUS_NULL`: If we instead used `E NO_CHANGE`, then the
-    > return type would look like it might include `null`.
+    > type parameters like the one on `ImmutableList.Builder`, one that avoids
+    > producing a type with nullness operator `UNSPECIFIED`. Since the special
```
netdpb, line 946: The two "one"s confused me. How about this?

```suggestion
    > type parameters (like the one on `ImmutableList.Builder`) that doesn't
    > produce a type with nullness operator `UNSPECIFIED`. Since the special
```

cpovirk:
So, uh. Sorry, I left that kind of a mess.

I think that the original commit here gave an accurate description of one edge-case reason to produce `MINUS_NULL` over `NO_CHANGE`.

Then, as I was writing the commit description, I realized that `MINUS_NULL` was probably the natural choice, anyway, so I probably didn't need to go to such lengths to justify it.

_Then_, I realized that it was not _just_ the natural choice that helps in an edge case but actually the way that it _clearly must be_.

At this point, if you look at the changed files, you'll see only the simpler explanation of that last part. Basically, it says that we want the `ImmutableList.Builder` rule (assuming that [we _do_ want it](https://github.com/jspecify/jspecify/issues/618) :)) to cover the `ImmutableList.Builder` case. And then, once we have the special case, we need for it to use `MINUS_NULL` so that the special case doesn't mess up cases that more directly _asked_ for `MINUS_NULL`.

You might now be able to ignore previous instructions and just read this post as preparation for the review.

docs/docs/spec.md
```diff
@@ -927,7 +927,7 @@ the output of the following operation:
     > demonstrated in
     > https://github.com/jspecify/jspecify-reference-checker/pull/197.
 
-    > The purpose of this special case is to improve behavior in "the
+    > The primary purpose of this special case is to improve behavior in "the
     > `ImmutableList.Builder` case": Because `ImmutableList.Builder.add` always
```
netdpb, line 931: I'd love to avoid referring to a Guava class. Could we motivate this with a hypothetical class (or one in the JDK&mdash;what about `Comparable`?), to make it clear that this isn't really a special case? It's asymmetrical, but that's a function of the fact that `Foo!` is a strict subtype of `Foo?`.

But if we're going to refer to `ImmutableList.Builder`, shouldn't we at least link to it? It's not the case that everyone reading this spec will be intimately familiar with that API.

cpovirk: I made an attempt to rewrite this to be about `Comparable` and to make clear that it's not a special case. PTAL.

docs/docs/spec.md
```diff
@@ -927,36 +927,28 @@ the output of the following operation:
...
-    > to return `E MINUS_NULL`: If we instead used `E NO_CHANGE`, then the
-    > return type would look like it might include `null`.
+    > -   If JSpecify were to directly subsitute `Foo UNSPECIFIED` in for `T`,
+    >     then the parameter type, which started out as non-null, would become
+    >     unspecified during subsitution. As a result, lenient checkers would
+    >     allow the call `foo().compare(null)`, since `Foo UNSPECIFIED` is
+    >     [null-inclusive under every parameterization] in [some world].
+    > -   To avoid that, JSpecify recognizes that the parameter is non-null, and
+    >     it performs substitution as if the parameter type were `T MINUS_NULL`
+    >     instead of `T NO_CHANGE`.
```
cpovirk, line 951: I was tempted to add here: "JSpecify treats those types equivalently before substitution, so it makes sense for JSpecify to ensure that that are equivalent after substitution, too."

I couldn't decide whether that would make things better or worse.

docs/docs/spec.md
```diff
@@ -927,36 +927,28 @@ the output of the following operation:
...
-    > Without this special case, the parameter of `builder.add` would have a
-    > nullness operator of `UNSPECIFIED`, too. Then, when a lenient tool would
-    > check whether the [some-world] subtyping relation holds for
-    > `builder.add(null)`, it would find that it does.
+    > For an example of such a type, consider `Comparable`, a `@NullMarked`
+    > interface that declares a non-nullable type parameter `T` and a method
+    > `compare(T)`. By JSpecify rules, the method parameter has type `T
+    > NO_CHANGE`, and that type is null-exclusive under every parameterization
+    > in all worlds. Now consider a null-unmarked class that declares a method
+    > `Comparable<Foo> foo()`, which by JSpecify rules has a type argument `Foo
```
netdpb, line 939: Is it worth having `@NonNull Comparable<Foo>` here, to avoid questions about dereferencing `foo()` below? Or is that too distracting, and people will know what you mean?

docs/docs/spec.md
```diff
@@ -927,36 +927,28 @@ the output of the following operation:
...
+    > UNSPECIFIED`. In this example, the question is what type
+    > `foo().compare(...)` accepts, and the answer changes as a result of this
```
netdpb, line 941: Framing this as "the answer changes" suggests dynamic rules somehow.

What about "this part of the substitution rule gives us a better answer"?

docs/docs/spec.md
```diff
@@ -927,36 +927,28 @@ the output of the following operation:
...
-    > for other APIs' *return types*. For example, consider `@NullMarked class
-    > Foo<E extends @Nullable Object>`, which somewhere uses the type
-    > [`FluentIterable<E>`]. `FluentIterable` has a method `Optional<E>
-    > first()`. Even when `E` is a type like `String UNION_NULL` (or `String
-    > UNSPECIFIED`), we know that `first().get()` will never return `null`. To
-    > surface that information to tools, we need to define our substitution rule
-    > to return `E MINUS_NULL`: If we instead used `E NO_CHANGE`, then the
-    > return type would look like it might include `null`.
+    > -   If JSpecify were to directly subsitute `Foo UNSPECIFIED` in for `T`,
+    >     then the parameter type, which started out as non-null, would become
```
netdpb, line 945:
```suggestion
    > -   Without this part of the rule, JSpecify would directly subsitute `Foo UNSPECIFIED` in for `T`;
    >     then the parameter type, which started out as non-null, would become
```

docs/docs/spec.md
```diff
@@ -927,36 +927,28 @@ the output of the following operation:
...
+    >     unspecified during subsitution. As a result, lenient checkers would
+    >     allow the call `foo().compare(null)`, since `Foo UNSPECIFIED` is
+    >     [null-inclusive under every parameterization] in [some world].
+    > -   To avoid that, JSpecify recognizes that the parameter is non-null, and
```
netdpb, line 949:
```suggestion
    > -   To avoid that, JSpecify uses this rule to recognize that the parameter is non-null, and
```

docs/docs/spec.md
netdpb, line 951: Should you describe what the `compare(...)` argument type becomes, in parallel to what you do in the previous bullet?

netdpb: I had to think for a bit before I understood what you meant by "those types", so I don't think that sentence helps.

But maybe if you said more explicitly something like "JSpecify treats `T NO_CHANGE` just like `T MINUS_NULL` when `T` is non-nullable, and this rule preserves that equivalence through substitution, even when substituting `UNSPECIFIED` types."

If you do that, you should do it after these two bullets, not as part of the second bullet.

cpovirk: Good point.

cpovirk (commented):
Thanks.

docs/docs/spec.md
```diff
@@ -938,17 +928,27 @@ the output of the following operation:
...
-    > part of the subsitution rules:
+    > `foo().compare(...)` accepts. That question demonstrates the effect of
+    > this part of the subsitution rule:
     >
-    > -   If JSpecify were to directly subsitute `Foo UNSPECIFIED` in for `T`,
-    >     then the parameter type, which started out as non-null, would become
-    >     unspecified during subsitution. As a result, lenient checkers would
-    >     allow the call `foo().compare(null)`, since `Foo UNSPECIFIED` is
+    > -   Without this part of the rule, JSpecify would directly subsitute `Foo
+    >     UNSPECIFIED` in for `T`. Then the parameter type, which started out as
```
netdpb, line 935: Do we need the "in" here? "substituted X in for Y" vs "substituted X for Y".

_This is the most consequential comment I've ever made._

docs/docs/spec.md
```diff
@@ -927,7 +927,7 @@ the output of the following operation:
     > demonstrated in
     > https://github.com/jspecify/jspecify-reference-checker/pull/197.
```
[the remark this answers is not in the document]
netdpb: Section headers inside bullets inside non-normative blocks inside parentheses.... where does it end?

netdpb (approved):
Nice work!

docs/docs/spec.md
cpovirk: Done.

(I started wondering whether we should say "for `T NO_CHANGE`" or otherwise attempt to be more precise, but then I stuffed it back down.)

cpovirk (commented):
Thanks!

PR #660, opened 2024-09-24
Make enum constants intrinsically non-nullable.

cpovirk:
They were already unrecognized locations for nullness annotations, but
now we also specify that they're non-nullable.

(Also, this PR includes a bit of reformatting that I'd apparently missed
in #649.)

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -391,8 +391,8 @@ condition is met, skip the remaining conditions.
     constant, its nullness operator is `MINUS_NULL`.
 
     > In source code, there is nowhere in the Java grammar for the type of an
-    > enum constant to be written. Still, enum constants have a type,
-    > which is made explicitly visible in the compiled class file.
+    > enum constant to be written. Still, enum constants have a type, which is
+    > made explicitly visible in the compiled class file.
 
 -   If the type usage is a component of a return type in an annnotation
     interface, its nullness operator is `MINUS_NULL`.
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -387,8 +387,8 @@ usage, this section covers only how to determine its [nullness operator].
 To determine the nullness operator, apply the following rules in order. Once one
 condition is met, skip the remaining conditions.
 
--   If the type usage is the type of an enum constant, its nullness operator is
-    `MINUS_NULL`.
+-   If the type usage is the type of the field corresponding to an enum
+    constant, its nullness operator is `MINUS_NULL`.
 
     > In source code, there is nowhere in the Java grammar for the type of an
     > enum constant to be written. Still, enum constants have an implicit type,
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -387,6 +387,9 @@ usage, this section covers only how to determine its [nullness operator].
 To determine the nullness operator, apply the following rules in order. Once one
 condition is met, skip the remaining conditions.
 
+-   If the type usage is the type of an enum constant, its nullness operator is
```
netdpb, line 390: When does the type of an enum constant appear as a type usage in code? I guess it's bytecode and not source code? Should we note that here, possibly with an example?

netdpb (approved):

docs/docs/spec.md
cpovirk: Hmm, good question. I am having more trouble getting a handle on this than I probably should.

I think we'd want to know the type of an enum constant `FOO` even within the same file, so it seems strange to have to appeal to the class file in order to explain that it has a type.

I'm thinking that maybe this is more a case of an _implicit_ type, sort of like how `<T>` means `<T extends Object>`—except without the option to make the type explicit in this case.

Maybe that's an argument that we should have a special section just for enum constants, just as we do for `<T>` and `<?>`? Or maybe that's overkill, since the rule for enum constants is easier to explain (I, uh, hope) than the rules for `<T>` and `<?>`?

Do you buy that? And what sort of example were you thinking might help?

(I'll push a commit with a specific proposal.)

netdpb: Well, [JLS 8.9.3](https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-8.9.3) says:

> For each enum constant c declared in the body of the declaration of E, E has an implicitly declared `public static final` field _of type E_ [emphasis added] that has the same name as c. The field has a variable initializer which instantiates E and passes any arguments of c to the constructor chosen for E. The field has the same annotations as c (if any).

So you could say that you're talking about the type of that implicit field.

cpovirk: Aha, thanks, that fixes a thing I was missing.

I've pushed a new commit. Let me know if you think it could use a JLS link, further rephrasing, or other changes.

cpovirk: (Maybe I should remove the non-normative comment now that we've clarified this? I like to think that I've already showed uncharacteristic restraint in refraining from inserting an additional non-normative bit about how _actually type-use annotations can be used on enum constants, but we don't consider those type usages, and JSpecify annotations there are unrecognized, as discussed in..._ :))

docs/docs/spec.md
```diff
@@ -387,8 +387,12 @@ usage, this section covers only how to determine its [nullness operator].
 To determine the nullness operator, apply the following rules in order. Once one
 condition is met, skip the remaining conditions.
 
--   If the type usage is the type of an enum constant, its nullness operator is
-    `MINUS_NULL`.
+-   If the type usage is the type of the field corresponding to an enum
+    constant, its nullness operator is `MINUS_NULL`.
+
+    > In source code, there is nowhere in the Java grammar for the type of an
+    > enum constant to be written. Still, enum constants have an implicit type,
```
netdpb, line 394:
```suggestion
    > enum constant to be written. Still, enum constants have a type,
```

netdpb (approved):

Issue #661, opened 2024-09-24
Cover elements in implementation code in "Augmented type of a type usage appearing in code"


cpovirk:
For spec 0.3, we said that annotations anywhere in implementation code were "unrecognized," which at the time was a broader category that contained not only "bad" locations (e.g., `class Foo extends @Nullable Bar`) but also locations that we just wanted to call out of scope.

For spec 1.0, we now use "unrecognized" only for "bad" locations, so locations like local-variable type arguments are now _recognized_. Yay!

For some future spec version, we should specify more about the _nullness_ of at least _elements_ in implementation code. This means things like:
- Exception parameters are non-null.
- Local variables (including resource parameters, meaning those in `try-with-resources`) are nullable. Or perhaps they're "inferred" in some different sense than nullness is normally inferred, as we describe it in [the Javadoc](https://jspecify.dev/docs/api/org/jspecify/annotations/Nullable.html):
   ```java
   @NullMarked
   class Foo {
     final @Nullable String s;

     Foo(String sParam) {
       this.s = sParam;
       // The element `s` has type `String?`, but dataflow knows that the expression `s` has type `String!`.
       Boolean b = true;
       // TBD:
       // - The element `b` has type `Boolean?`, but dataflow knows that the expression `b` has type `Boolean!`?
       // - The element `b` has type `Boolean` at this point?
     }
   }
   ```
   - Local variables declared by `instanceof` are non-null. (This is separate from [another issue that is about _subcomponents_](https://github.com/jspecify/jspecify/issues/626).)

My thinking is that this initial work would still not need to get into complexities like [how casts work](https://github.com/jspecify/jspecify/issues/252), let alone dataflow or even the basics of expressions (e.g., the fact that string concatenation returns a non-null string): I'm thinking here only about "elements"/"symbols," to use javac terms.

cpovirk:
(related: [rules for union types](https://github.com/jspecify/jspecify/issues/683))

Issue #622, opened 2024-09-06
Define what _types_ are enclosed by


cpovirk:
https://github.com/jspecify/jspecify/pull/620/files#r1746129824

We currently discuss only what _other elements_ are enclosed by.

agentgt:
Probably a dumb question that I just don't know because annotation processors do not have access to local types but what happens for local types?

```java

public class Blah {
  public void someMethod() {
    record LocalType() {}
  }

}
```

I would assume `LocalType` enclosed type is `Blah` but maybe it returns `null`.

cpovirk:
We currently [have](https://jspecify.dev/docs/spec/#null-marked-scope):

> "classes... may be enclosed by other class members or classes"

I think our model and the Java reflection model hold that `LocalType` is enclosed by `someMethod`, which is in turn enclosed by `Blah`. As you note, that middle part can matter now that we permit `@NullMarked` and `@NullUnmarked` on methods.

cpovirk:
One "recent" development for us to include is record copmonents.

It could also be nice to clarify that nothing is enclosed by a _field_, even an anonymous class that appears in a field initializer. (But presumably a field is _enclosed by_ the containing class. Maybe we could avoid having to say even that once we work out enclosure of _types_. But I'm not sure it would be worth avoiding, especially since I imagine that the annotation-processing API _would_ consider the field to be enclosed by the class.)

netdpb:
> It could also be nice to clarify that nothing is enclosed by a _field_

Why is this nice?

We don't allow `@NullMarked` or `@NullUnmarked` on fields, but we don't have to say that fields can't enclose any type uses, do we?

cpovirk:
Partially, it's nice to match [the `java.lang.model` API](https://docs.oracle.com/en/java/javase/23/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement()), since that what at least some tool authors will be using. (But note that we are currently [inconsistent with it about _type parameters_](https://github.com/jspecify/jspecify/pull/681), though that could change.)

It's also kind of nice that, when a tool that checks "all enclosing elements" for `@NullMarked` and `@NullUnmarked`, it's not checking elements on which those annotations wouldn't be recognized, anyway, such as the field that might appear to "enclose" an anonymous class. But that's a very minor concern, and we can't accommodate _both_ it _and_ the concern about type parameters mentioned above.

[...]

PR #662, opened 2024-09-25
Clarify that tools should look for annotations only in recognized locations.

cpovirk:
We already say this for our type-use annotations, so this PR adds
similar phrasing for our declaration annotations.

Motivation:

- Make the two sections consistent.
- Further emphasize that `@NullMarked` and `@NullUnmarked` can't appear
  on fields. (Another effort that could help with that is
  https://github.com/jspecify/jspecify/issues/622.) This is mostly a
  future concern, since the annotations' `@Target` dosen't permit usage
  on fields, but weird things might happen if we add `@Implies` someday.
- Further emphasize that `@Nullable` and `@NonNull` have no effect on
  local variables, etc. (This change is a tiny step toward
  https://github.com/jspecify/jspecify/issues/661.)

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -369,10 +369,8 @@ If none of the enclosing declarations meet either rule, then the type usage is
 
 ## Augmented type of a type usage appearing in code {#augmented-type-of-usage}
 
-For most type usages in source code or bytecode where JSpecify nullness
-annotations are [recognized](#recognized-type-use), this section defines how to
-determine their [augmented types]. Note, however, that rules for specific cases
-below take precedence over the general rule here.
+This section defines how to determine the [augmented types] of most type usages
+in source code or bytecode where JSpecify nullness annotations are [recognized].
 
 > The rules here should be sufficient for most tools that care about nullness
 > information, from build-time nullness checkers to runtime dependency-injection
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -205,7 +205,7 @@ All annotations have runtime retention. None of the annotations are marked
 
 We provide two parameterless type-use annotations: `@Nullable` and `@NonNull`.
 
-### Recognized locations for type-use annotations
+### Recognized locations for type-use annotations {#recognized-type-use}
 
 A location is a *recognized* location for our type-use annotations in the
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -351,7 +351,8 @@ innermost.
 > defining that there exists a series of enclosing declarations for any type
 > usage, not just for a declaration.
 
-At each declaration, check the following rules in order:
+At each declaration that is a [recognized](#recognized-declaration) location,
```
netdpb, line 354:
```suggestion
At each declaration that is in a [recognized](#recognized-declaration) location,
```

docs/docs/spec.md
```diff
@@ -369,9 +370,9 @@ If none of the enclosing declarations meet either rule, then the type usage is
 ## Augmented type of a type usage appearing in code {#augmented-type-of-usage}
 
 For most type usages in source code or bytecode on which JSpecify nullness
-annotations are [recognized], this section defines how to determine their
-[augmented types]. Note, however, that rules for specific cases below take
-precedence over the general rule here.
+annotations are [recognized](#recognized-type-use), this section defines how to
+determine their [augmented types]. Note, however, that rules for specific cases
```
netdpb, line 374: It won't let me use the "suggestion" button here, but how about:

> This section defines how to determine the [augmented types] of most type usages in source code or bytecode where JSpecify nullness annotations are [recognized].

docs/docs/spec.md
```diff
@@ -369,9 +370,9 @@ If none of the enclosing declarations meet either rule, then the type usage is
 ## Augmented type of a type usage appearing in code {#augmented-type-of-usage}
 
 For most type usages in source code or bytecode on which JSpecify nullness
```
netdpb, line 372:
```suggestion
For most type usages in source code or bytecode where JSpecify nullness
```

netdpb (approved):

docs/docs/spec.md
cpovirk: Hmm, I think of the declaration itself as _being_ the recognized location: A method is a recognized location for a `@Target(METHOD)` annotation, etc.

I'll merge this PR as it stands now, but we can continue discussing if you'd like.

PR #664, opened 2024-09-25
Adjust motivating nullness questions

wmdietl:
The first question about `@Nullable` is subsumed by the second question about all `nullness operators`.
If that is the case, I would only ask the general question.
If that is a mis-understanding on my side, can the questions be clarified?

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -23,9 +23,10 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  What is the [nullness operator] of that type usage?
-2.  Is it reasonable to assume that `null` will not come "out" of it?
-3.  Is it reasonable to assume that `null` cannot be put "in" to it?
+1.  What is the [augmented type] of that type usage?
+2. Do I have to handle the case where null comes out of it?
+3. Do I have to prevent null from going into it?
+4. Is this type a subtype of that type with respect to nullness?
 
 ### The scope of this spec
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -29,10 +29,9 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  Does `@Nullable` appear directly on that type usage?
-2.  What is the [nullness operator] of that type usage?
-3.  Is it reasonable to assume that `null` won't come "out" of it?
-4.  Is it reasonable to assume that `null` can't be put "in" to it?
+1.  What is the [nullness operator] of that type usage?
+2.  Is it reasonable to assume that `null` won't come "out" of it?
+3.  Is it reasonable to assume that `null` can't be put "in" to it?
 
 ### The scope of this spec
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -23,7 +23,7 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  What is the [augmented type] of that type usage?
+1. What is the [augmented type] of that type usage?
 2. Do I have to handle the case where null comes out of it?
 3. Do I have to prevent null from going into it?
 4. Is this type a subtype of that type with respect to nullness?
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -24,8 +24,8 @@ draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
 1. What is the [augmented type] of that type usage?
-2. Do I have to handle the case where null comes out of it?
-3. Do I have to prevent null from going into it?
+2. Do I have to handle the case where `null` comes out of it?
+3. Do I have to prevent `null` from going into it?
 4. Is this type a subtype of that type with respect to nullness?
 
 ### The scope of this spec


cpovirk (commented):
There is at least one way to get a type with nullness operator `UNION_NULL` without using `@Nullable`, and that's by writing `Foo<?>`: The wildcard's upper bound is `Object UNION_NULL`. Now that's kind of a weird case because the the upper bound is implicit. (And, really, we'd mostly prefer to think about it as not having an upper bound at all.)

There may be other such ways:
- We're about to [treat record `equals` parameters as implicitly `UNION_NULL`](https://github.com/jspecify/jspecify/pull/645#pullrequestreview-2328629207).
- We may follow the Checker Framework model of [considering local variables to be implicitly `UNION_NULL`](https://github.com/jspecify/jspecify/issues/661).

Backing up from all that, though: I could believe that this section is not really accomplishing what I set out to accomplish. I could also believe that we could accomplish the goal by merely saying that there are "at least _three_ kinds of questions" we can ask, which still leaves open the possibility that we're glossing over this unusual case about `@Nullable` annotations. Maybe that would be the best way forward?

docs/docs/spec.md
```diff
@@ -29,10 +29,9 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  Does `@Nullable` appear directly on that type usage?
-2.  What is the [nullness operator] of that type usage?
-3.  Is it reasonable to assume that `null` won't come "out" of it?
-4.  Is it reasonable to assume that `null` can't be put "in" to it?
+1.  What is the [nullness operator] of that type usage?
+2.  Is it reasonable to assume that `null` won't come "out" of it?
+3.  Is it reasonable to assume that `null` can't be put "in" to it?
```
netdpb, line 34: You're going to have to change a word on line 29. 😆

netdpb:
> Backing up from all that, though: I could believe that this section is not really accomplishing what I set out to accomplish.

What in fact did you set out to accomplish here?

wmdietl:
Sorry, I should have done this against the 'specification-1.0' branch.
As this PR maybe started a larger discussion on this section, feel free to close this PR when you open a PR with your own way to address this.

If there are things that are `UNION_NULL` without using `@Nullable`, it seems better to leave out the first question.

netdpb:
I feel like the questions that people (tools) want to ask about type usages are:

1. Do I have to handle the case where `null` comes out of it?
2. Do I have to prevent `null` values from going into it?
3. Is this type a subtype of that type with respect to nullness?

The way they answer those questions using the spec involves determining the augmented type based on annotations and substitution rules and, for question 3, the subtyping rules.

The augmented type is a means to answering those questions; it isn't an independent question.

cpovirk:
> What in fact did you set out to accomplish here?

Mainly to bash on the word "nullable" :) I'd have to dig up links to verify, but I'm pretty confident that I've seen various projects that perform checks that could be described as `isNullable` checks, and they mean various different things. I _think_ have additionally seen this cause trouble or at least confusion, like when a type parameter `<T extends @Nullable Object>` is counted as "nullable" but that doesn't mean that you can actually put a `null` into a field of type `T`.

RE: [new netdpb post](https://github.com/jspecify/jspecify/pull/664#issuecomment-2374470757):

Yes, that looks like a good overview to me. I definitely want to avoid implying that these questions are independent; they're all related, and it's just important to identify which question you're trying to answer at any given point. (I think I may have previously had a statement about how the answers to the later questions are derived from the answers to the earlier questions, but clearly we don't know.)

I would be fine with just ripping this section out. I think I had previously said that it might fit better in a separate doc that would at most be linked from here.

docs/docs/spec.md
```diff
@@ -23,10 +23,9 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  Does `@Nullable` appear directly on that type usage?
-2.  What is the [nullness operator] of that type usage?
-3.  Is it reasonable to assume that `null` will not come "out" of it?
-4.  Is it reasonable to assume that `null` cannot be put "in" to it?
+1.  What is the [nullness operator] of that type usage?
+2.  Is it reasonable to assume that `null` will not come "out" of it?
+3.  Is it reasonable to assume that `null` cannot be put "in" to it?
```
wmdietl, line 28:
```suggestion
1.  What is the [augmented type] of that type usage?
2. Do I have to handle the case where null comes out of it?
3. Do I have to prevent null from going into it?
4. Is this type a subtype of that type with respect to nullness?
```

wmdietl:
Keeping the motivation for the spec at the beginning of the doc seems important to me. The questions can ground the further discussion.

I changed the first point to augmented types. For some tools, e.g. an IDE, it could be nice to show the augmented type in e.g hover notifications.

docs/docs/spec.md
```diff
@@ -23,10 +23,10 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  Does `@Nullable` appear directly on that type usage?
-2.  What is the [nullness operator] of that type usage?
-3.  Is it reasonable to assume that `null` will not come "out" of it?
-4.  Is it reasonable to assume that `null` cannot be put "in" to it?
+1.  What is the [augmented type] of that type usage?
```
wmdietl, line 26:
```suggestion
1. What is the [augmented type] of that type usage?
```

docs/docs/spec.md
```diff
@@ -23,10 +23,10 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
 draw some distinctions, creating at least four kinds of "Is it nullable?"
 questions we can ask for any given type usage:
 
-1.  Does `@Nullable` appear directly on that type usage?
-2.  What is the [nullness operator] of that type usage?
-3.  Is it reasonable to assume that `null` will not come "out" of it?
-4.  Is it reasonable to assume that `null` cannot be put "in" to it?
+1. What is the [augmented type] of that type usage?
+2. Do I have to handle the case where null comes out of it?
+3. Do I have to prevent null from going into it?
```
wmdietl, line 28:
```suggestion
2. Do I have to handle the case where `null` comes out of it?
3. Do I have to prevent `null` from going into it?
```

docs/docs/spec.md
```diff
@@ -23,10 +23,10 @@ In this doc, we aim not to refer to whether a type "is nullable." Instead, we
...
 questions we can ask for any given type usage:
 
-1.  Does `@Nullable` appear directly on that type usage?
-2.  What is the [nullness operator] of that type usage?
-3.  Is it reasonable to assume that `null` will not come "out" of it?
-4.  Is it reasonable to assume that `null` cannot be put "in" to it?
+1. What is the [augmented type] of that type usage?
+2. Do I have to handle the case where `null` comes out of it?
+3. Do I have to prevent `null` from going into it?
+4. Is this type a subtype of that type with respect to nullness?
```
cpovirk, line 29: I could see an argument that this one doesn't fit as neatly into a list of "Is it nullable?" questions. But I think it's close enough: It's kind of about whether one type is "more nullable" than the other. So I think it works.

cpovirk (approved):

docs/docs/spec.md
netdpb: You could phrase it as "Is this type no more nullable than that type, so that it can be a subtype?"

netdpb (approved):

PR #666, opened 2024-09-25
Add reference to Java version

wmdietl:
I would have expected to see a statement on the supported Java version.
Earlier, there is the sentence "Or the tool might target a future version of Java whose language features would not be covered by this version of this spec."
But there is no explicit statement of what version is supported by this spec.

Maybe also something on future versions should be added? Otherwise, whenever there is a new JLS, we need a corresponding JSpecify spec update.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -94,6 +94,8 @@ definition.
 In particular, when a JLS rule refers to types, apply this spec's definition of
 [augmented types] \(as opposed to [base types]).
 
+This specification covers all JLS constructs up to Java SE 22.
+
 ## Base type
 
 A *base type* is a type as defined in [JLS 4].
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -94,7 +94,7 @@ definition.
 In particular, when a JLS rule refers to types, apply this spec's definition of
 [augmented types] \(as opposed to [base types]).
 
-This specification covers all JLS constructs up to Java SE 22.
+This specification covers all JLS constructs up to [Java SE 22].
 
 ## Base type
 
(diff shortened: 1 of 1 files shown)


cpovirk (approved):
Thanks. I expected to feel somewhat nervous about whether we really got everything in Java 22, but I think I actually feel OK: We've covered some things about records, and while we know that we might want to think more about [annotations in `instanceof`](https://github.com/jspecify/jspecify/issues/626), we do take _some position_ on them.

Plus, if anyone sees something that is missing in Java 22 that we're _not_ aware of, we'd like that person to bring it to our attention.

(We of course don't try to get into implementation code much at all. That saves us from some work that we'd otherwise need for Java 22.)

As for talking about the future... I'm not sure what's best, nor really even how carefully we're going to version the spec.

wmdietl:
I used Java 22, because all the `JLS` links used that version.

Some statement about evolution would be good. We either need to make a JSpecify spec every six months to keep up with JLS releases or the JSpecify spec will start to look stale.

cpovirk:
Yeah, I almost bumped all the links to 23 as part of https://github.com/jspecify/jspecify/pull/663.... As far as I know, we could do that if we wanted.

Well, or at least 23 is probably no worse than 22 :) I remembered part of what I'd been nervous about with claiming support for newer JDKs: pattern matching, which I covered only incompletely earlier. I've mailed you https://github.com/jspecify/jspecify/pull/668. Maybe _then_ we're good for 22/23?

I think it's probably reasonable for us to have to consider each new Java version and then bump the spec from "Java 23" to "Java 24," etc. I don't know how strong a statement to make, but something about how we "intend" to update it for new language features would seem uncontroversial but hopefully still helpful?

PR #667, opened 2024-09-25
Comment on "operator"

wmdietl:
This is not a real PR, but a comment on a specific part of the specification.
Let me know if you prefer using issues for this....
If you agree this is an issue, we can use the PR to come up with a solution.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -122,6 +122,14 @@ A nullness operator is one of four values:
 -   `UNSPECIFIED`
 -   `MINUS_NULL`
 
+TODO: This is the first appearance of "operator" and there is no formal definition
+of what they are used for and what they mean.
+
+The "values" is maybe also misleading - these are type operators, not runtime values.
+
+This would also be a good place to introduce the postfix notation and contrast it
+to the prefix notation used by annotations.
+
 > The informal meaning of the operators is:
 >
 > -   `UNION_NULL`: This is the operator produced by putting `@Nullable` on a


docs/docs/spec.md
```diff
@@ -122,6 +122,14 @@ A nullness operator is one of four values:
 -   `UNSPECIFIED`
 -   `MINUS_NULL`
 
+TODO: This is the first appearance of "operator" and there is no formal definition
```
cpovirk, line 125: I've never been sure what the best way to order the sections of the spec is.... I want to introduce "nullness operators" early so that I can define "augmented types" in terms of them. But then until I define that latter term, the former definition is just kind of hanging there.

As a formal definition, I want to claim something similar: The "formal definition" of a nullness operator is either:
- that it's one of the four operators listed here, or...
- that it's defined by its effect on substitution and subtyping rules, as discussed later.

I'm not sure that a formal definition could add anything beyond that, though maybe we could foreshadow that those rules are coming?

Basically, I think that it would be valuable to give some _non-normative, informal_ definition, since we currently say _nothing_ about this fundamental concept except for discussing the four operators individually later on.

Or maybe we can get away with a formal definition along the lines of the following...?

"A nullness operator is the main way that JSpecify represents information about the nullness of a given type. There are four nullness operators:"

docs/docs/spec.md
```diff
@@ -122,6 +122,14 @@ A nullness operator is one of four values:
...
+of what they are used for and what they mean.
+
+The "values" is maybe also misleading - these are type operators, not runtime values.
```
cpovirk, line 128: Makes sense. I've taken a shot at this in my comment above.

docs/docs/spec.md
```diff
@@ -122,6 +122,14 @@ A nullness operator is one of four values:
...
+
+This would also be a good place to introduce the postfix notation and contrast it
```
cpovirk, line 130: Hmm, yes, I see that we use the notation in the non-normative text below, and we haven't defined it yet... and in fact don't do so _ever_? It looks like we might use the notation only in non-normative text, but that's still not great.

We should at least define it under "augmented types," and then maybe the non-normative comment here can refer readers ahead to that section. I could see defining it here instead, but users are going to have to understand the basics of augmented types before the section on "the informal meaning" below makes any sense, anyway. So we might as well define it under "augmented types," since it's the notation for, well, augmented types :)

(This is another manifestation of the problem of trying to define "nullness operator" before defining "augmented type.")

PR #668, opened 2024-09-25
Deem annotations in *all* patterns to be unrecognized.

cpovirk:
And more clearly distinguish between:
- `instanceof Foo`, the "type comparison operator"
- `instanceof Foo foo`, the "pattern match operator"

(per
https://docs.oracle.com/javase/specs/jls/se22/html/jls-15.html#jls-15.20.2)

This prepares us for:

- possibly someday recognizing annotations in the pattern match operator
  while likely still not recognizing them in the type comparison
  operator, as touched on in
  https://github.com/jspecify/jspecify/issues/626

- claiming to cover all of Java 22, as planned in
  https://github.com/jspecify/jspecify/pull/666

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -252,10 +252,12 @@ However, the type-use annotation is unrecognized in any of the following cases:
 
 -   type arguments of a receiver parameter's type
 
--   any component of the type after the `instanceof` operator
+-   any component of the type after the `instanceof`
+    [type comparison operator][JLS 15.20.2]
 
-    > We may revisit this rule in the future in light of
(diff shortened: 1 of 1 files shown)


cpovirk:
To give a specific example of an annotation in a pattern other than in `instanceof`:

```java
  int x(Foo<String> foo) {
    return switch (foo) {
      case Bar<@Nullable String> bar -> 0;
      case Baz<String> baz -> 1;
    };
  }
```

wmdietl (approved):
Looks good. I'll separately comment on #626.

PR #573, opened 2024-07-17
Update documentation for 1.0.0

netdpb:
Please do not make any changes to this branch except via pull request onto the branch.

PR #669, opened 2024-09-25
Replay specification-1.0 commits onto main branch.

cpovirk:
(DO NOT MERGE through GitHub? We want some kind of rebase, rather than a merge, I assume similar to what we did in https://github.com/jspecify/jspecify/pull/565 / https://github.com/jspecify/jspecify/pull/573.)

I claim that we should do this now, even if we're going to wait a little bit longer for final reviews before 1.0:

- We are telling people to "review the spec" if they're interested. They're nearly certain to try to read https://jspecify.dev/docs/spec instead of https://github.com/jspecify/jspecify/blob/specification-1.0/docs/docs/spec.md.
- We had wanted a separate branch in order to avoid getting caught with half of the spec reflecting decisions from 0.3 and half of the spec reflecting decisions from 1.0. We have since completed the updates for decisions changed since 0.3 (`Unmarked<@Nullable>` substitution, recognized annotation locations, new annotations). While we will merge some more changes between now and spec 1.0, they reflect other sorts of improvements, not changes in design decisions.
- Whenever we are actively working on two branches, there is overhead and opportunity for merge conflicts.

My plan after "merging" this will be to delete the specification-1.0 branch and see if we can reparent all pending PRs. The sooner we do this, the fewer such PRs we'll have, I hope :)

FYI @wmdietl @kevinb9n

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -823,7 +823,7 @@ hold:
 > Thus, the rules here are restricted to type variables and intersection types,
 > whose supertypes may have nullness annotations.
 
-`T` has nullness-subtype-establishing direct-supertype edges to the all
+`T` has nullness-subtype-establishing direct-supertype edges to all the
 following types, subject to the exception given below:
 
 -   if `T` is an augmented [intersection type]: all the intersection type's
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -405,6 +405,14 @@ condition is met, skip the remaining conditions.
     > If the type usage is annotated with both `@Nullable` and `@NonNull`, these
     > rules behave as if neither annotation is present.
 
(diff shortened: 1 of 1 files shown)


wmdietl (approved):
I'm fine with merging this. I still have a bunch of comments that I'll turn into PRs or issues.

netdpb (approved):
I'm happy to work with you on making this rebase happen.

cpovirk:
Thanks. I can check with you when we're both in tomorrow, but... do you remember if you temporarily enabled "Rebase and merge" and used that or if you did something at the command line? If the latter, do you know if you first rebased onto `main` so that the result could be a fast-forward that autoclosed the PR and otherwise looked neat and tidy?

cpovirk:
I'm going to open a separate PR with what I think is the approach you were describing, and we can work on that later this morning.

PR #673, opened 2024-09-26
Remove "draft" from the spec title.

cpovirk:
We may end up waiting a bit for final reviews, but let's queue this up.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -2,7 +2,7 @@
 sidebar_position: 6
 ---
 
-# Nullness Specification (draft)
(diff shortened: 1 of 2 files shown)


netdpb (approved):

cpovirk:
I've made a few other changes. I figure that I'll merge this PR after https://github.com/jspecify/jspecify/pull/603, since the bug fixed by that PR is the reason that I wouldn't yet consider this a "release candidate" in the sense of "There is a chance that we could ship this unmodified as 1.0.0."

image

Issue #580, opened 2024-07-27
Document what `@Nullable` annotations on varargs mean?


msridhar:
I was going to add some text like this around [here](https://github.com/jspecify/jspecify/blob/c40e5fa5346883760753bb6a3231d1afac95216c/docs/docs/user-guide.md?plain=1#L472) regarding varargs, something like:

```
    A related corner case is a varargs argument declaration, e.g.,
    `foo(Object... args)`.  To say that each individual argument passed in a
    varargs position can be `null`, write `foo(@Nullable Object... args)`.  To
    say that the array holding the varargs at a call site may be `null` (e.g., to
    allow `Object[] x = null; foo(x);`), write `foo(Object @Nullable... args)`.
```

But then I thought this is too much of a corner case to put in the user guide.  But should we have something like this somewhere in the docs?  Or do we already have it?

agentgt:
I think it is worthwhile to add a note like this at least to show the placement and to show how yes the array passed in can be null (I would wager a lot of java devs just assume the var args array is not going to be null ever).

kevinb9n:
Indeed I missed mentioning this in the javadoc; will fix.

netdpb:
Does this mean that within `@NullMarked`, implementations can in fact assume that the array isn't going to be `null` (when coming from null-checked code)? That's actually pretty nice.

kevinb9n:
Yup.

Issue #550, opened 2024-07-09
Doc briefly explaining the value of type-use annotations and how to use them


cpovirk:
@msridhar dug up https://www.oracle.com/technical-resources/articles/java/ma14-architect-annotations.html, and @netdpb pointed out that we have [a glossary entry](https://docs.google.com/document/d/1nbTnJ0-HubLnQPKSjK5CDZyoe6Al64vdlkoJxaYX9XY/preview?resourcekey=0-ADjPZnp8LN3dRX_ptjlagw#bookmark=id.e1wg9ys7ch8l), which eventually leads to http://github.com/jspecify/jspecify/wiki/type-usages.

The niche I'm looking to fill is "Why should I use these JSpecify 'type-use annotations' instead of the 'declaration annotations' I'm currently familiar with?" That includes giving a brief explanation of what makes an annotation a type-use annotation (i.e., `@Target(TYPE_USE)`), which common annotations are type-use annotations, maybe something about "hybrid" type-use/declaration annotations, and then an explanation of what type-use annotations unlock in the context of nullness (currently, annotating type arguments, type-parameter bounds, and array-element types). It could also get into how to annotate [arrays](https://checkerframework.org/manual/#writing-annotations) and [qualified types](https://checkerframework.org/manual/#faq-annotate-fully-qualified-name).

This could connect to https://github.com/jspecify/jspecify/issues/123.

cpovirk:
Such a doc should also be [another](https://jspecify.dev/docs/whether/#annotation-processors) place where we call out the problems for annotation processors (notably Dagger) before JDK 22 finally fixed javac's handling of type-use annotations.

It's also surely the case that some _other_ annotation processors / static analyzers lack support for type-use annotations simply because those tools haven't implemented it. (We could even outline what such tools need to do to add support.) This should be less the case over time, and it's likely already less of an issue that the javac bug. But if we're writing a whole doc about type-use annotations, then that will be a good place to call this out.

cpovirk:
Sorry, I had forgotten until a question today that we already have https://jspecify.dev/docs/user-guide/#type-use-annotation-syntax that we can link to for that last part. That leaves only the part about "the value of type-use annotations."

cpovirk:
Given that we've heard a few times about problems with type-use annotations with incomplete classpaths (beyond the problems that incomplete classpaths in general, even when using _declaration_ annotations), we should acknowledge that somewhere, too.

PR #674, opened 2024-09-26
Touch on the syntax for annotating arrays.

cpovirk:
This builds on the work of
https://github.com/jspecify/jspecify/pull/592 in sorta kinda partially
addressing https://github.com/jspecify/jspecify/issues/580, though that
request is more about the user guide.

We do want to save most of this discussion for the user guide and
perhaps [other docs](https://github.com/jspecify/jspecify/issues/550),
but we also want to make clear that this kind of annotating is
_supported_, especially since we already call out the similar case of
varargs.

(@sdeleuze)

This PR replaces https://github.com/jspecify/jspecify/pull/670.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -240,18 +240,18 @@ exceptions in the subsequent sections:
 
 -   an array component type
 
-    > An array of nullable strings is written `@Nullable String[]`. Similarly, a
-    > variadic parameter whose type is "array of nullable strings" is written
+    > For an array of nullable strings, write `@Nullable String[]`. Similarly,
+    > for a variadic parameter whose type is "array of nullable strings," write
     > `@Nullable String...`.
     >
     > You can annotate array component types independently from the array
-    > itself. The array itself can be annotated in the same cases as any
-    > non-array type in the same position can, albeit with different syntax. For
-    > example, a method parameter could be annotated as `@NonNull String
-    > @Nullable [] strings`, which means `strings` is a nullable array
-    > containing non-null elements. Similarly for variadic parameters, `void
-    > method(@Nullable String @NonNull ... strings)` means `strings` is a
-    > non-null array containing nullable elements.
+    > itself. You can annotate the array itself in the same cases as any
+    > non-array type in the same position, albeit with different syntax. For
+    > example, you can annotate a method parameter as `@NonNull String @Nullable
+    > [] strings`, which means `strings` is a nullable array containing non-null
+    > elements. Similarly for variadic parameters, `void method(@Nullable String
+    > @NonNull ... strings)` means `strings` is a non-null array containing
+    > nullable elements.
 
 -   an array creation expression
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -225,13 +225,8 @@ exceptions in the subsequent sections:
     [JLS 8.4.1]
 
     > This excludes the receiver parameter but includes variadic parameters (in
-    > varargs methods). Because *array component* types are also a recognized
-    > location for annotations, variadic parameters (which are arrays) can have
-    > annotations on the array itself and/or its component type. For example,
-    > `@NonNull String @Nullable [] strings` means `strings` is a nullable array
-    > containing non-null elements. Similarly, `void method(@Nullable String
-    > @NonNull ... strings)` means `strings` is a non-null array containing
-    > nullable elements.
+    > varargs methods). For examples of variadic paramters, see the comment
+    > about array components below.
 
 -   a field type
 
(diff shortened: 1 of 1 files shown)


sdeleuze:
Thanks @cpovirk, those proposed changes brings a much more clarity for implementors on how arrays and varargs are supported.

docs/docs/spec.md
```diff
@@ -224,10 +224,13 @@ exceptions in the subsequent sections:
 -   a formal parameter type of a method or constructor, as defined in
     [JLS 8.4.1]
 
-    > This excludes the receiver parameter but includes variadic parameters.
-    > Specifically, you can add `@Nullable` before the `...` token to indicate
-    > that a variadic method accepts `null` arrays: `void foo(String @Nullable
-    > ... strings)`.
+    > This excludes the receiver parameter but includes variadic parameters (in
+    > varargs methods). Because *components* types are also a recognized
```
netdpb, line 228:
```suggestion
    > varargs methods). Because *array component* types are also a recognized
```

docs/docs/spec.md
```diff
@@ -224,10 +224,13 @@ exceptions in the subsequent sections:
...
+    > location for annotations, variadic paramters (like arrays) can have
+    > annotations on multiple components. For example, `@NonNull String
```
netdpb, line 230:
```suggestion
    > annotations on the array itself and/or its component type. For example, `@NonNull String
```

docs/docs/spec.md
```diff
@@ -224,10 +224,13 @@ exceptions in the subsequent sections:
...
+    > @Nullable [] strings` means non-null elements in a nullable array.
```
netdpb, line 231:
```suggestion
    > @Nullable [] strings` means `strings` is a nullable array containing non-null elements.
```

docs/docs/spec.md
```diff
@@ -224,10 +224,13 @@ exceptions in the subsequent sections:
...
+    > Similarly, `void method(@Nullable String @NonNull ... strings)` means
+    > nullable elements in a non-null array.
```
netdpb, line 233:
```suggestion
    > `strings` is a non-null array containing nullable elements.
```

docs/docs/spec.md
```diff
@@ -241,6 +244,11 @@ exceptions in the subsequent sections:
 
 -   an array component type
 
+    > An array of nullable strings is written `@Nullable String[]`. A variadic
+    > parameter whose type is "array of nullable strings" is written `@Nullable
+    > String...`. Compare this to the syntax for a nullable array of strings,
+    > shown above in the discussion of variadic parameters.
```
netdpb, line 250: It's too bad that we're putting variadic parameters first, and that we have to split these up at all.

Could the variadic note above simply point to this paragraph, and put all the examples here?

cpovirk: Yes, that sounds better. Let me know what you think of this attempt.

cpovirk (commented):
Thanks.

docs/docs/spec.md
```diff
@@ -244,10 +240,17 @@ exceptions in the subsequent sections:
...
-    > An array of nullable strings is written `@Nullable String[]`. A variadic
-    > parameter whose type is "array of nullable strings" is written `@Nullable
-    > String...`. Compare this to the syntax for a nullable array of strings,
-    > shown above in the discussion of variadic parameters.
+    > An array of nullable strings is written `@Nullable String[]`. Similarly, a
+    > variadic parameter whose type is "array of nullable strings" is written
+    > `@Nullable String...`.
+    >
+    > Additionally, whenever JSpecify recognizes an annotation at a given
+    > location, it also recognizes an annotation on *an array itself* at that
```
netdpb, line 248: This confused me. I thought you were saying that whenever an annotation can go on an array's component type it can also go on the array itself, which is not true...and then I realized it wasn't what you were saying.

How can we clarify? Maybe start with "When JSpecify recognizes a type-use annotation on an array itself (because it is the type of a field, parameter, return type, etc.), that annotation comes directly before the `[]`, `[dimension]`, or `...` token. For example...."

cpovirk: I had a flash of insight—or maybe just the _feeling_ that I'd had a flash of insight. You be the judge :)

docs/docs/spec.md
```diff
@@ -244,13 +244,14 @@ exceptions in the subsequent sections:
...
     >
-    > Additionally, whenever JSpecify recognizes an annotation at a given
-    > location, it also recognizes an annotation on *an array itself* at that
-    > location. For example, a method parameter could be annotated as `@NonNull
-    > String @Nullable [] strings`, which means `strings` is a nullable array
-    > containing non-null elements. Similarly, `void method(@Nullable String
-    > @NonNull ... strings)` means `strings` is a non-null array containing
-    > nullable elements.
+    > The ability to annotate array component types is orthogonal to the ability
+    > to annotate the array itself. The array itself can be annotated in the
```
netdpb, line 248:
```suggestion
    > You can annotate array component types independently from
    > the array itself. The array itself can be annotated in the
```

docs/docs/spec.md
```diff
@@ -244,13 +244,14 @@ exceptions in the subsequent sections:
...
+    > same cases as any other type in the same position can, albeit with
```
netdpb, line 249:
```suggestion
    > same cases as any non-array type in the same position can, albeit with
```

docs/docs/spec.md
```diff
@@ -244,13 +244,14 @@ exceptions in the subsequent sections:
...
+    > different syntax. For example, a method parameter could be annotated as
+    > `@NonNull String @Nullable [] strings`, which means `strings` is a
+    > nullable array containing non-null elements. Similarly, `void
```
netdpb, line 252:
```suggestion
    > nullable array containing non-null elements. Similarly for variadic parameters, `void
```

netdpb (approved):

cpovirk:
Sigh, somehow it did not dawn on me that @wmdietl and I were editing the exact same section in separate PRs and with slightly different goals :( I have now attempted to reconcile the two, but this deserves more review from one or both of you.

netdpb (approved):

PR #675, opened 2024-09-27
Remove the warning about IntelliJ and local variables.

cpovirk:
The issue I was referring to was fixed in IDEA 2024.2.1:
https://youtrack.jetbrains.com/issue/IDEA-348069

(Thanks to @amaembo for asking about this
https://youtrack.jetbrains.com/issue/IDEA-355699/Warning-when-annotating-array-elements-as-nullable-with-JSpecify#focus=Comments-27-10743536.0-0)

### DIFF
--- a/docs/docs/whether.md
+++ b/docs/docs/whether.md
@@ -52,8 +52,7 @@ they claim about their current and planned JSpecify support.
     but does not yet analyze generics.
 
 *   [IntelliJ IDEA](https://www.jetbrains.com/idea/) supports JSpecify
-    annotations but has incomplete support for generics and some trouble with
-    local variables.
+    annotations but has incomplete support for generics.
 
 *   The [Checker Framework](https://checkerframework.org/) understands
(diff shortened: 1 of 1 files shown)


netdpb (approved):
Is there a YouTrack link for "incomplete support for generics" we should add?

amaembo:
I know that we don't support every possible case with generics, so it would probably always be 'incomplete'. However, if you see some really annoying problems, please report them. I've got [one report](https://youtrack.jetbrains.com/issue/IDEA-356144) today, will take a look at it.

cpovirk:
I cannot for the life of me find the doc I wrote about IntelliJ support for JSpecify years ago :\

[IDEA-356144](https://youtrack.jetbrains.com/issue/IDEA-356144) does sound like the main thing that I remember. I will file actual issues when I play around with IntelliJ and our nullness annotations again, and I'll continue to encourage other people to do so. In the meantime, I've added a link to it as an example.

PR #678, opened 2024-09-27
Array creation is intrinsically non-null

wmdietl:
A newly created array, like a newly created object, is non-null. So creating a nullable array should make no sense, like creating a nullable object makes no sense.

Or am I missing what case this is covering?

I also added some non-normative examples.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -244,6 +244,12 @@ exceptions in the subsequent sections:
 
 -   an array component type
 
+    > Note that this rule recognizes the annotations in `@Nullable String []`
+    > and `String [] @Nullable []`.
+    > Whether the annotation in `String @Nullable []` is recognized depends on
+    > where that type is used - it would e.g. be recognized if that type is
+    > used as a field type.
+
 However, the type-use annotation is unrecognized in any of the following cases:
 
 -   a type usage of a primitive type, since those are intrinsically non-nullable
@@ -299,8 +305,8 @@ All locations that are not explicitly listed as recognized are unrecognized.
 >
 >         > For example, `new String @Nullable [5]` has an unrecognized
 >         > annotation. However, note that the component type in an array
->         > creation expression can be annotated. For example, `new int [5]
->         > @Nullable []` has a recognized annotation.
+>         > creation expression can be annotated. For example, `new @Nullable
+>         > String [5]` has a recognized annotation.
 >
 >     -   outer type qualifying an inner type
 >
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -288,18 +288,16 @@ All locations that are not explicitly listed as recognized are unrecognized.
 >     -   object creation expression
 >
 >         > For example, `new @Nullable ArrayList<String>()` has an unrecognized
->         > annotation.
->         > However, note that type arguments in an object creation expression can
->         > be annotated.
->         > For example, `new ArrayList<@Nullable String>()` has a recognized
->         > annotation.
+>         > annotation. However, note that type arguments in an object creation
+>         > expression can be annotated. For example, `new ArrayList<@Nullable
+>         > String>()` has a recognized annotation.
 >
 >     -   array creation expression
 >
->         > For example, `new String @Nullable [5]` has an unrecognized annotation.
->         > However, note that the component type in an array creation expression
->         > can be annotated.
->         > For example, `new int [5] @Nullable []` has a recognized annotation.
+>         > For example, `new String @Nullable [5]` has an unrecognized
+>         > annotation. However, note that the component type in an array
+>         > creation expression can be annotated. For example, `new int [5]
+>         > @Nullable []` has a recognized annotation.
 >
 >     -   outer type qualifying an inner type
 >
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -241,8 +241,6 @@ exceptions in the subsequent sections:
 
 -   an array component type
```
wmdietl, line 242: I was originally worried about what rule covers the root array type. But that is covered by e.g. a parameter type. So in `void foo(int @Nullable [] p)` the annotation is recognized because it is on the parameter.

cpovirk: I guess that it is a little unfortunate that we use the term ["type component"](https://github.com/jspecify/jspecify/wiki/type-usages#type-component) and then _also_ speak of "array component types." But yeah, I think you're right that this rule is intended to refer only to the types of the elements of the array.

docs/docs/spec.md
```diff
@@ -290,7 +288,16 @@ All locations that are not explicitly listed as recognized are unrecognized.
...
+>         > annotation. However, note that type arguments in an object creation
+>         > expression can be annotated. For example, `new ArrayList<@Nullable
+>         > String>()` has a recognized annotation.
+>
+>     -   array creation expression
+>
+>         > For example, `new String @Nullable [5]` has an unrecognized
+>         > annotation. However, note that the component type in an array
+>         > creation expression can be annotated. For example, `new int [5]
+>         > @Nullable []` has a recognized annotation.
```
cpovirk, line 300: I would be tempted to keep this example to a one-dimensional array because of how confusing the syntax gets. Do you think ghat `new @Nullable String[5]` could still get the point across while being a little simpler and closer to the earlier "bad" example?

cpovirk (commented):
Thanks! This seems like either a simple mistake or maybe an attempt to say that "array components types" are things that you see in type "declarations" while `new @Nullable Foo[5]` is different (and so requires its own bullet point) because it's not _just_ a Java type but also some information about the array size. Even if it's the latter, I think it would still qualify as a "mistake" that we don't restrict it to the component types.

Maybe @netdpb has thoughts on what we need to accomplish here?

---

RE: my autoformatting:

@netdpb recently sent me https://sembr.org/, which sounds like the kind of thing you may have been going for here. I see the value there, which sadly has to compete against the value of autoformatting :(

I'm not opposed to using more vertical space in the rendered text (by leaving "blank" lines with just `>`s on them) in order to keep things separate if you'd like that.

docs/docs/spec.md
netdpb: I agree.

Although it may be useful to add multidimensional array examples to the array component type entry, if only because the syntax is so odd. I wouldn't add it here.

wmdietl:
Sorry for the delay. I'll take a pass through my open spec PRs tonight.

wmdietl:
I expanded the examples, hopefully as your comments suggested.

Re "sembr": is there a command that you use to reformat the file? I do like sembr, although I hadn't heard of that term for this style.
I agree that reformatting is also important and can run some command to do that after my edits.

cpovirk (approved):
Pardon my directly pushing a few changes that I'd normally leave as suggested edits. I figured that, if I was going to reformat the file, I might as well make the changes first so as to avoid another reformatting pass afterward. (In hindsight, I could have instead left the whole batch as a suggested edit.)

I will acknowledge here that `String[] @Nullable []` looks a little weird to my eyes compared to your `String [] @Nullable []`, which feels a bit more principled in inserting spaces everywhere that we could. Then again, I'm still not sure I'd extend that principle to `String [] @Nullable [] []` for a 3D array (vs. `String[] @Nullable [][]`), so I might not have as principled a stance as I'd like to think :) Anyway, I'm following google-java-format, so if you think we should revisit that, you might file a bug there.

As best I can tell, the `mdformat` tool [that netdpb and I have been using](https://github.com/jspecify/jspecify/pull/636#issuecomment-2358874200) is different from any publicly available tool under that name :\

docs/docs/spec.md
```diff
@@ -244,7 +244,10 @@ exceptions in the subsequent sections:
 
 -   an array component type
 
--   an array creation expression
+    > Note that this rule recognizes the annotations in `@Nullable String []`
```
wmdietl, line 247:
```suggestion
    > Note that this rule recognizes the annotations in `@Nullable String[]`
```

Should that space be removed, to be consistent with the other `String[` occurrences?
Following what GJF does sounds okay to me, but this now looks inconsistent.

wmdietl (commented):
Thanks for reformatting and your tweaks! They all look fine to me.

Using some open mdformat would allow us to check formatting in CI and allow all of us to reformat the document before pushing. I have no experience with open mdformat tools, so I can't suggest one.

I'm fine with you merging this PR once you're happy with the spacing of arrays.

cpovirk:
Thanks, yes, I missed that one. 

Maybe we can play with Markdown formatters. netdpb@ had dug up a couple, and I was initially not a fan, but I've forgotten details already. Maybe I can recover some tomorrow and file an issue about all that.

Issue #252, opened 2022-08-02
Questions about cast expressions [working decision: only type components are nullness-applicable]


kevinb9n:
For what kinds of usages within cast expressions would JSpecify annotations (`@Nullable/@NonNull`) be required/prohibited/optional?

JSpecify has a much higher priority to standardize how APIs are annotated than to do the same for implementation code. That seems to argue for leaving this tool-dependent. But, pre-1.0, it would seem smarter to at least attempt agreement first, and then back off if/when tool owners sufficiently protest.

So:

1. Can the root type of the cast type be annotated `@NonNull`? The user's goal must be to signal to an analyzer that an expression seen as nullable can't actually have a null value, so that it can be safely assigned to a non-null type (or, I suppose, dereferenced). But there seem to be plenty of good ways to do that already: a conditional, an assertion, passing it through `checkNotNull`, or suppression. (Plus a less-good-but-okay option too: passing it through a method like `uncheckedCastToNonNull` which just returns its argument.) Plus, even if it was legal, it should generate a warning anyway, by analogy with `(T) o` -- both violate the expectation that (non-widening) casts are fully validated at runtime. So little is gained by allowing it.

2. Can the root type in the cast expression be annotated `@Nullable`? As a widening cast... where would this ever be needed? Perhaps for some type inference scenario?

3. When the user is casting "anyway" (i.e. from one base type to another), should the appropriate *non-root* usages of `@Nullable` (i.e., that you expect in the resulting type) be mandatory/explicit, or optional but inferrable, or prohibited and *only* inferred? (Or heaven forbid, *ignored* and only inferred?) My first question is whether they always *can* be inferred. Mandatory/explicit also helps preserve the general notion that "string means string".

3b. If 3 is answered mandatory or optional, how are these to be validated?

4. Would we allow casting a base type to itself, altering only the nullness of type arguments? One argument, which might be flawed (?): either the cast is widening and thus unnecessary, or it's not and thus would need to produce a warning anyway. So what has been gained by the cast? And again, casts "should" be a visual suggestion that something is being validated at runtime.

kevinb9n:
Looking for feedback from tool owners @msridhar @lazaroclapp @wmdietl @artempyanykh @akozlova. I _expect_ library owners to not care strongly about this one.

msridhar:
[...]

PR #679, opened 2024-09-30
Examples of annotation usages in cast expressions

wmdietl:
`cast` appears only once in `spec.md` before this PR.
Annotations in `instanceof` are unrecognized per #626, #668.
In #252 we decided only real sub-components of a cast expression can be annotated.

This PR contradicts what we decided in #252 and forbids annotations in both instanceof and casts.
The types in both are closely related and we should have similar rules for them.

[This comment](https://github.com/jspecify/jspecify/issues/626#issuecomment-2358825421) suggests not recognizing the instanceof usages until further discussion.
Similarly, we should re-open #252 and not recognize annotations in casts.

Or what discussion about the difference between casts and instanceofs am I forgetting? If there is a reason, a quick note in the spec would be good, as otherwise questions about cast types will linger.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -297,10 +297,9 @@ All locations that are not explicitly listed as recognized are unrecognized.
 > -   root types in the reference type(s) in a [cast expression][JSL 15.16]
 >
 >     > For example, `String s = (@NonNull String) o;` has an unrecognized
->     > annotation. However, note that type arguments in a cast expression
->     > can be annotated. For example,
->     > `ArrayList<@Nullable String> al = (ArrayList<@Nullable String>) o;`
->     > has a recognized annotation.
+>     > annotation. However, note that type arguments in a cast expression can
+>     > be annotated. For example, `ArrayList<@Nullable String> al =
+>     > (ArrayList<@Nullable String>) o;` has a recognized annotation.
 >
 > -   some additional intrinsically non-nullable locations:
 >
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -294,7 +294,7 @@ All locations that are not explicitly listed as recognized are unrecognized.
 >     > For example, `@Nullable List<String> strings = ...` or `String @Nullable
 >     > [] strings = ...` have unrecognized annotations.
 >
-> -   root types in the reference type(s) in a [cast expression][JSL 15.16]
+> -   root types in the reference type(s) in a [cast expression][JLS 15.16]
 >
 >     > For example, `String s = (@NonNull String) o;` has an unrecognized
 >     > annotation. However, note that type arguments in a cast expression can
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -294,7 +294,7 @@ All locations that are not explicitly listed as recognized are unrecognized.
 >     > For example, `@Nullable List<String> strings = ...` or `String @Nullable
 >     > [] strings = ...` have unrecognized annotations.
 >
-> -   root types in the reference type(s) in a [cast expression][JLS 15.16]
+> -   root type in a reference type in a [cast expression][JLS 15.16]
 >
 >     > For example, `String s = (@NonNull String) o;` has an unrecognized
 >     > annotation. However, note that type arguments in a cast expression can
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -269,11 +269,10 @@ However, the type-use annotation is unrecognized in any of the following cases:
 
 -   type arguments of a receiver parameter's type
 
--   any component of the reference type(s) in a
-    [cast expression][JSL 15.16]
-
 -   any component of the type after the `instanceof`
     [type comparison operator][JLS 15.20.2]
+
+    > We are likely to revisit this rule in the future.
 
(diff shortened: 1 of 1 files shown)


cpovirk:
I have basically three levels of thoughts:

1. We know that the spec is in what is probably not the right long-term state, and we mostly wanted to do _something_ to unblock 1.0. We are in that situation because `instanceof Foo<@Nullable Bar>` is a case that we didn't consider until late in the game (in part because it's a "new" Java feature), and we figured that the conservative thing to do was to not allow annotations there until we had more time to think about it. We attempt to acknowledge the current weird state with the non-normative comment after the bullet about patterns. We could put a similar acknowledgment on casts if it would help.
2. I do personally expect that we'll eventually allow annotations in patterns. This might not actually be too controversial, and we could try to resolve it if that would be the easiest way forward.
3. Even if somehow we choose not to allow annotations in patterns (again: unlikely IMO), I think we've established a good case for annotations in _casts_ in that past discussions that you've linked. What makes patterns interesting is that `instanceof` historically suggests that all parts of the given type will be checked. Contrast this to casts, where users are already used to the idea of unchecked casts.

netdpb:
> The types in both [`instanceof` and casts] are closely related and we should have similar rules for them.

But they are different. The type in an `instanceof` type comparison operation must be reifiable, but that is not true for the type in cast expressions. In fact people do (unchecked) casts to non-reifiable types all the time:

```java
if (foo instanceof List<?>) {
  @SuppressWarnings("unchecked")
  List<String> strings = (List<String>) foo;
  useStrings(strings);
}
```
You cannot put anything but a wildcard in the `instanceof List`; you definitely want to put other things in the cast.

That's what makes it necessary to treat them differently with respect to type-use annotations.

wmdietl:
> But they are different. The type in an `instanceof` type comparison operation must be reifiable, but that is not true for the type in cast expressions. In fact people do (unchecked) casts to non-reifiable types all the time:
> ...
> You cannot put anything but a wildcard in the `instanceof List`; you definitely want to put other things in the cast.

That is no longer true since Java SE 16, see the discussion at the end of the [type comparison operator](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.20.2) discussion.

As neither the annotations in the instanceof nor in the cast are checked at run time, it seems odd to treat them differently.

I have now added an example for allowed and forbidden cast annotations and added a note that this might be revisited to the instanceof rule.

cpovirk (approved):
Thanks!

We could _potentially_ still make a case that annotations in the type comparison operator `instanceof` are _unhelpful_. But I also see the point that it would be weird to be able to write something like...

```java
boolean isStringList(Iterable<@Nullable String> i) {
  return i instanceof List<String>;
}
```

...when the _lack_ of annotations isn't being checked, either.

So I would personally encourage just `instanceof List` when using the type comparison operator. But if you're going to include type arguments, then I would see value in including annotations, too, with those presumably checked by tools.

That's on top of the argument for consistency, which I am coming to appreciate more, as well.

Clearly we should at least revisit this :)

netdpb:
> > But they are different. The type in an `instanceof` type comparison operation must be reifiable, but that is not true for the type in cast expressions. In fact people do (unchecked) casts to non-reifiable types all the time:
> > ...
> > You cannot put anything but a wildcard in the `instanceof List`; you definitely want to put other things in the cast.
> 
> That is no longer true since Java SE 16, see the discussion at the end of the [type comparison operator](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.20.2) discussion.

What does "checked cast compatible" mean in [15.20.2](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.20.2)?

wmdietl:
> What does "checked cast compatible" mean in [15.20.2](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.20.2)?

That refers to [JLS 5.5](https://docs.oracle.com/javase/specs/jls/se23/html/jls-5.html#jls-5.5).
Should this specifically be called out in the spec? It should just fall out from our nullness subtyping rules.

netdpb:
> > What does "checked cast compatible" mean in [15.20.2](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.20.2)?
> 
> That refers to [JLS 5.5](https://docs.oracle.com/javase/specs/jls/se23/html/jls-5.html#jls-5.5). Should this specifically be called out in the spec? It should just fall out from our nullness subtyping rules.

Yes, I found "If a casting context makes use of a narrowing reference conversion that is checked or partially unchecked" there. I figured that meant a checked narrowing reference conversion, and I saw in [5.1.6.2](https://docs.oracle.com/javase/specs/jls/se23/html/jls-5.html#jls-5.1.6.2) that "A narrowing reference conversion from a type S to a parameterized class or interface type T is unchecked, unless at least one of the following is true: [skipping irrelevant case] T <: S, and S has no subtype X other than T where the type arguments of X are not contained in the type arguments of T." I found it difficult to understand that.

In any case, could you show me a conformance-test-style snippet with a non-pattern `instanceof` that uses annotations in the parameterized types, where a tool might emit useful information?

PR #680, opened 2024-09-30
Expand list of declaration locations that are unrecognized

wmdietl:
I found the subset of the declaration locations where annotation locations are unrecognized confusing.
I went through [ElementType](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/annotation/ElementType.java) and list all the places that are unrecognized, in the non-normative part.
(I didn't mention "type use", that should be clear as we talk about declaration annotations.

I also found the "named class" confusing, as it seems to exclude interfaces. I'm not sure "named type" is better... The document in other places makes a distinction between class and interface, so just listing "class" here seems odd.
In some earlier document we explicitly stated that we'll use "class" for all kinds of type declarations. Maybe we should add that to this document? I have the same issue with the definition of enclosing talking about "class members" and "class".

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -321,15 +321,16 @@ We provide two parameterless declaration annotations: `@NullMarked` and
 Our declaration annotations are specified to be *recognized* when applied to the
 locations listed below:
 
--   A *named* class.
+-   A *named* type.
 -   A package.
 -   A module (for `@NullMarked` only, not `@NullUnmarked`).
 -   A method or constructor.
 
-> *Not* a field or a record component.
-
 If our declaration annotations appear in any other location, they have no
 meaning.
+
+> That is, they are *not* recognized on a field, a parameter, a local variable,
+> an annotation interface, a type parameter, or a record component declaration.
 
 ## Null-marked scope
 
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -104,6 +104,8 @@ A *base type* is a type as defined in [JLS 4].
 > JLS 4 does not consider a type-use annotation to be a part of the type it
 > annotates, so neither does our concept of "base type."
 
+We use *class* for classes, interfaces, enums, annotations, and records.
+
 ## Type components
 
 A *type component* of a given type is a type that transitively forms some part
@@ -171,7 +173,7 @@ corresponding to *each* of its [type components].
 > nullness portion of the type.
 
 For our purposes, base types (and thus augmented types) include not just class
-and interface types, array types, and type variables but also
+types, array types, and type variables but also
 [intersection types] and the null type.
 
 > This spec aims to define rules for augmented types compatible with those that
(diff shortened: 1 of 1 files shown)


docs/docs/spec.md
```diff
@@ -321,15 +321,15 @@ We provide two parameterless declaration annotations: `@NullMarked` and
...
 -   A module (for `@NullMarked` only, not `@NullUnmarked`).
 -   A method or constructor.
 
-> *Not* a field or a record component.
+All locations that are not explicitly listed as recognized are unrecognized.
 
-If our declaration annotations appear in any other location, they have no
-meaning.
+> That is, they are *not* recognized on a field, a parameter, a local variable,
+> an annotation interface, a type parameter, or a record component declaration.
```
cpovirk, line 332:
```suggestion
> a type parameter, or a record component declaration.
```

I think we should leave out "an annotation interface," since our annotations are still recognized there by virtue of being recognized on _any_ class/type.

cpovirk (commented):
> The document in other places makes a distinction between class and interface, so just listing "class" here seems odd.

I did a quick search for "interface" and turned up this:

> For our purposes, base types (and thus augmented types) include not just class and interface types, array types, and type variables but also [intersection types](https://jspecify.dev/docs/spec/#intersection-types) and the null type.

Do you have other examples in mind?

I suspect you're right that it would make sense to say that we're saying "class" for all kinds of type declarations. I haven't dug up specifically where that was mentioned, but we can write something from scratch. (I may well have dropped it by accident when trying to make the spec able to stand alone, integrating any glossary concepts where necessary.)

wmdietl:
I think the JLS mostly uses "class or interface types", so always makes this distinction between classes and interfaces.
So when just reading "classes" here, I wasn't sure whether this excludes interfaces or not.
So I would prefer some explicit statement about how to read things, ideally using JLS vocabulary.

cpovirk:
I think it would be good to define that we use "class" as shorthand for "class or interface." I think that would be preferable to using the term "type" (or even "named type"), which could in theory be read as including a type-parameter declaration (though of course your new text should help correct anyone who reads it that way) and which of course is a term that we already use for a lot of things.

I could see adding a section with the title "Class" near the top, perhaps just above the existing section "Base type" or something, and putting the definition there. But if you already have a vision for how to introduce it, I'm not attached to this particular approach.

cpovirk:
Oh, and "class or interface" might also work, but we would probably have to rephrase sections like "Each non-top-level class \[or interface\] is enclosed by a class \[or interface\] or class \[or interface\] member" :)

cpovirk:
(Also: While the JLS's usage of "class or interface" would give me a clear conscience about using it from the perspective of technical accuracy (setting aside [my concerns about awkwardness](https://github.com/jspecify/jspecify/pull/680#issuecomment-2403310907)), I wouldn't mind using the opportunity to define "class" more broadly as an opportunity to mention that yes, we do include enums and annotations and record classes. (Perhaps it should even be a _non-normative_ comment?). Maybe we can even figure out whether we need to be careful about saying "a _named_ class" (which I think is in opposition to "an anonymous class," which maybe we could make clearer) in more places that we currently do.)

netdpb:
If we're going to be precise, maybe we should say "named class or interface declaration", and add "declaration" to the others.

Otherwise, it's ambiguous whether `new Foo()` contains a named class.

cpovirk (approved):

PR #681, opened 2024-09-30
Clarify that JSpecify declaration annotations are meant

wmdietl:
Because a declaration annotation with `@Target(ElementType.TYPE_PARAMETER)` can annotate type parameters.

One question about "enclosing": is it clear what the enclosing declaration of type parameters and extends or implements clauses is? The rules talk about "class member", but I'm not sure I would consider those class members.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -351,8 +351,8 @@ innermost.
 > This definition of "enclosing" largely matches
 > [the definition in the Java compiler API](https://docs.oracle.com/en/java/javase/23/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement\(\)).
 > The JSpecify definition differs slightly by skipping type-parameter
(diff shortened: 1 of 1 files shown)


cpovirk (approved):
Good point about type parameters. I think that there is a deeper problem here, which is that we rely broadly on a vague syntax-tree-based sense of "enclosing," not just on the definition that we give here. I have https://github.com/jspecify/jspecify/issues/622 filed about that. It is possible that we actually should just keep the type-parameter case. We'll need some rule that says what _bounds_ of a type parameter are enclosed by, and that rule could either be they're enclosed by the type parameter or that they're enclosed by the class or method/constructor that declares the type parameter. The virtue of the former rule is that it follows the Java compiler's definition; the virtue of the latter rule is that it avoids having a step that says to look for (and then ignore) our declaration annotations on the type-parameter declaration. I'm not sure what will be best there.

netdpb (approved):

PR #682, opened 2024-09-30
Provide motivation for the `@kotlin.Metadata` rule

wmdietl:
I missed the discussion of #611 and #637.

I find it a bit odd that Kotlin code is always treated as unmarked, even when it appears within a marked package/module.
Won't this lead to always requiring that Kotlin classes have `@Nullmarked`, even in a package/module that is already `@Nullmarked`?
Wouldn't this make them second-class citizens?

I'm fine with keeping the rule, but would prefer some motivation in the specification document.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -366,7 +366,7 @@ check the following rules in order:
 > If a given declaration is annotated with both `@NullMarked` and
 > `@NullUnmarked`, these rules behave as if neither annotation is present.
 
-If none of the enclosing declarations meet either rule, then the type usage is
+If none of the enclosing declarations meet any rule, then the type usage is
 *not* in a null-marked scope.
 
 ## Augmented type of a type usage appearing in code {#augmented-type-of-usage}
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -362,8 +362,6 @@ check the following rules in order:
     `@NullUnmarked`, the type usage is in a null-marked scope.
 -   If the declaration is annotated with `@NullUnmarked` and *not* with
     `@NullMarked`, the type usage is *not* in a null-marked scope.
--   If the declaration is a top-level class annotated with `@kotlin.Metadata`,
-    then the type usage is *not* in a null-marked scope.
 
 > If a given declaration is annotated with both `@NullMarked` and
 > `@NullUnmarked`, these rules behave as if neither annotation is present.
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -386,9 +386,33 @@ check the following rules in order:
     `@NullUnmarked`, the type usage is in a null-marked scope.
 -   If the declaration is annotated with `@NullUnmarked` and *not* with
     `@NullMarked`, the type usage is *not* in a null-marked scope.
+-   If the declaration is a top-level class annotated with `@kotlin.Metadata`,
+    then the type usage is *not* in a null-marked scope.
 
 > If a given declaration is annotated with both `@NullMarked` and
 > `@NullUnmarked`, these rules behave as if neither annotation is present.
+
+> The Kotlin compiler emits nullness annotations (currently,
+> non-JSpecify annotations) on root types for method parameters,
+> method returns, and fields, but it does not emit annotations
+> elsewhere. In [some
+> cases](https://youtrack.jetbrains.com/issue/KT-13228), Kotlin allows
+> authors to add nullness annotations in those locations
+> manually. However, in general, Kotlin code is "missing" annotations,
+> so it should not be treated as null-marked. As a pragmatic way to
+> accommodate that, the spec has this rule to treat Kotlin code as
+> null-unmarked, even when the code is located in a null-marked
+> package or module.
+
+> In the future, Kotlin may [emit full nullness
+> information](https://youtrack.jetbrains.com/issue/KT-47417),
+> including a `@NullMarked` annotation at the class level. The spec rule
+> for `@kotlin.Metadata` is formulated so that such code will
+> automatically be treated as null-marked at that point. Additionally,
+> it is formulated so that Kotlin code can be explicitly annotated as
+> `@NullMarked` to override the `@kotlin.Metadata` rule for a class and
+> any nested classes. (Another possibility even today is for tools to
+> read the full Kotlin nullness information from `@kotlin.Metadata`.)
 
 If none of the enclosing declarations meet any rule, then the type usage is
 *not* in a null-marked scope.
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks.

First, "either" -> "any" LGTM :)

For the actual rule: Good point. We should get into that.

My attempt would be something like this:

> The Kotlin compiler emits nullness annotations (currently, non-JSpecify annotations) on root types for method parameters, method returns, and fields, but it does not emit annotations elsewhere. In [some cases](https://youtrack.jetbrains.com/issue/KT-13228), Kotlin allows authors to add nullness annotations in those locations manually. However, in general, Kotlin code is "missing" annotations, so it should not be treated as null-marked. As a pragmatic way to accommodate that, the spec has this rule to treat Kotlin code as null-unmarked, even when the code is located in a null-marked package or module.
>
> In the future, Kotlin may [emit full nullness information](https://youtrack.jetbrains.com/issue/KT-47417), including a `@NullMarked` annotation at the class level. The spec rule for `@kotlin.Metadata` is formulated so that such code will automatically be treated as null-marked at that point. Additionally, it is formulated so that Kotlin code can be explicitly annotated as `@NullMarked` to override the `@kotlin.Metadata` rule for a class and any nested classes. (Another possibility even today is for tools to read the full Kotlin nullness information from `@kotlin.Metadata`.)

cpovirk:
> The Kotlin compiler emits nullness annotations

One thing that I forgot to call out: Since it doesn't take advantage of `@NullMarked`, it emits even _`@NonNull`_ annotations. That at least means that the _need_ for `@NullMarked` is weaker, though of course we still ultimately want Kotlin to generate `@NullMarked` (along with appropriate annotations in "type-use-only" locations), since "weaker" doesn't mean "negligible" :)

netdpb:
I think the motivation is that since Kotlin already has full nullness information in its type system, we don't expect Kotlin authors to add JSpecify annotations explicitly. We expect Java analysis tools to interpret Kotlin's metadata when considering Java code that interacts with Kotlin symbols, and we don't want to specify anything about it even when the Kotlin symbol is in a null-marked package or module.

On the other hand, if Kotlin ever does emit JSpecify annotations, I hope it will emit `@NullMarked` on each top-level class. In that case, the spec is prepared.

wmdietl:
I like [the motivation](https://github.com/jspecify/jspecify/pull/682#issuecomment-2383202348) provided by Chris.

I'm still a bit concerned that a Kotlin classfile will never be treated like a Java classfile. The Kotlin class will always (?) have the additional `@kotlin.Metadata` annotation. So it will never be possible to have a Kotlin classfile that is covered by a `@Nullmarked` package annotation. Kotlin classfiles will always stick out with a separate `@Nullmarked` on the class.

I'm also curious how this rule would convert to a possible Java nullness system.

cpovirk:
In principle, I dislike the `@kotlin.Metadata` rule, but in practice, I think it provides the best user experience for users who mix Java and Kotlin in the same package (perhaps because they null-marked some Java code and are gradually converting it to Kotlin?) and who use `@NullMarked` in `package-info`. (But really, I would personally still advise using class-level annotations over package-level ones, as much as I would like for package-level annotations to work well.)

I think of it conceptually somewhat like if `@kotlin.Metadata` implied `@NullUnmarked`, though that's not the exact behavior because of the caveats like what happens if the class is _also_ annotated with `@NullMarked`. That might even literally be what's going on as far as users can tell unless they actually look at the Kotlin class file and other details :)

The goal of course is to render the rule mostly irrelevant by having Kotlin emit `@NullMarked` and the other annotations in newly compiled code in the future. Assuming that that happens, we may well reach a point at which we can actually rip the rule out.

cpovirk:
Sorry, I forgot to respond about a future possible Java nullness system.

And... I don't know :) Rough possibilities include:

- Any Java information (from `?`/`!` or a file-level(?) default) is authoritative (since Java-recognized non-nullable types in particular can lead to NPE no matter what JSpecify says).
- JSpecify annotations that contradict Java information are unrecognized.
- JSpecify annotations that agree with Java information are probably redundant, though perhaps we'd encourage tools to recognize a transition period during which they don't aggressively report those to users (in case some tools haven't been updated to recognize Java's nullness information yet). (Such a rule might be conceptually similar to https://github.com/jspecify/jspecify/pull/645 / https://github.com/jspecify/jspecify/pull/646, though it would not actually match.)

I may have said before that I _speculate_ that the Java nullness system will be useful _initially_ mostly for:
- optimizations in the VM (obviously)
- as a more ergonomic way of writing `@Nullable` (and `@NonNull` in the occasions that it's required), with the expectation that JSpecify users will continue to supplement with `@NullMarked` until Java perhaps provides some kind of defaulting mechanism

netdpb:
> I'm still a bit concerned that a Kotlin classfile will never be treated like a Java classfile. The Kotlin class will always (?) have the additional `@kotlin.Metadata` annotation. So it will never be possible to have a Kotlin classfile that is covered by a `@Nullmarked` package annotation. Kotlin classfiles will always stick out with a separate `@Nullmarked` on the class.

I think that's appropriate. Packages and JPMS modules are not [valid Kotlin annotation targets](https://kotlinlang.org/spec/annotations.html#annotation-targets).

I would never expect Kotlin authors to annotate their Kotlin classes with JSpecify annotations; do you? Given that, I would not want JSpecify to incorrectly interpret type usages without `@Nullable`within those classes to be non-null just because there's a package-info.java file that Kotlin ignores.

If and when the Kotlin compiler emits complete JSpecify nullness annotations based on Kotlin nullness metadata, it should emit `@NullMarked` on those classes as well.

wmdietl:
I added the [extra motivation](https://github.com/jspecify/jspecify/pull/682#issuecomment-2383202348) provided by Chris.

If Kotlin will ever support `package-info.java` and `module-info.java`, this might need a revision, to allow seamless integration into a Java project. Until then, this rule seems like the best option.

cpovirk (approved):
I pushed a commit with an organizational proposal because AFAIK there is no way to suggest a multi-line edit(?).

netdpb had [pointed out](https://github.com/jspecify/jspecify/pull/657#discussion_r1769241725) that adjacent but separate blocks look weird, so I merged these together. Then I put in some introduction and transition. Feel free to counter-edit before merging.

wmdietl:
> I pushed a commit with an organizational proposal because AFAIK there is no way to suggest a multi-line edit(?).

You can mark multiple lines with the "+" in the file diff view. Then you can suggest edits for multiple lines.

> netdpb had [pointed out](https://github.com/jspecify/jspecify/pull/657#discussion_r1769241725) that adjacent but separate blocks look weird, so I merged these together. Then I put in some introduction and transition. Feel free to counter-edit before merging.

Edits look good, thanks!

PR #694, opened 2024-10-18
Remove array creation expressions, again

wmdietl:
#674 undid the change from #678.
I think this is just a merge issue, as the later text that says array creations are intrinsically non-null is still there.


cpovirk (approved):
Yes, sorry, merge trouble. Thanks for catching it.

PR #695, opened 2024-10-19
Examples and tweaks

wmdietl:


docs/docs/spec.md
```diff
@@ -452,6 +459,10 @@ condition is met, skip the remaining conditions.
 -   If the type usage is a component of a return type in an annnotation
     interface, its nullness operator is `MINUS_NULL`.
 
+-   TODO: Why does this only list two of the intrinsically non-null locations listed earlier?
```
cpovirk, line 462: That is a good point.

I don't think I'd thought about this in much detail, but I suspect that my thought process was effectively this:

- For some of those locations, we don't need to care, since we [haven't made much attempt to address implementation code](https://github.com/jspecify/jspecify/issues/661). (This would apply to exception parameters and local variables, for example.)

- For some others, there's clearly no rule about nullness that would apply to them. (For example, what would it even mean to write `class Foo extends @Nullable Object`?)

- That leaves us with some cases that _look_ like fields and _look_ like methods but are actually _special_ fields (enum constants) and _special_ methods (annotation elements).

Given that, I was drawn to covering that final kind of case in the spec.

But I think that's wrong:

> This section defines how to determine the augmented types of most type usages in source code or bytecode where JSpecify nullness annotations are recognized.

An enum constant and a return type in an annotation interface\[*\] are both unrecognized locations. So even if we want to have rules like this, they don't belong here.

And then, once we do have them, there's not a clear reason to cover them but not cover other locations. The closest thing I have to a reason is the "criteria" in the bullet points above. Even those points would leave us open to questions about whether cases like receiver parameters meet the criteria.

I see at least two courses of action:

1. Just remove these rules. (We could then consider adding a non-normative comment in this section that refers back to the intrinsically non-nullable types mentioned ealier, perhaps with particular emphasis on the enum and annotation cases on account of my third bullet above.)

2. Split this section in two, with the first half covering recognized locations and the second half covering (for now, just these two) unrecognized locations.

Do you have a preference or another idea? The obvious other option would be to try to cover _all_ locations, but I'd at least like to declare that part of 1.0.1 or whatever we call the next version of the spec.

\[*\] I was having trouble finding that, it turns out because of the typo that I'll fix in https://github.com/jspecify/jspecify/pull/697 :)

docs/docs/spec.md
```diff
@@ -426,7 +432,8 @@ If none of the enclosing declarations meet any rule, then the type usage is
 ## Augmented type of a type usage appearing in code {#augmented-type-of-usage}
 
 This section defines how to determine the [augmented types] of most type usages
-in source code or bytecode where JSpecify nullness annotations are [recognized].
+in source code or bytecode where JSpecify nullness annotations are
+[recognized](#recognized-type-use).
```
cpovirk, line 436: I spent a few minutes on trying to figure out whether we could get Docusaurus to fail for broken links like these. But we already set the settings that look relevant:

https://github.com/jspecify/jspecify/blob/996b10e7786a46a9d31d0110d4c70e95258a28bd/docs/docusaurus.config.js#L29-L30

Maybe it's just technically "not even a link" if the contents of the bracket don't have a corresponding link definition at the end. Oh, well. I guess that we just search the page for `[` (and ignore code samples about arrays :)). Thanks for fixing this one.

docs/docs/spec.md
```diff
@@ -467,7 +478,11 @@ condition is met, skip the remaining conditions.
     > This special case handles the fact that the Java compiler automatically
     > generates an implementation of `equals` in `Record` but does not include a
     > `@Nullable` annotation on its parameter, even when the class is
-    > `@NullMarked`.
+    > `@NullMarked`. Note that this special handling is not necessary for the
+    > return type of `String toString()`.
+
+    TODO: should this only apply if the method isn't manually declared in the record?
+    As written now, it would also apply for a manually implemented `equals`, which I would find inconsistent.
```
cpovirk, line 485: We don't _want_ it to apply for a manually implemented `equals`, but there's no way to tell that apart from an automatically generated one in the class file :(

There's some discussion scattered throughout https://github.com/jspecify/jspecify/issues/301 and https://github.com/jspecify/jspecify/pull/645.

What we have at the moment is the best we've been able to come up with, with the combination of this and https://jspecify.dev/docs/spec/#expected-annotations-on-record-classes-equals-methods to make clear that people _should_ annotate any `equals` method that they hand-write but that we essentially "accidentally" accommodate them if they don't.

docs/docs/spec.md
cpovirk: Hmm, enum constants may be a special case: We don't currently say that annotations are unrecognized on _the type of_ an enum constant, since there's no way to write them there in source code. But there _is_ a type in bytecode, as we note in https://jspecify.dev/docs/spec/#augmented-type-of-usage. So it would seem that either we need to say that annotations are unrecognized there (so that the normal rules for computing the type don't apply) or that we need to keep the rule for computing the nullness operator there (for the same reason).

docs/docs/spec.md
wmdietl: Thanks for these links! I've adapted the text and included a forward reference. (You'll need to re-format again, sorry about that.)

docs/docs/spec.md
wmdietl: I've added one rule that covers all intrinsically non-null locations and calls out two examples.
What do you think?

docs/docs/spec.md
```diff
@@ -440,15 +447,16 @@ usage, this section covers only how to determine its [nullness operator].
 To determine the nullness operator, apply the following rules in order. Once one
 condition is met, skip the remaining conditions.
 
--   If the type usage is the type of the field corresponding to an enum
-    constant, its nullness operator is `MINUS_NULL`.
+-   If the type usage is in an intrinsically non-null locations listed earlier, its nullness operator is `MINUS_NULL`.
```
wmdietl, line 450: What isn't nice about this is that the "intrinsically non-nullable locations" given earlier was in non-normative text, so referring to it in normative text seems bad.

cpovirk: Oh, good point.

Maybe the most technically correct thing would be to add an entry about enum-constant _types_ to "However, the type-use annotation is unrecognized in any of the following cases," presumably moving the non-normative comment about enum constants themselves up to there. Then we at least would clearly not be saying that our rules apply to enum-constant types.

We could then still include the text you have here, but it too could be a non-normative comment.

Really, though, I think the chance of _practical_ problems from all this is approximately zero :) It's worth getting right if we can, but if you don't see an approach that you think works well, then I'm fine with what you have or anything else reasonable.

wmdietl: I tried doing what you suggested. I highlighted that this is only for *recognized* locations, so the previous rules didn't make sense, as they are unrecognized.

cpovirk (approved):

PR #700, opened 2024-10-21
My batch of nitpicking:

cpovirk:
- Update "multiple worlds" non-normative text to reflect that [we also
  consider the possibility of
  `MINUS_NULL`](https://github.com/jspecify/jspecify/issues/248).
- Make the initial non-normative comments look like other non-normative
  comments, moving the explanation of non-normative comments before most
  of them.
- Join together some adjacent non-normative comments, as in
  https://github.com/jspecify/jspecify/pull/657#discussion_r1769241725,
  adding some transition words in one case.

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -8,8 +8,8 @@ sidebar_position: 6
 
 This document specifies the semantics of our set of nullness annotations.
 
-> This document is mainly useful to authors of analysis tools. Some
-> very advanced users might find it interesting, but it would make a very poor
+> This document is mainly useful to authors of analysis tools. Some very
+> advanced users might find it interesting, but it would make a very poor
 > introduction for anyone else; instead see our **[Start Here](/docs/start-here)
 > page**.
 
@@ -27,8 +27,8 @@ rules.
 > ### The word "nullable"
 >
 > In this doc, we try not to refer to whether a type "is nullable." Instead, we
-> define at least four kinds of "Is it nullable?"
-> questions we can ask for any given type usage:
+> define at least four kinds of "Is it nullable?" questions we can ask for any
+> given type usage:
 >
 > 1.  What is the [augmented type] of that type usage?
 > 2.  Do I have to handle the case where `null` comes out of it?
@@ -41,11 +41,11 @@ rules.
 > spec. For example, it does not state when tools must check that the
 > [subtyping] relation holds.
 >
-> We anticipate that tools will typically apply each part of this spec in the same
-> cases that they (or `javac`) already apply the corresponding part of the Java
-> Language Specification. For example, if code contains the parameterized type
-> `List<@Nullable Foo>`, we anticipate that tools will check that `@Nullable
-> Foo` is a subtype of the bound of the type parameter of `List`.
+> We anticipate that tools will typically apply each part of this spec in the
+> same cases that they (or `javac`) already apply the corresponding part of the
+> Java Language Specification. For example, if code contains the parameterized
+> type `List<@Nullable Foo>`, we anticipate that tools will check that
+> `@Nullable Foo` is a subtype of the bound of the type parameter of `List`.
 >
 > However, this is up to tool authors, who may have reasons to take a different
 > approach. For example:
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -8,8 +8,8 @@ sidebar_position: 6
 
 This document specifies the semantics of our set of nullness annotations.
 
-> The primary audience for this document is the authors of analysis tools. Some
-> very advanced users might find it interesting. But it would make a very poor
+> This document is mainly useful to authors of analysis tools. Some
+> very advanced users might find it interesting, but it would make a very poor
 > introduction for anyone else; instead see our **[Start Here](/docs/start-here)
 > page**.
 
@@ -26,8 +26,8 @@ rules.
 
 > ### The word "nullable"
 >
-> In this doc, we aim not to refer to whether a type "is nullable." Instead, we
-> draw some distinctions, creating at least four kinds of "Is it nullable?"
+> In this doc, we try not to refer to whether a type "is nullable." Instead, we
+> define at least four kinds of "Is it nullable?"
 > questions we can ask for any given type usage:
 >
 > 1.  What is the [augmented type] of that type usage?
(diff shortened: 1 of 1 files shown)


netdpb:
> * Update "multiple worlds" non-normative text to reflect that [we also
>   consider the possibility of
>   `MINUS_NULL`](https://github.com/jspecify/jspecify/issues/248).
> * Make the initial non-normative comments look like other non-normative
>   comments, moving the explanation of non-normative comments before most
>   of them.
> * Join together some adjacent non-normative comments, as in
>   [Improve explanation of "the `ImmutableList.Builder` rule." #657 (comment)](https://github.com/jspecify/jspecify/pull/657#discussion_r1769241725),
>   adding some transition words in one case.

Why do some of your
descriptions contain lots of
arbitrary breaks?

docs/docs/spec.md
```diff
@@ -8,67 +8,10 @@ sidebar_position: 6
...
-tool may recognize additional annotations. Or a tool may omit the concept of
-`UNSPECIFIED` and apply a policy that type usages like `Object` are always
-non-nullable.
-
-### That's all!
-
-On to the spec.
-
---------------------------------------------------------------------------------
+> The primary audience for this document is the authors of analysis tools. Some
```
netdpb, line 11:
```suggestion
> This document is mainly useful to authors of analysis tools. Some
```

docs/docs/spec.md
```diff
@@ -8,67 +8,10 @@ sidebar_position: 6
...
+> very advanced users might find it interesting. But it would make a very poor
```
netdpb, line 12:
```suggestion
> very advanced users might find it interesting, but it would make a very poor
```

docs/docs/spec.md
```diff
@@ -81,6 +24,52 @@ This document also links to other documents. Those documents are non-normative,
 except for when we link to the Java Language Specification to defer to its
 rules.
 
+> ### The word "nullable"
+>
+> In this doc, we aim not to refer to whether a type "is nullable." Instead, we
```
netdpb, line 29:
```suggestion
> In this doc, we try not to refer to whether a type "is nullable." Instead, we
```

docs/docs/spec.md
```diff
@@ -81,6 +24,52 @@ This document also links to other documents. Those documents are non-normative,
...
+> draw some distinctions, creating at least four kinds of "Is it nullable?"
+> questions we can ask for any given type usage:
+>
+> 1.  What is the [augmented type] of that type usage?
```
netdpb, line 33: I'm still not convinced that this is an "is it nullable" question. This is how JSpecify proposes that people _answer_ any of the "is it nullable" questions. But okay.

docs/docs/spec.md
```diff
@@ -81,6 +24,52 @@ This document also links to other documents. Those documents are non-normative,
 except for when we link to the Java Language Specification to defer to its
 rules.
 
+> ### The word "nullable"
+>
+> In this doc, we aim not to refer to whether a type "is nullable." Instead, we
+> draw some distinctions, creating at least four kinds of "Is it nullable?"
+> questions we can ask for any given type usage:
```
netdpb, line 31:
```suggestion
> define at least four kinds of "Is it nullable?"
> questions we can ask for any given type usage:
```

Also, why hedge with "at least"? Are there more than four?

docs/docs/spec.md
```diff
@@ -81,6 +24,52 @@ This document also links to other documents. Those documents are non-normative,
...
+>
+> 1.  What is the [augmented type] of that type usage?
+> 2.  Do I have to handle the case where `null` comes out of it?
+> 3.  Do I have to prevent `null` from going into it?
+> 4.  Is this type a subtype of that type with respect to nullness?
+>
+> ### The scope of this spec
+>
+> Currently, this spec does not address *when* tools must apply any part of the
```
netdpb, line 40: "Currently"? How about "Not since 1947"? 😆 

Maybe "importantly" or "crucially" (too strong?).

docs/docs/spec.md
```diff
@@ -81,6 +24,52 @@ This document also links to other documents. Those documents are non-normative,
...
+> spec. For example, it does not state when tools must check that the
+> [subtyping] relation holds.
+>
+> We anticipate that tools will typically apply parts of this spec in the same
+> cases that they (or `javac`) already apply the corresponding parts of the Java
```
netdpb, line 45:
```suggestion
> We anticipate that tools will typically apply each part of this spec in the same
> cases that they (or `javac`) already apply the corresponding part of the Java
```

I think this is stronger. Otherwise it sounds like they'll just randomly pick parts of the spec to apply.

docs/docs/spec.md
```diff
@@ -81,6 +24,52 @@ This document also links to other documents. Those documents are non-normative,
...
+>     those types are "correct."
+>
+> -   Even when a tool is a "nullness checker," it might be written for another
+>     language, like Kotlin, with its own rules for when to perform type checks.
+>     Or the tool might target a future version of Java whose language features
+>     would not be covered by this version of this spec.
+>
+> Note also that this spec covers only nullness information *from JSpecify
+> annotations*. Tools may have additional sources of information. For example, a
+> tool may recognize additional annotations. Or a tool may omit the concept of
```
netdpb, line 69:
```suggestion
> tool may recognize additional annotations, or a tool may omit the concept of
```

docs/docs/spec.md
```diff
@@ -609,8 +598,9 @@ rules.
...
 > strictness in the presence of `UNSPECIFIED`. The basic idea is that, every
 > time a tool encounters a type component with the nullness operator
-> `UNSPECIFIED`, it has the option to fork off two "worlds": one in which the
-> operator is `UNION_NULL` and one in which it is `NO_CHANGE`.
+> `UNSPECIFIED`, it has the option to fork off three "worlds": one in which the
+> operator is `UNION_NULL`, one in which it is `MINUS_NULL`, and one in which it
+> is `NO_CHANGE`.
 >
 > In more detail: When tools lack a nullness specification for a type, they may
 > choose to assume that either of the resulting worlds may be the "correct"
```
wmdietl, line 606:
```suggestion
> choose to assume that any of the resulting worlds may be the "correct"
```

I agree that it should be three options. This reference to "either" should be changed to "any". I didn't look whether there are other references to "two" or "either".

netdpb (approved):
I have picked your nitpicks' nits.

docs/docs/spec.md
cpovirk: Yeah, I think we've made a couple attempts at how to phrase this. I'm sure we could find something that works better.

docs/docs/spec.md
cpovirk: I'll update it to say "four."

docs/docs/spec.md
cpovirk: Maybe we can just remove the word? I don't know, would something be "missing?"

cpovirk (commented):
Thanks.

cpovirk:
I'll hold off until after https://github.com/jspecify/jspecify/pull/695 is merged, and then I'll resolve the resulting merge conflicts.

wmdietl:
> I'll hold off until after #695 is merged, and then I'll resolve the resulting merge conflicts.

Thanks, I've merged #695.

docs/docs/spec.md
netdpb: That would currently be fine.

PR #702, opened 2024-10-30
Replace http://jspecify.org with https://jspecify.dev

KengoTODA:
I found that URLs in the official site still uses legacy URL with HTTP scheme.
Since `jspecify.org` provides no valid certificates, I suggest using `jspecify.dev` with HTTPS scheme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 An artifact of well-specified annotations to power static analysis checks and
 JVM language interop. Developed by consensus of the partner organizations listed
-at our main web site, [jspecify.org](http://jspecify.org).
+at our main web site, [jspecify.org](https://jspecify.dev).
 
 Our current focus is on annotations for nullness analysis.
 
@@ -14,4 +14,4 @@ finalizes our initial nullness annotations.
 
 ## Things to read
 
-See [jspecify.org/docs/start-here](http://jspecify.org/docs/start-here).
+See [jspecify.org/docs/start-here](https://jspecify.dev/docs/start-here).
--- a/conformance-tests/build.gradle
+++ b/conformance-tests/build.gradle
@@ -87,7 +87,7 @@ publishing {
                 version = project.version
                 name = 'JSpecify Conformance Test Suite'
                 description = 'Assertions and dependencies representing a suite of conformance tests for JSpecify'
-                url = 'http://jspecify.org/'
+                url = 'https://jspecify.dev/'
                 artifact distZip
                 licenses {
                     license {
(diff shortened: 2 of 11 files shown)


netdpb:
What's the deal here? I thought we wanted `jspecify.org` to work (since that's how the Java package is named), and that `jspecify.dev` was temporary, until we worked out something something.

kashike:
https://github.com/jspecify/jspecify/issues/267#issuecomment-1317157694

It looks like @cgdecker was working on getting this addressed back in 2022?

sdeleuze:
+1 to what @kevinb9n said.

wimdeblauwe:
Any progress on this? I can't really tell people to go to jspecify.org to learn about the tool if the first thing they get is a security warning about the website.

sdeleuze:
Indeed that far from ideal.
image

I suggest we close this PR (which is about replacing `https://jspecify.org` occurences by `https://jspecify.org` which can't be done everywhere due to the domain used for the packages). I will create a related issue to switch the domain to https://jspecify.org.

cpovirk:
I'm not sure exactly when the security warning began to appear: I think it was hidden for many of us, presumably because Chrome kept a record that we had already visited jspecify.org before? I learned of it when @sdeleuze raised this again recently.

As much as we're still hoping to fix jspecify.org to be a proper https site, I think we probably should merge this PR in the meantime. Would anyone be too bent out of shape about that?

cpovirk:
Yeah, there are at least two levels of issues here:

- https://jspecify.org has IIRC never worked. If you go there, you get a warning that is difficult or impossible to bypass. (I'm testing in Chrome here, but I'm sure the experience is similar in all browsers.)
- http://jspecify.org used to work, I think. It still "works" but only if you're willing to click through a maybe somewhat new-ish scary warning. (Again, that's what I'm seeing in Chrome, but I imagine all browsers are moving in this direction.) I didn't use to see this, unless _maybe_ I saw it once long, long ago, after which Chrome cached my answer and/or the redirect from that URL.

When the problem was just the first thing, that was bad but AFAIK only a problem if someone somewhere literally typed "https://jspecify.org." But nowadays people see warnings even when someone types plain "jspecify.org" or "http://jspecify.org."

cpovirk (approved):

cpovirk:
(A Google search for JSpecify, at least for me (including in an incognito window) still tries to take me to jspecify.org, so we still have plenty of reason to make that work better.)

sdeleuze:
Yeah + the dependency groupId. I still think what I described in https://github.com/jspecify/jspecify/issues/267#issuecomment-2654352301 is the only way to fully solve this issue.

PR #714, opened 2024-12-05
Update info for Kotlin 2.1, which now emits errors by default for JSp…

netdpb:
…ecify-annotated Java symbol usages.

### DIFF
--- a/docs/docs/whether.md
+++ b/docs/docs/whether.md
@@ -79,14 +79,14 @@ relaxing some of the nullness checks Kotlin otherwise performs. However, the
 compiler understands JSpecify annotations and is able to use them to make Kotlin
 code see null-safe types when calling into Java code. The Kotlin compiler
 correctly interprets `@Nullable` and `@NullMarked` starting at version 1.8.20,
-`@NonNull` starting at 2.0.0, and `@NullUnmarked` coming in 2.0.20.
+`@NonNull` starting at 2.0.0, and `@NullUnmarked` starting in 2.0.20.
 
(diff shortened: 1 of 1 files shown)


docs/docs/whether.md
```diff
@@ -79,14 +79,14 @@ relaxing some of the nullness checks Kotlin otherwise performs. However, the
...
-[`-Xjspecify-annotations=strict` and
-`-Xtype-enhancement-improvements-strict-mode`
-flags](https://kotlinlang.org/docs/whatsnew1520.html#support-for-jspecify-nullness-annotations).
-The Kotlin team plans to emit errors by default in Kotlin 2.1.0.
+`@NonNull` starting at 2.0.0, and `@NullUnmarked` starting in 2.0.20.
+
+As of [version 2.1.0][kotlin-2.1.0], the Kotlin compiler emits errors by default
+for problems found using JSpecify nullness.
+To change those to warnings, pass the
+`-Xnullability-annotations=org.jspecify.annotations=warning` flag.
```
cpovirk, line 87:
```suggestion
`-Xnullability-annotations=@org.jspecify.annotations:warn` flag.
```

That's what works for me, anyway. (I based it on an old command I used for RxJava's annotations a few years ago.)

We should see how we can get the Kotlin docs fixed, too....

cpovirk (approved):

docs/docs/whether.md
netdpb: No, that's on me for not reading the docs correctly.

PR #747, opened 2025-05-28
Set version in compiled module descriptor.

### LINKED ISSUE #746: Set module version in compiled module descriptor
As shown in

- #484 

the `org.jspecify` module doesn't come with a version in its compiled module descriptor.

```
jar --describe-module --file jspecify-1.0.0.jar --release 9
releases: 9

org.jspecify jar:file:///.../jspecify-1.0.0.jar!/META-INF/versions/9/module-info.class
exports org.jspecify.annotations
requires java.base mandated
```

It'd be nice to use `--module-version 1.0.0` when compiling the `module-info.java` file, or `jar --update` the version after packaging. That'll not only improve the module description produced by the `jar` tool, but also helps users when analyzing stacktraces and when listing available modules via `java --module-path lib --list-modules`:
```
[...]
jdk.xml.dom@22
jdk.zipfs@22
org.apiguardian.api@1.1.2 file:///.../apiguardian-api-1.1.2.jar
org.jspecify file:///.../jspecify-1.0.0.jar
org.junit.jupiter@6.0.0-SNAPSHOT file:///C.../junit-jupiter-6.0.0-SNAPSHOT.jar
[...]
```

cpovirk:
Fixes https://github.com/jspecify/jspecify/issues/746

### DIFF
--- a/gradle/mrjar.gradle
+++ b/gradle/mrjar.gradle
@@ -45,6 +45,10 @@ sourceSets {
 
 tasks.named('compileJava9Java', JavaCompile).configure {
     options.release = 9
+    options.compilerArgs.addAll([
+        '--module-version',
+        project.version.toString()
+    ])
 }
 
 jar {


sormuras:
Looks good to me - @marcphilipp, do you agree?

netdpb (approved):

marcphilipp:
> Looks good to me - @marcphilipp, do you agree?

Yes, the change looks good to me, too.

> Yeah, I'd be interested in thoughts from those of you on the (I assume :)) JUnit side. On the one hand, a new release is going to churn a bunch of people's build configurations and perhaps make them nervous that we plan to make a habit of frequent releases. On the other hand, it should _usually_ have no cost aside from what it takes to approve an automated update, and the sooner we publish a release, the smaller the number of affected users.

In my personal opinion, the biggest advantage of versioned module descriptors is when receiving bug reports that include stack traces. The versions included in the stack trace can help quickly spotting misalignment issues and other version conflicts. Since JSpecify only consists of annotations, I don't think it will ever be part of a stacktrace, though, will it? Therefore, I don't think this change alone justifies another release but that's your decision, of course.

sormuras:
> Do we need to release this? Would that make it 1.0.1?

With @raphw's https://github.com/raphw/jenesis-modules being used to resolve modular dependencies, it now would make more sense to have a `1.0.1` release, including the version in the compiled module descriptor.

cpovirk:
Thanks, I hadn't known about that.

Coincidentally, I just created https://github.com/jspecify/jspecify/milestone/9 just a couple weeks ago to compile possible motivation for another release. Maybe the combination of a few changes would be enough to justify the flood of Dependabot PRs to come :) I'll run it by the group.

cpovirk:
There has been general interest in publishing a 1.0.1 release, possibly in about a month. I don't expect a ton of controversy, but we'll see what might come up in the intervening time.

cpovirk:
(released in [1.0.1](https://github.com/jspecify/jspecify/releases/tag/v1.0.1), FYI @raphw and others)

PR #753, opened 2025-07-25
Document that JDK 21.0.8 has a fix for the type-use problem, with caveats.

cpovirk:


docs/docs/whether.md
```diff
@@ -103,13 +103,20 @@ are "declaration" annotations and hence are not impacted by this issue; however,
...
+The issue is fixed starting in JDK 22.
+
+Additionally, there is work underway to backport the fix to older versions of
+`javac`. As of July 2025, the backport has only been applied to JDK 21.0.8, and
+it is off by default, so you must pass `-XDaddTypeAnnotationsToSymbol=true` in
+addition to using that version of javac. The support was disabled by default
+because of some crashes with incomplete classpaths, so you should also be aware
+of that risk. We continue to work on making the backport more widely available.
+You can track that effort in
 [JDK-8323093](https://bugs.openjdk.org/browse/JDK-8323093).
```
cpovirk, line 115: Maybe this is not the best link nowadays?

cushon: The blocker for re-enabling the fix on 21u is https://bugs.openjdk.org/browse/JDK-8359336, and the current plan for resolving that depends on https://bugs.openjdk.org/browse/JDK-8360895.

PR #765, opened 2025-09-17
Update try-with-resources tests with test:irrelevant-annotation:NonNull to use @NonNull instead of @Nullable

pyltsinmikhail:
Just align try-with-resources tests with their intended `NonNull` expectations.
Replace `@Nullable` with `@NonNull` for try-with-resources resource variables in tests marked with `test:irrelevant-annotation:NonNull`.
Keeps test names, metadata, and behavior consistent; no functional changes outside of annotation correctness in conformance tests.

### DIFF
--- a/conformance-tests/src/assertions/java/org/jspecify/conformance/tests/irrelevantannotations/notnullmarked/Other.java
+++ b/conformance-tests/src/assertions/java/org/jspecify/conformance/tests/irrelevantannotations/notnullmarked/Other.java
@@ -87,7 +87,7 @@ public class Other {
     try (@Nullable AutoCloseable a = () -> {}) {}
     // test:name:NonNull try-with-resources
     // test:irrelevant-annotation:NonNull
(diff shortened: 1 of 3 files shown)


cpovirk:
Thanks. If the CLA is more trouble than it's worth, I can tell from your PR description that you're dealing with https://github.com/jspecify/jspecify/blob/f78e5332ed76676e3bc746911985a00ef57aed4d/conformance-tests/src/assertions/java/org/jspecify/conformance/tests/irrelevantannotations/nullmarked/Other.java#L89 and similarly in the other two files, and I am up for updating them myself if you'd prefer. But you're also welcome to sort out the CLA, too.

pyltsinmikhail:
If you don't mind, could you update these files, please? Thank you!

cpovirk:
Sure. I've opened https://github.com/jspecify/jspecify/pull/766.

PR #766, opened 2025-09-18
Fix try-with-resources tests to use `@NonNull` where expected.

cpovirk:
This problem was reported by @pyltsinmikhail in
https://github.com/jspecify/jspecify/pull/765.


wmdietl (approved):
Looks consistent.

PR #755, opened 2025-08-04
Update to Gradle 9.0.0 and update plugin versions

wmdietl:
Update of Gradle also required some plugin updates. I tried to go through all versions I found.

Running

````
 > ./gradlew clean jar --warning-mode all

> Task :jar
Invocation of Task.project at execution time has been deprecated. This will fail with an error in Gradle 10. This API is incompatible with the configuration cache, which will become the only mode supported by Gradle in a future release. Consult the upgrading guide for further information: https://docs.gradle.org/9.0.0/userguide/upgrading_version_7.html#task_project

[Incubating] Problems report is available at: file:///home/wdietl/workspaces-server/jspecify/jspecify/build/reports/problems/problems-report.html

BUILD SUCCESSFUL in 744ms
6 actionable tasks: 6 executed
````

I didn't immediately see where the `jar` task is using `Task.project`. Maybe [here](https://github.com/jspecify/jspecify/blob/2563dc75c40685401ea650b76626642378c5befc/gradle/mrjar.gradle#L56)?

PR #768, opened 2025-09-19
Update samples-google-prototype-eisop to gradle 9.0.0

wmdietl:
Just enough of the changes in https://github.com/jspecify/jspecify/pull/755 to make usage of the samples in the Reference Checker work.


cpovirk (approved):
Thanks!

Issue #305, opened 2022-10-18
Write instructions for publishing a release


cpovirk:
like https://github.com/jspecify/jspecify/pull/187#issuecomment-873183875 but nicer

cpovirk:
- Maybe [don't use JDK 21+](https://github.com/jspecify/jspecify/pull/517), though _probably_ it would be fine.

cpovirk:
TODO: Do we need to change [the `-snapshot` line in `bnd.bnd`](https://github.com/jspecify/jspecify/pull/428/files#r1678183933)?

cpovirk:
If we someday publish stable Javadoc at a separate URL for each release (though I have some reservations about that), then we'll want to make sure that we [use a new enough JDK that we actually generate Javadoc](https://github.com/jspecify/jspecify/pull/293).

cpovirk:
We'll also want to update [the SHA-256 value in our usage instructions (currently only in the docs-1.0 branch)](https://github.com/jspecify/jspecify/blob/docs-1.0/docs/docs/using.md#bazel).

For 1.0 specifically, we'll also merge (well, probably rebase/fast-forward) the docs-1.0 branch onto the main branch. And for subsequent releases, we'll want to update the version numbers in the usage instructions.

cpovirk:
For 1.0, we should also replace https://github.com/jspecify/jspecify/wiki/adoption with a link to https://jspecify.dev/using (which is currently a broken link but which will go live with the 1.0 release).

And we should update https://github.com/jspecify/jspecify/wiki/jspecify-faq#user-content-is-now-the-time-to-adopt to link to https://jspecify.dev/whether (ditto).

(I skimmed the rest of the wiki and FAQ for other obvious references to the version that would need to be updated, and I didn't see any more.)

cpovirk:
Update 1: Some prep for https://github.com/jspecify/jspecify/pull/187#issuecomment-873183875 (not that this is hard, just that it's nice to have ready to copy and paste in the future, ideally into some GitHub Actions setup), with everything to be run in a clean clone to be safe:

`VERSION=1.0.0 && git checkout --detach && perl -pi -e "s/^version = .*/version = $VERSION/" gradle.properties && git commit -m "Set version to $VERSION." gradle.properties && git tag v$VERSION`

cpovirk:
[...]

PR #769, opened 2025-09-24
Disable the configuration cache for publish tasks

wmdietl:
After merging #755 CI failed when trying to publish artifacts, which is not tested in the PR CI.

I looked around a bit and it looks like the Nexus plugin is incompatible with the configuration cache. So, this PR selectively disables the configuration cache.

Tested with `ORG_GRADLE_PROJECT_sonatypeUsername="abc" ORG_GRADLE_PROJECT_sonatypePassword="def" ./gradlew publishConformanceTestsPublicationToSonatypeRepository --no-configuration-cache` which now fails because of invalid credentials, not because of configuration cache issues.

A previous issue/PR mentioned alternative "publish" plugins. Maybe somebody more familiar with those could rewrite this part.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@v4
       - name: Publish conformance tests snapshot
-        run: ./gradlew publishConformanceTestsPublicationToSonatypeRepository
+        run: ./gradlew publishConformanceTestsPublicationToSonatypeRepository --no-configuration-cache
         env:
           ORG_GRADLE_PROJECT_sonatypeUsername: ${{ secrets.sonatype_username }}
           ORG_GRADLE_PROJECT_sonatypePassword: ${{ secrets.sonatype_password }}


cpovirk (approved):
Thanks, I see https://github.com/gradle-nexus/publish-plugin/issues/221.

As I understand things, the configuration cache is just a performance feature (as you'd hope from the word "cache" :)), so this seems like a good way to go. I could see disabling the cache entirely just to make our lives potentially simpler for a _non_-snapshot release in the future. But this PR addresses the immediate problem, and the plugin may well be fixed before we attempt a non-snapshot release. As an additional line of defense, I'll link this issue from our release "instructions" in https://github.com/jspecify/jspecify/issues/305.

cpovirk:
> I could see disabling the cache entirely just to make our lives potentially simpler

Heh: https://github.com/jspecify/jspecify/pull/812, https://github.com/jspecify/jspecify/pull/836

The good news is that, between that first fix and https://github.com/jspecify/jspecify/pull/860, we should be able to remove this flag. I've created https://github.com/jspecify/jspecify/pull/862, which will fail CI until we've merged that fix and some other changes to our GitHub Actions configuration.

PR #796, opened 2026-04-17
Clarifying user-guide type annotations vs type use annotation

mipo256:
This change, although it is quite small, I think it is useful, since strictly speaking in these cases we're not annotating the `String` type declaration, we're annotating the type usage.

It just confused me a bit, and I thought it might be a good idea to clarify this.


mipo256:
CLA with Google signed

cpovirk (approved):

PR #809, opened 2026-06-30
Place `module-info` in the root of the jar.

### LINKED ISSUE #484: Move module-info back to the main root?
Currently, it's under the 9+ section of the mrjar. That's fine, but I wonder if we'd be better off getting modular-style Javadoc output, as we might if we moved it to the root? (Such output might either help or hurt people who are trying to link to us. Maybe it is part of the story with https://github.com/maxcellent/javadoc.io/issues/182, but I haven't tried to understand that....)

cpovirk:
...instead of in a multi-release directory.

Fixes https://github.com/jspecify/jspecify/issues/484

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,7 +31,7 @@ repositories { mavenCentral() }
 def javaVersion = JavaLanguageVersion.of(System.env.JAVA_VERSION ?: 11)
 java.toolchain.languageVersion = javaVersion
 
-apply from: "gradle/mrjar.gradle"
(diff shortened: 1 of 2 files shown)


netdpb (approved):
Looks good to me, although I'm no expert.

cpovirk:
Thanks! I continue to invite comment here from anyone more experienced in Gradle than I am. (That is a low bar.) But I am fairly comfortable merging this as it is if no one jumps in.

PR #810, opened 2026-06-30
Warn that `@NullMarked` on a module has no effect unless users put it on the module path.

### LINKED ISSUE #496: Figure out and document whether module-level `@NullMarked` helps users who don't declare a module
@xenoterracide asks: Suppose that I put `@NullMarked` in my library's `module-info`. Then:

> [If the consumers of the module aren't using JPMS is it still supposed to be respected by tools?](https://groups.google.com/g/jspecify-discuss/c/wpSl8ViESHQ/m/SNyG4kYlAAAJ)

@msridhar is [optimistic that tools will still recognize it](https://groups.google.com/g/jspecify-discuss/c/wpSl8ViESHQ/m/H-bmAVZYAwAJ).

I am more pessimistic. I worry on two levels:

- First, the obvious way to implement the search for `@NullMarked` is (in javac terms) to keep calling `getEnclosingElement()` until there are no enclosing elements left. I would assume that `pkg.getEnclosingElement()` will _not_ return the module for a package unless the package actually comes from the module path. (I have not verified this.) If so, then tools would need to go out of their way to look for the `module-info` for the package. (And if not _all_ the tools you use do this (e.g., Kotlin, IntelliJ, a build-time Java checker), then you'll get inconsistent behavior.)
- I fear that "the `module-info` for the package" is _technically_ not a well-defined question unless Java "loaded the module for real." _In practice_, you could probably look up the jar/directory that contains a given package (or a specific class from that package, since multiple jars/directories might contain classes from the same package!). Then you could look in that jar for `module-info.class` (possibly under `META-INF/versions/9` or similar) and read the file with ASM. But I wouldn't be shocked if that weren't _technically_ enough. Might the jar come from an in-memory filesystem that you might not have access to? Even if it comes from the filesystem, might it be in some kind of special bundle that you need to understand? And even if it comes from a plain old jar, what if the system is doing some wild bytecode rewriting during class loading, which _should_ have gotten applied to the `module-info` but never did because _Java_ didn't load the `module-info`? Does anything change if we're talking about the _annotation-processor_ module path? I suspect that this is all not a big practical concern, but it's at least going to be a reason that tool authors are not enthusiastic about putting in any special handling for the non-JPMS case: It feels like another case in which people may want to avoid ["swimming upstream."](https://github.com/jspecify/jspecify/issues/8#issuecomment-1889843135)

Once we know more about this, we should write something for tool authors and something for users (possibly even in the Javadoc). It's very possible that the advice for users will end up being: "You don't actually want to rely on module-level `@NullMarked` for libraries that you publicly release; it's best reserved for closed systems, like if your company has fully bought in to modules."

cpovirk:
Fixes https://github.com/jspecify/jspecify/issues/496

### DIFF
--- a/src/main/java/org/jspecify/annotations/NullMarked.java
+++ b/src/main/java/org/jspecify/annotations/NullMarked.java
@@ -82,7 +82,8 @@ import java.lang.annotation.Target;
  *
  * {@code @NullMarked} and {@link NullUnmarked @NullUnmarked} can be used on any package, class,
  * method, or constructor declaration; {@code @NullMarked} can be used on a module declaration as
- * well. Special considerations:
+ * well. ({@link NullUnmarked @NullUnmarked} is not supported on modules, since it's already the
+ * default.) Special considerations:
  *
  * <ul>
  *   <li>To apply this annotation to an entire (single) <b>package</b>, create a <a
@@ -100,8 +101,11 @@ import java.lang.annotation.Target;
  *   <li>Applying this annotation to an instance <b>method</b> of a <i>generic</i> class is
  *       acceptable, but is not recommended because it can lead to some confusing situations.
  *   <li>An advantage of Java <b>modules</b> is that you can make a lot of code null-marked with
- *       just a single annotation (before the {@code module} keyword). {@link NullUnmarked} is not
- *       supported on modules, since it's already the default.
+ *       just a single annotation (before the {@code module} keyword). <b>Warning</b>: Even if you
+ *       annotate the module for a library as {@code @NullMarked}, this has no effect for users who
+ *       place the library on the class path, only those who place the library on the <a
+ *       href="https://openjdk.org/projects/jigsaw/quick-start">module path</a>, which is less
+ *       commonly used.
  *   <li>If both {@code @NullMarked} and {@code @NullUnmarked} appear together on the same element,
  *       <i>neither</i> one is recognized.
  * </ul>
--- a/src/main/java/org/jspecify/annotations/NullMarked.java
+++ b/src/main/java/org/jspecify/annotations/NullMarked.java
@@ -103,7 +103,7 @@ import java.lang.annotation.Target;
  *   <li>An advantage of Java <b>modules</b> is that you can make a lot of code null-marked with
  *       just a single annotation (before the {@code module} keyword). <b>Warning</b>: Even if you
  *       annotate the module for a library as {@code @NullMarked}, this has no effect for users who
- *       place the library on the class path, only those who place the library on the <a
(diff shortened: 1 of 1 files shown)


src/main/java/org/jspecify/annotations/NullMarked.java
```diff
@@ -82,7 +82,8 @@
  *
  * {@code @NullMarked} and {@link NullUnmarked @NullUnmarked} can be used on any package, class,
  * method, or constructor declaration; {@code @NullMarked} can be used on a module declaration as
- * well. Special considerations:
+ * well. ({@link NullUnmarked @NullUnmarked} is not supported on modules, since it's already the
```
wmdietl, line 85:
```suggestion
 * well. ({@link NullUnmarked @NullUnmarked} is not supported on modules, since it is already the
```

wmdietl: I think in technical writing it is better to avoid contractions. There are a few others we could expand. Or we ignore this.

src/main/java/org/jspecify/annotations/NullMarked.java
```diff
@@ -100,8 +101,11 @@
  *   <li>Applying this annotation to an instance <b>method</b> of a <i>generic</i> class is
  *       acceptable, but is not recommended because it can lead to some confusing situations.
  *   <li>An advantage of Java <b>modules</b> is that you can make a lot of code null-marked with
- *       just a single annotation (before the {@code module} keyword). {@link NullUnmarked} is not
- *       supported on modules, since it's already the default.
+ *       just a single annotation (before the {@code module} keyword). <b>Warning</b>: Even if you
+ *       annotate the module for a library as {@code @NullMarked}, this has no effect for users who
+ *       place the library on the class path, only those who place the library on the <a
```
wmdietl:
```suggestion
 *       place the library on the class path; it affects only those who place the library on the <a
```

The transition reads a bit odd to me. Alternatively "class path, but only on those".

wmdietl (approved):
Two things to consider.

src/main/java/org/jspecify/annotations/NullMarked.java
cpovirk: I started to write a response here, and as it grew, I filed https://github.com/jspecify/jspecify/issues/819 instead. I will plan to merge like this but continue the discussion there.

src/main/java/org/jspecify/annotations/NullMarked.java
cpovirk: Thanks, definitely better your way.

PR #814, opened 2026-07-07
Fix links in `tool-conformance`, and enable strict checking of anchors.

cpovirk:
There were two issues:

First, we were pointing to an incorrect anchor:

```
Exhaustive list of all broken anchors found:
[INFO] Use `npm run serve` command to test your build locally.
- Broken anchor on source page path = /docs/tool-conformance/:
   -> linking to /docs/spec/#recognized-locations-for-declaration-annotations
```

Second, we were using that anchor as the destination for a number of links, each of which should have been pointing to a different section.

To detect future instances of the first problem, I've set `onBrokenAnchors: 'throw'` and confirmed that it does as I'd hope:

```
[INFO] Docusaurus version: 3.6.3
  [cause]: Error: Docusaurus found broken anchors!
Node version: v18.17.1

  Please check the pages of your site in the list below, and make sure you don't reference any anchor that does not exist.
  Note: it's possible to ignore broken anchors with the 'onBrokenAnchors' Docusaurus configuration, and let the build pass.
```

### DIFF
--- a/docs/docs/tool-conformance.md
+++ b/docs/docs/tool-conformance.md
@@ -50,14 +50,15 @@ The questions are as follows:
 > **NOTE:** The examples for each question are not meant to be exhaustive.
 
 1.  Is a given JSpecify annotation recognized or unrecognized
-    ([type-use][spec-locations], [declaration][spec-locations]) in its context?
+    ([type-use][spec-type-use-locations],
+    [declaration][spec-declaration-locations]) in its context?
 
     *   Unrecognized: `@Nullable int`
     *   Unrecognized: `class Foo<@NonNull T extends Foo>`
     *   Recognized: `void foo(@Nullable Bar bar)`
 
-1.  What is the [null-augmented type][spec-locations] of a given type usage,
-    based only on JSpecify annotations on it or surrounding scopes?
+1.  What is the [null-augmented type][spec-augmented-type] of a given type
+    usage, based only on JSpecify annotations on it or surrounding scopes?
 
     ```java
     @NullMarked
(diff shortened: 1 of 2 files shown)


docs/docs/tool-conformance.md
```diff
@@ -157,4 +158,6 @@ all without affecting its conformance to JSpecify.
 [expression context]: https://docs.google.com/document/d/1nbTnJ0-HubLnQPKSjK5CDZyoe6Al64vdlkoJxaYX9XY/preview?resourcekey=0-ADjPZnp8LN3dRX_ptjlagw&tab=t.0#bookmark=kix.w6xfjhkftb9r
 [expression]: https://docs.google.com/document/d/1nbTnJ0-HubLnQPKSjK5CDZyoe6Al64vdlkoJxaYX9XY/preview?resourcekey=0-ADjPZnp8LN3dRX_ptjlagw&tab=t.0#bookmark=kix.2r97mw74ac6r
 [Map.get]: https://github.com/jspecify/jdk/blob/b7435cff373c527aad82a062c5605f6f9c1bb0de/src/java.base/share/classes/java/util/Map.java#L255
-[spec-locations]: /docs/spec#recognized-locations-for-declaration-annotations
+[spec-augmented-type]: /docs/spec#augmented-type
```
wmdietl, line 162: I don't understand how to read these links. Should this go to
https://github.com/jspecify/jspecify/blob/8bb43e2905eda9ed1598b832d8fcbbe8b116305a/docs/docs/spec.md?plain=1#L432 which uses "#augmented-type-of-usage"?
For the two below I find the expected labels in spec.md.

cpovirk: My change here is questionable on two levels, so let me respond on both :)

First: `augmented-type` is [a real anchor in the doc](https://jspecify.dev/docs/spec/#augmented-type). Since we don't specify an explicit name for it, its name is derived automatically from the header text of "Augmented type." Arguably we should write `{#augmented-type}`, etc. in the source anyway, both to reduce the chances that we break people's links/bookmarks without noticing and to make the anchors easier to grep for in the source.

Second: `augmented-type` may or may not be the best section to link to here. Your suggestion of `augmented-type-of-usage` gets to [the actual rules](https://jspecify.dev/docs/spec/#augmented-type-of-usage), whereas `augmented-type` is more about the concept. Either could be useful, and we could even try to incorporate both, maybe by linking to `augmented-type-of-usage` from "based only on JSpecify annotations on it or surrounding scopes."

Any preferences/recommendations?

wmdietl: Re 1: I would definitely be for making the anchors explicit. I had missed that there are automatically derived anchors.

Re 2: In that sentence you could link to both, first the one you already have and then the `augmented-type-of-usage` for "given type usage". But then again, there are several H2 that talk about types of usages.

One thing I notice about the spec document is that it is very flat. Maybe it would be helpful to change some sections from H2 to H3 to have more structure? We could have H2 #augmented-type-of-usage and move #intersection-types, #unbounded-wildcard, and a few others under that as H3. Then we would just need to link to the H2.

wmdietl (approved):
Changes look good. My comments can be done in follow-up PRs, here and in the spec.

docs/docs/tool-conformance.md
cpovirk: Thanks. I added the second link, and I filed https://github.com/jspecify/jspecify/issues/833 and https://github.com/jspecify/jspecify/issues/834 for the proposed spec changes.

PR #671, opened 2024-09-26
batch of javadoc revisions for clarity and completeness

kevinb9n:
I honestly don't remember everything I did here or why :-) so let's just go ahead and discuss how to make it best from here!

PR #815, opened 2026-07-07
Bump more links to Java 26, and remove an unused reference definition.

cpovirk:
We should probably get in the habit of bumping these at the same time as we bump the version we use to generate Javadoc (as done in PRs like https://github.com/jspecify/jspecify/pull/793).

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -1144,7 +1144,7 @@ If a type usage is the parameter of `equals(Object)` in a subclass of
 
 [#49]: https://github.com/jspecify/jspecify/issues/49
 [#65]: https://github.com/jspecify/jspecify/issues/65
-[Java SE 23]: https://docs.oracle.com/javase/specs/jls/se26/html/index.html
+[Java SE 23]: https://docs.oracle.com/javase/specs/jls/se23/html/index.html
 [JLS 1.3]: https://docs.oracle.com/javase/specs/jls/se26/html/jls-1.html#jls-1.3
 [JLS 15.16]: https://docs.oracle.com/javase/specs/jls/se26/html/jls-15.html#jls-15.16
 [JLS 15.20.2]: https://docs.oracle.com/javase/specs/jls/se26/html/jls-15.html#jls-15.20.2
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -384,7 +384,7 @@ innermost.
 > Packages are *not* enclosed by "parent" packages.
 >
 > This definition of "enclosing" largely matches
-> [the definition in the Java compiler API](https://docs.oracle.com/en/java/javase/26/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement\(\)).
+> [the definition in the Java compiler API](https://docs.oracle.com/en/java/javase/23/docs/api/java.compiler/javax/lang/model/element/Element.html#getEnclosingElement\(\)).
 > The JSpecify definition differs slightly by skipping type-parameter
 > declarations (which cannot be annotated with our declaration annotations) and
 > by defining that there exists a series of enclosing declarations for any type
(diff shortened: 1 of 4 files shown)


docs/docs/spec.md
```diff
@@ -1144,23 +1144,22 @@ If a type usage is the parameter of `equals(Object)` in a subclass of
...
-[JLS 4.5.1]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-4.html#jls-4.5.1
-[JLS 4.5.2]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-4.html#jls-4.5.2
-[JLS 4.5]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-4.html#jls-4.5
-[JLS 4.9]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-4.html#jls-4.9
-[JLS 4]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-4.html
-[JLS 5.1.10]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-5.html#jls-5.1.10
-[JLS 8.4.1]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-8.html#jls-8.4.1
-[JLS 8.4.8.1]: https://docs.oracle.com/javase/specs/jls/se23/html/jls-8.html#jls-8.4.8.1
-[JVMS 5.4.5]: https://docs.oracle.com/javase/specs/jvms/se14/html/jvms-5.html#jvms-5.4.5
+[Java SE 23]: https://docs.oracle.com/javase/specs/jls/se26/html/index.html
```
wmdietl: Should the "Java SE 23" also be updated, here and in the use?
Did we ensure that all features from SE 24-26 are covered?

cpovirk: Ouch, thanks for catching that. This was a total blind search-and-replace job by me, and I remember checking some links (like https://docs.oracle.com/en/java/javase/26/language/pattern-matching.html), but I didn't pay any attention to the left side of this line.

There _might_ not be anything else that we need to cover, but I'd want to at least think about [developments in pattern matching](https://openjdk.org/jeps/530). The point of this PR, though, isn't to do that, so I've reverted the change on this line.

wmdietl (approved):
You propose: "We should probably get in the habit of bumping these at the same time as we bump the version we use to generate Javadoc."

An alternative would be to consistently update the links together with the spec-supported JDK version. Then there is only one version used in the specification. The version used for generating Javadoc seems quite separate from the spec.

But the changes here look consistent with the first policy.

cpovirk:
That is a good point. On the one hand, a new version of the spec might contain corrections, or it might cover new features for which our existing rules naturally apply. But on the other, we might end up linking to a section that covers new features for which our existing rules would be incorrect.

I have pushed a new commit that limits the version number we use in the spec to 23. That means that I'm no longer making actual changes to `spec.md` except to remove an unused link, though I did keep making changes to the (obsolete?) `docs/locale/ja/LC_MESSAGES/tsttcpw.po`, just bumping it to 23 instead of to 26.

cpovirk:
(It's fair to observe that the links I'm updating in the _Javadoc_ include JLS links for `package-info` and for receiver parameters, along with a link to a page about records, whose features could also change from version to version. I think that those particular JLS sections are likely to remain good references across versions (or maybe we can [do better than a JLS link for at least one](https://github.com/jspecify/jspecify/pull/671#discussion_r3582095026)), and I think the world is probably better off when a Google search for "Java records" is more likely to lead to the most recent such documentation available, so I feel OK with this approach. It's still possible that I'm "defaulting" to linking to new pages more than makes sense here; we've just slowly found more and more (small) reasons that that's a good practice for _Javadoc_, so I'm carrying over those lessons to a different context where they might not apply.)

cpovirk:
(@kevinb9n, this, too, will conflict with you PR, though I think you may have bumped a version in a link or two already. Also, oops, sorry, I see now that the `package-info` link that I flagged in your PR was preexisting!)

Issue #813, opened 2026-07-07
Bump the build's default Java version past JDK 11?


cpovirk:
\[edit: This is all about the versions of build tools that we use to build JSpecify. JSpecify will remain compatible with Java 8.\]

Currently:
https://github.com/jspecify/jspecify/blob/8bb43e2905eda9ed1598b832d8fcbbe8b116305a/build.gradle#L31

But we already [use Java 26 for our Javadoc CI task](https://github.com/jspecify/jspecify/blob/8bb43e2905eda9ed1598b832d8fcbbe8b116305a/.github/workflows/docs.yml#L35), and in fact our current Javadoc sources are incompatible with Java 11.

Similarly, we need to be careful to set `JAVA_VERSION` [during releases](https://github.com/jspecify/jspecify/issues/305#issuecomment-2231830271).

I also wonder how javadoc.io picks a version of the Javadoc binary to use. It looks like [1.0.0 used Java 21](https://javadoc.io/static/org.jspecify/jspecify/1.0.0/org/jspecify/annotations/NonNull.html).

It would be nice to use Java 26 automatically everywhere, as [I've written in our wiki](https://github.com/jspecify/jspecify/wiki/version-compatibility#very-rough-notes-on-selecting-a-jdk-to-build-with). We would still test under older versions, but that's separate.

My understanding is that it should be very doable to have Gradle download the JDKs that we need. (Maybe we're already doing this for testing under older versions? I haven't checked.)

We'd bump the version periodically, as in https://github.com/jspecify/jspecify/pull/793. If there can trivially be a single source of truth for the version, then that's ideal. But we could just bump it in multiple places if that seems best.

Really, we'll probably end up wanting to bump the version in various other places, anyway, like:
- https://github.com/jspecify/jspecify/blob/8bb43e2905eda9ed1598b832d8fcbbe8b116305a/docs/docs/spec.md?plain=1#L1151
- https://github.com/jspecify/jspecify/blob/8bb43e2905eda9ed1598b832d8fcbbe8b116305a/src/main/java/org/jspecify/annotations/Nullable.java#L124

I don't remember the history for our current choice of version. I'm filing this issue now because a higher default Java version would have saved me from https://github.com/jspecify/jspecify/pull/812. (But of course that was a problem that was nice to fix anyway.)

tbroyer:
> I also wonder how javadoc.io picks a version of the Javadoc binary to use.

[...]

PR #816, opened 2026-07-07
Create a `projection` anchor in `Nullable`, and use `##` links.

cpovirk:
Currently, `NonNull` links to `Nullable.html#projection`, but there is no such anchor by that name. I've added one at the definition. This is still not as clear as having a section would be, but I'm not sure whether we want a section. Another option would be to change the text in `NonNull` to refer to "nullable projection" as...

```
"when `Nullable` is [applied to](Nullable#meaning-per-each-kind-of-type-usage-heading) a type variable
```

...or something to that effect.

(For the anchor, I introduced a `span`. If I make it completely empty, then I get a warning from Javadoc. Thus, I decided to make it cover the "title" of the list item.)

As for using [`##` links](https://docs.oracle.com/en/java/javase/26/docs/specs/javadoc/doc-comment-spec.html#:~:text=references%20to%20arbitrary%20URI%20fragments): That's what I had come here to do before I found the problem with `#projection`. Compare https://github.com/google/guava/pull/8346. From what I've seen, their main advantage is when linking to other projects (since you can link to, say, `Character##supplementary` instead of `https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/lang/Character.html#supplementary`. But I figure we might as well use them for links inside JSpecify, too.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ apply from: "gradle/format.gradle"
 apply from: "gradle/publish.gradle"
 
 java {
-    if (javaVersion.canCompileOrRun(15)) {
+    if (javaVersion.canCompileOrRun(21)) {
         withJavadocJar()
     }
     withSourcesJar()
--- a/src/main/java/org/jspecify/annotations/NonNull.java
+++ b/src/main/java/org/jspecify/annotations/NonNull.java
@@ -79,16 +79,16 @@ import java.lang.annotation.Target;
  * <p>If {@code E} has a non-null upper bound, then the apparent projection {@code @NonNull E} is
  * redundant but harmless.
  *
- * <p><a href="Nullable.html#projection">Nullable projection</a> serves the equivalent purpose in
- * the opposite direction, and is far more commonly useful.
+ * <p>{@linkplain Nullable##projection Nullable projection} serves the equivalent purpose in the
+ * opposite direction, and is far more commonly useful.
  *
  * <p>If a type variable has <i>all</i> its usages being projected in one direction or the other, it
  * should be given a non-null upper bound, and any non-null projections can then be removed.
  *
  * <h2>Where it is applicable</h2>
  *
- * <p>{@code @NonNull} is applicable in all the <a href="Nullable.html#applicability">same
- * locations</a> as {@link Nullable}.
+ * <p>{@code @NonNull} is applicable in all the {@linkplain Nullable##applicability same locations}
+ * as {@link Nullable}.
  */
 @Documented
 @Target(TYPE_USE)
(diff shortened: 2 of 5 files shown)


cpovirk:
Huh: From my experience in https://github.com/jspecify/jspecify/pull/812, I thought I'd confirmed that Javadoc runs already didn't work with versions like Java ~21~ 11. Yet I'm seeing [a failure under at least Java 17](https://github.com/jspecify/jspecify/actions/runs/28904748227/job/85749264463):

```
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/Nullable.java:85: error: unexpected text

 *       this way is called "nullable projection" ({@linkplain NonNull##projection non-null
                                                   ^
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/NullMarked.java:67: error: unexpected text
 *       and {@code @NonNull} are {@linkplain Nullable##applicability <b>not applicable</b>}, such
                                  ^
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/NonNull.java:41: error: anchor already defined: "projection"
 * <h2 id="projection">Non-null projection</h2>
       ^
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/NonNull.java:82: error: unexpected text
 * <p>{@linkplain Nullable##projection Nullable projection} serves the equivalent purpose in the
      ^
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/NonNull.java:90: error: unexpected text
 * <p>{@code @NonNull} is applicable in all the {@linkplain Nullable##applicability same locations}
                                                ^
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/NullUnmarked.java:55: error: unexpected text
 * <p>The effects of being null-marked are described in the {@linkplain NullMarked##effects Effects}
                                                            ^
Error: /home/runner/work/jspecify/jspecify/src/main/java/org/jspecify/annotations/NullUnmarked.java:75: error: unexpected text
 * The information in the {@linkplain NullMarked##where Where it can be used} section of {@code
                          ^
7 errors
> Task :javadoc FAILED
```

Maybe 17 is new enough to be stricter about errors but not new enough to recognize the new syntax? But I'd still have to figure out why 11 works at all....

Ah, now I remember:

https://github.com/jspecify/jspecify/blob/8bb43e2905eda9ed1598b832d8fcbbe8b116305a/build.gradle#L40-L42

Yet more motivation for https://github.com/jspecify/jspecify/issues/813?

wmdietl (approved):
Changes look consistent to me, but I do not understand the intricacies of different Javadoc versions.

cpovirk:
Thanks. Since you approved https://github.com/jspecify/jspecify/pull/818 (which I've merged), we shouldn't have to worry about different Java versions, so I think we're good.

And thank you for raising _that_: In #818, I should really have removed the bit that I'm updating here:

https://github.com/jspecify/jspecify/blob/0f5f17f78cdaab9156cdf6885787d93c01af0b55/build.gradle#L40-L42

I'll do that separately.

cpovirk:
@kevinb9n, this is going to create merge conflicts with your changes :\

kevinb9n:
No problem and I replicated this approach to my new links.

PR #820, opened 2026-07-14
Increase strictness of Docusaurus, updating headings accordingly.

cpovirk:
Apparently "Docusaurus v3 is using MDX v3," which requires escaping of braces in strict mode: https://docusaurus.io/docs/markdown-features/react#markdown-and-jsx-interoperability

(When I put together https://github.com/jspecify/jspecify/pull/814 to enable strict checking of anchors, I also asked Gemini for some ideas for other settings to make stricter. This PR enables some of those.)

### DIFF
--- a/docs/docs/spec.md
+++ b/docs/docs/spec.md
@@ -70,7 +70,7 @@ rules.
 > `UNSPECIFIED` and apply a policy that type usages like `Object` are always
 > non-nullable.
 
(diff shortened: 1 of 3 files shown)


wmdietl (approved):
Trivial addition of backslashes. I guess Docusaur would complain if you had missed one.

Issue #821, opened 2026-07-14
Consider some clarifications to recognized and unrecognized locations


cpovirk:
@kevinb9n pointed out that https://jspecify.dev/docs/spec/#recognized-type-use doesn't mention lambda parameters, even though https://github.com/jspecify/jspecify/issues/261 says that it should. We should deal with that, but it's (famous last words!) a simple PR, so this issue is about something else:

Prompted by that, I got Gemini to review the recognized-location section for other possible oversights. It didn't find any other errors, but the resulting report might be useful to tool authors who are looking to produce findings for JSpecify annotations in other locations:

> ### 1. Missing from **Recognized locations for type-use annotations**
> 
> * **Lambda formal parameter types ([JLS 15.27.1](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.27.1))**
>   * **Why it's missing:** `spec.md` (lines 220–221) recognizes:
>     `- a formal parameter type of a method or constructor, as defined in [JLS 8.4.1]`
>     In the JLS, lambda formal parameters are defined under `[JLS 15.27.1]` (`LambdaParameters`), not `[JLS 8.4.1]` (`FormalParameters` of methods and constructors).
>   * **Impact:** Explicitly typed lambda parameters such as `(@Nullable String s) -> ...` or `(@Nullable var s) -> ...` are currently not covered by any recognized bullet point.
> 
> * **Explicit type arguments on method/constructor invocations and method references ([JLS 15.12.2.1](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.12.2.1), [JLS 15.9](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.9), [JLS 15.13](https://docs.oracle.com/javase/specs/jls/se23/html/jls-15.html#jls-15.13)) *(Clarification)***
>   * **Where they appear:** e.g., `Collections.<@Nullable String>emptyList()`, `this.<@Nullable Foo>myMethod()`, `new <@Nullable Bar>MyClass()`, and `MyClass::<@Nullable Baz>someMethod`.
>   * **Impact:** While these are technically covered under `- a non-wildcard type argument`, readers might assume that bullet only refers to type arguments of parameterized reference types (like `List<@Nullable String>`). Noting invocation type arguments explicitly could prevent confusion.
> 
> ---
> 
> ### 2. Missing from **Exceptions / Notable unrecognized locations**
> 
[...]

Issue #261, opened 2022-08-11
Null-annotating a lambda parameter type: when can it be, when must it be, what does it mean if it's not?


kevinb9n:
The overall topic of implementation code, #114, is getting split out into #251 and #252 and now this. There is some brief discussion of lambda parameters in the original issue.

What is the deal with null-annotating lambda parameters?

Let's ignore nullness-unspecified types for now.

WHEN THE TARGET TYPE HAS NO WILDCARDS

All the lambda parameter base types will match those of the functional interface method signature exactly -- either they are inferred as such, or they themselves triggered that particular target type to be selected (target typing and/or maybe generic type inference too??), or they are spelled out just for clarity but still must match.

So, bringing nullness annotations into the picture: for ALL type components of the lambda parameter types, IF the types are given at all, our main choices are:

1. JSpecify dictates that all must be matched explicitly (max clarity, but a pain)
1. JSpecify dictates that none may be annotated because all will be inferred from the interface
1. JSpecify dictates that all will be inferred but it's harmless to annotate redundantly (could be confusing)
1. JSpecify dictates that lambda parameters are handled like local variables (annotate the subcomponents, don't annotate the root type)
1. JSpecify ignores how the lambda parameters are annotated because only the F.I. matters (let's scratch this)
1. JSpecify doesn't care what happens because this is implementation code

WHEN THE TARGET TYPE HAS WILDCARDS

Back to base types (without nullness annotation/analysis):

```
// m can pass only Strings into this consumer
void m(Consumer<? super String> c) {}

m(a -> out.println(a.hashCode()));
```

Javac target-types the base type of the lambda to `Consumer<String>`. You *can* coerce it to `Consumer<Object>` like this, but I see no real reason to:

```
m((Object a) -> out.println(a.hashCode()));
```

(Note this would've caused a pointless problem if `a.length()` were being called instead of `a.hashCode()`.)

Now to add nullness to the mix (note: assume `@NullMarked` context from here on),

First note that a functionally-covariant type parameter should always simulate having no upper bound, i.e. always have `@Nullable Object` as its upper bound:

```
interface Consumer<T extends @Nullable Object> {
  void accept(T t);
}
```

Let's assume the method above is left as it was:

```
// m can pass only non-null Strings into this consumer
void m(Consumer<? super String> c) {}
```

(because if it had gotten annotated `<? super @Nullable String>` then for purposes of _nullness_ annotations it should reduce to the same considerations as the no-wildcard case above.)

[...]

PR #822, opened 2026-07-14
List lambda parameter types as a recognized location.

cpovirk:
This was our decision back in https://github.com/jspecify/jspecify/issues/261, but we missed it in the spec until @kevinb9n pointed it out today.

Hopefully it's [the only true mistake in this section](https://github.com/jspecify/jspecify/issues/821).


kevinb9n (approved):
LGTM

PR #838, opened 2026-07-16
Import the nullness design FAQ as Markdown.

cpovirk:
Previously, it was [a Google Document](https://docs.google.com/document/d/1H3TdqjRZQiGEP7ZePLpTnbYxj9s5hJAn9sp6eOiuj_c/edit).

Known/intentional changs:

- removing the introductory material
- adding named anchors
- undoing smart quotes
  - Maybe we should use them throughout our documentation, but currently we don't.
  - Or maybe we can get Docusaurus to translate to them automatically.
- replacing `http` with `https`

I note that "What about more complex kinds of nullness" trails off with a "TODO." I left it that way.

PR #839, opened 2026-07-16
Render smart/curly quotes.

cpovirk:
I had wondered about this in https://github.com/jspecify/jspecify/pull/838.


cpovirk:
I'm not sure if we care, but I wanted to see how easy this was. It does appear to work.

If anyone thinks this is a good idea, feel free to approve, and I'll merge it. But I'm also fine with having it sit indefinitely.

kevinb9n:
Eh? Never could care about smart quotes.

PR #854, opened 2026-07-24
Bump Dependabot cooldown period from three days to seven.

cpovirk:
Three is [the default](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-).

Seven is a recommendation that I've seen a couple places, including in some Google-internal documentation. I'd hope that it would never make a difference in practice here either way.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
   directory: "/"
   schedule:
     interval: "weekly"
+  cooldown:
+    default-days: 7
   groups:
     dependencies:
       applies-to: version-updates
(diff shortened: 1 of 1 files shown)


cpovirk:
I'm [increasingly unsure as to whether this is worth it](https://github.com/google/jimfs/pull/497#issuecomment-5076554544), so I'm trying to get some updated advice from GitHub folks at Google. My sense from a brief look is that it's not worth it for an average project but that it might be defensible for a project that aims to be extremely widely used.

cpovirk:
Let's do this after all: As I just noted in https://github.com/jspecify/jdk/pull/155:

> This is [still](https://github.com/zizmorcore/zizmor/pull/2193) what [`zizmor` wants](https://docs.zizmor.sh/audits/#rulesdependabot-cooldownconfigdays).

PR #859, opened 2026-07-27
Codex found a few more questions the FAQ already answers.

kevinb9n:


cpovirk:
This is showing up with the same changes as https://github.com/jspecify/jspecify/pull/848, and I think I see those changes at head. Is this the right branch, or am I confused?

kevinb9n:
Yeah sorry, I just saw it and for some reason had no memory of having actually sent it

PR #860, opened 2026-07-29
Migrate to `com.vanniktech.maven.publish`.

cpovirk:
(It sets up `withJavadocJar()` and `withSourcesJar()` automatically.)

I used this to publish release 1.0.1.

Fixes https://github.com/jspecify/jspecify/pull/827 (by eliminating that section entirely)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,10 +53,10 @@ jobs:
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@v6
       - name: Publish conformance tests snapshot
-        run: ./gradlew publishConformanceTestsPublicationToSonatypeRepository --no-configuration-cache
+        run: ./gradlew publishConformanceTestsPublicationToMavenCentralRepository --no-configuration-cache
         env:
-          ORG_GRADLE_PROJECT_sonatypeUsername: ${{ secrets.sonatype_username }}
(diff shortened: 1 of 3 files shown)


cpovirk:
- I updated https://github.com/jspecify/jspecify/pull/187#issuecomment-873183875 with revised release instructions for the new plugin.
- The GitHub Actions Scan errors are preexisting issues that will be fixed by https://github.com/jspecify/jspecify/pull/861.

jpenilla:
Hey, I ran into a small problem with the 1.0.1 publication while migrating a project to JSpecify and filed #867 with the details. The Javadoc jar itself is there, but `javadocElements` is missing from the Gradle module metadata, unlike in 1.0.0.

This looks like the same behavior described in vanniktech/gradle-maven-publish-plugin#1028. The full plugin does create and publish a Javadoc jar automatically, but that apparently isn't quite equivalent to `java.withJavadocJar()` for Gradle metadata consumers. That issue mentions the base plugin as a workaround.

From the related PRs, it looks like the main goal here is to move off the Nexus publishing plugin onto the current Central Portal API and restore configuration-cache compatibility. I've had good results with [`dev.lukebemish.central-portal-publishing`](https://github.com/lukebemishprojects/CentralPortalPublishing) in my projects. It handles Central Portal bundle publication while leaving the normal `maven-publish`/`signing` setup in place, and is designed for configuration cache and isolated projects. It might be worth a look as an alternative to the full Vanniktech plugin here.

PR #875, opened 2026-08-15
NullnessUnspecifiedTypeParameter: also allow error on variable declaration

wmdietl:
Expect two errors:
- On the variable declaration, as the explicit type argument is invalid;
- On the object creation - the diamond is inferred from the LHS and therefore also invalid.

The second one is up for debate, but the current EISOP behavior.


cpovirk (approved):
Agreed that at minimum the variable declaration should be an error.

I could see just removing the initializers at some point.

Also for the future: I think I've noticed before that "NullnessUnspecifiedTypeParameter" strikes me as a strange name for this sample file, since it uses specified nullness throughout. Maybe someday we'll rename it.

PR #899, opened 2026-09-16
Enable more verbose test output and stricter runtime checking, and fix a JUnit warning that either reveals.

cpovirk:
This includes displaying stderr+stdout and displaying stack traces.

Once we can see stderr, we see this warning:

`@Nested class 'org.jspecify.annotations.NullMarkedTest$WithJava8' must not be static. It will only be executed if discovered as a standalone test class. You should remove the annotation or make it non-static to resolve this warning.`

You may recall that we can't make the class non-static because of https://github.com/jspecify/jspecify/pull/517.

As best I can tell, there was no need to use nested classes here at all back in https://github.com/jspecify/jspecify/pull/163, and incidentally we can use `@EnabledOnJre`+`@DisabledOnJre`.

This PR was prompted by https://github.com/uber/NullAway/pull/1835. Apparently I could just have looked at [the similar configuration in the reference checker](https://github.com/jspecify/jspecify-reference-checker/commit/9ca43aa9b80db1e7aac69c0331437386bb8f68d5#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R49). I did deviate slightly from that here:
- I didn't carry over `showStackTraces = false`: We didn't need stack traces for failures in the samples because the interesting part there is the line number for the samples, but in a more "normal" project like, they're nice. (At least they add the failure message and full test class's name, though sadly Gradle doesn't trim the stack trace as I would have hoped.)
- I didn't carry over: `events "failed"`: That tells Gradle to output events of the `FAILED` type, but that type that is already displayed by default (along with `STANDARD_OUT` and `STANDARD_ERROR` from the new `showStandardStreams = true`). We could remove `events "failed"` from the reference-checker build, too.

Demo of the before-and-after output for a failure:

```
@NullMarked > onlyBasicReflectionWorksUnderJava8() FAILED
    org.opentest4j.AssertionFailedError at NullMarkedTest.java:53
```

```
@NullMarked > onlyBasicReflectionWorksUnderJava8() FAILED
    org.opentest4j.AssertionFailedError: Expected java.lang.ArrayStoreException to be thrown, but nothing was thrown.
        at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:152)
        at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:73)
        at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:35)
        at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3128)
        at org.jspecify.annotations.NullMarkedTest.onlyBasicReflectionWorksUnderJava8(NullMarkedTest.java:53)
```


cpovirk:
Gemini thinks that the current setup is even worse than I would have guessed:

> You should be **fairly concerned**—in fact, the current setup is *already* partially broken for targeted test runs, even before removing `@Nested`.
> 
> ### 1. What is happening right now (and why `@Nested` isn't helping)
> Even with `@Nested` present, JUnit Jupiter completely ignores `static` member classes when discovering children of an enclosing class. As a result, running [`NullMarkedTest`](https://github.com/jspecify/jspecify/blob/main/src/integrationTest/java/org/jspecify/annotations/NullMarkedTest.java#L35) directly right now **fails**:
> 
> ```sh
> $ ./gradlew integrationTest --tests org.jspecify.annotations.NullMarkedTest
> > No tests found for given includes: [org.jspecify.annotations.NullMarkedTest](--tests filter)
> ```
> 
> The only reason the tests run during `./gradlew check` is that Gradle scans `build/classes/java/integrationTest` for all `.class` files and passes `NullMarkedTest$WithJava8` and `NullMarkedTest$WithJava9OrLater` to JUnit Platform as individual standalone `ClassSelector`s. Removing `@Nested` wouldn't change Gradle's current behavior, because `@Nested` is already being ignored.
> 
> ### 2. Why `static class WithJava8` / `WithJava9OrLater` is risky going forward
> If left as `static` member classes named `WithJava8` and `WithJava9OrLater`:
> * **Name-based filtering will skip them**: Their class names (`...NullMarkedTest$WithJava8`) do **not** match JUnit Platform's default class name include pattern (`^(Test.*|.+[.$]Test.*|.*Tests?)$`) because the simple name after `$` neither starts nor ends with `Test`. Any runner using classpath/package discovery (such as JUnit's `ConsoleLauncher`, `@Suite`, Maven Surefire, or future Gradle discovery modes that delegate filtering to the JUnit Platform engine) will silently ignore them.
> * **IDE / `--tests` runs remain broken**: Running `NullMarkedTest` in an IDE or via `--tests org.jspecify.annotations.NullMarkedTest` will continue to report 0 tests.
> 
> ### 3. Why non-static `@Nested` failed on Java 8
> When `javac` (JDK 21+) compiles an inner class constructor (`WithJava8(NullMarkedTest this$0)`), even with `--release 8`, it emits a `MethodParameters` attribute marking the synthetic `this$0` parameter as mandated with an empty name (`""`). When JUnit Jupiter calls `Constructor.getParameters()` to instantiate the `@Nested` class, Java 8's reflection implementation throws `MalformedParametersException: Invalid parameter name ""` ([JDK-8075498](https://bugs.openjdk.org/browse/JDK-8075498)).
> 
> ### 4. An alternative to top-level classes: flatten into `NullMarkedTest`
> Since [`WithJava8`](https://github.com/jspecify/jspecify/blob/main/src/integrationTest/java/org/jspecify/annotations/NullMarkedTest.java#L49) and [`WithJava9OrLater`](https://github.com/jspecify/jspecify/blob/main/src/integrationTest/java/org/jspecify/annotations/NullMarkedTest.java#L70) each contain only a single `@Test` method, we don't even need separate top-level classes—we can put both `@Test` methods directly inside [`NullMarkedTest`](https://github.com/jspecify/jspecify/blob/main/src/integrationTest/java/org/jspecify/annotations/NullMarkedTest.java#L35) and check the Java version inside each method (or use JUnit Jupiter's `@EnabledOnJre(JRE.JAVA_8)` / `@DisabledOnJre(JRE.JAVA_8)`):
> 
> ```java
> @DisplayName("@NullMarked")
> class NullMarkedTest {
>   ...
> 
>   @Test
>   @EnabledOnJre(JRE.JAVA_8)
>   void onlyBasicReflectionWorksUnderJava8() {
>     Object unused = NullMarked.class.getMethods();
>     assertThrows(ArrayStoreException.class, NullMarked.class::getAnnotations);
>   }
> 
>   @Test
>   @DisabledOnJre(JRE.JAVA_8)
>   void annotationIncludesModuleAsTargetUnderJava9OrLater() {
>     Set<String> targets = loadTargets();
>     assertEquals(
>         new HashSet<String>(Arrays.asList("TYPE", "METHOD", "CONSTRUCTOR", "PACKAGE", "MODULE")),
>         targets);
>   }
> }
> ```

PR #900, opened 2026-09-18
Remove no longer needed 'but_expect_nothing' markers and obsolete error expectations

wmdietl:
Removes `jspecify_but_expect_nothing` markers across samples where the prototype checker now properly reports diagnostics, and removes obsolete `jspecify_conflicting_annotations` expectations on `? super Object` wildcards.


cpovirk (approved):
Huh, I'd forgotten the history of the `jspecify_conflicting_annotations`. I see that we have them only in the `samples-google-prototype-eisop` branch, so a change here SGTM.

It's possible that an even better solution someday will be to avoid creating that situation in our samples at all (by using `? super SomeOtherType`?), except perhaps for having one sample that exercises that case intentionally (like the existing `ContainmentSuperVsExtendsSameType`). But I'm happy not to think about that now.

wmdietl:
With this, https://github.com/eisop/jspecify-reference-checker/pull/90 should now be green :-)

